PR #25, opened 2018-08-22
Internationalization

Ordissimo:
Partial functioning of translations.
Full French translation.


src/main.cpp
```diff
@@ -8,6 +10,20 @@ int main(int argc, char *argv[])
     QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
     KiwixApp a(argc, argv);
 
+    // format systems language
+    QString defaultLocale = QLocale::system().name(); // e.g. "de_DE"
+    defaultLocale.truncate(defaultLocale.lastIndexOf('_')); // e.g. "de"
+    QLocale::setDefault(defaultLocale);
```
mgautierfr: Why do you change the default locale ?
We should keep the full locale, Qt already has a mechanism to fallback to `fr` if `fr_FR` doesn't exist (https://doc.qt.io/qt-5/qtranslator.html#load)

If we change the locale to `fr` we will not be able to have different translation for `fr_FR` and `fr_CA`

src/main.cpp
```diff
@@ -8,6 +10,20 @@ int main(int argc, char *argv[])
...
+    
+    QTranslator qtTranslator;
+    qtTranslator.load("qt_" + QLocale::system().name(),
+                             QLibraryInfo::location(QLibraryInfo::TranslationsPath));
+    a.installTranslator(&qtTranslator);
+
+    QTranslator myappTranslator;
+    myappTranslator.load(":/i18n/kiwix-desktop.qm");
+    a.installTranslator(&myappTranslator);
+
```
mgautierfr: All those initialization would be better directly in the KiwixApp code.

src/library.cpp
```diff
@@ -15,7 +17,7 @@ QString Library::openBook(const QString &zimPath)
         if(QString::fromStdString(it->second->getZimFilePath()) == zimPath)
             return it->first;
     }
-    qInfo() << "Opening" << zimPath;
+    qInfo() << QObject::tr("Opening") << zimPath;
```
mgautierfr: I'm not sure we need to translate log messages.

kiwix-desktop.pro
```diff
@@ -82,6 +83,9 @@ isEmpty(PREFIX) {
 target.path = $$PREFIX/bin
 INSTALLS += target
 
+TRANSLATIONS = "resources/i18n/kiwix-desktop_fr.ts"
+CODECFORSRC = UTF-8
+
```
mgautierfr: It is just cosmetic, but maybe move them just after `FORMS`

kiwix-desktop.pro
```diff
@@ -6,6 +6,7 @@
 
 QT       += core gui
 QT       += webenginewidgets
+QT       += widgets qml quick
```
mgautierfr: Why this change ?

Ordissimo:
I'm new to Qt. Accustomed to gettext :-)

src/main.cpp
Ordissimo: Ok

src/library.cpp
Ordissimo: effectively.

kiwix-desktop.pro
Ordissimo: Ok

src/main.cpp
Ordissimo: Great

mgautierfr:
I have updated your PR and add some changes in the code and update the translations.

I wanted to merged it using the github's command line instructions but it seems that github did catch the merge. Anyway, you PR is merged and I'm closing it.

Thanks again for this PR.

LAfricain:
New ts file with some corrections and more translation. 
[kiwix-desktop_fr.zip](https://github.com/kiwix/kiwix-desktop/files/3370085/kiwix-desktop_fr.zip)
Some part are not yet translated and I did not found it in the sources. Like Search or Open, in the library area.

PR #28, opened 2018-08-28
Honor the application color scheme

niemand-deu:
Right now most colors are fixed to a permanent light theme, which can make Kiwix not fit in with other Qt applications.
Additionally the interface can become unreadable as undefined colors revert to the system's default color theme. In the case of Breeze Dark (KDE's default dark theme) the text camouflages into the menu.
This should alleviate the problem, although the icons will still look out of place in dark themes.

### DIFF
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -1,8 +1,3 @@
-QWidget {
-    background-color: #EAECF0;
-}
-
-
 /* ------------------------------------
    The topBar
 */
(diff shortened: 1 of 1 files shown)


kelson42:
@jetownfeve21 @mgautierfr This is a really old PR, but I would like to read your comment on it.

kelson42:
This has been superseeded by https://github.com/kiwix/kiwix-desktop/pull/191

PR #89, opened 2018-12-28
README: Fix typo in package name

legoktm:
I assume it was supposed to be libqt5svg5-dev with a q, not a g.


legoktm:
Also, I can't find a `libqt5gui` package, just `libqt5gui5`.

https://packages.debian.org/unstable/libqt5gui5

PR #116, opened 2019-03-26
Update searchbar.cpp

Ritvyk:
Structural Errors Are Removed.

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -1,8 +1,6 @@
 #include "searchbar.h"
-
 #include <QCompleter>
 #include <QTimer>
-
 #include "kiwixapp.h"
 
 SearchButton::SearchButton(QWidget *parent) :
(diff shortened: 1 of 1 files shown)


mgautierfr:
Thank you @Ritvyk for your PR but I don't see the need for it.

- Your commit said "Structural Errors Are Removed." but your remove nothing, you simply add few comments.
- Your comments are somehow useless, they simply describe what the code does, line by line, and this can be easily inferred by simply reading the code.

I have nothing about writing few comments to describe complex part in the code [*]. But adding a comment "adding the bookmark in the library" to the code `library->addBookmark(bookmark);` seems useless and even counterproductive because it add unnecessary things to read.

[*] And if the code is too complex, it may be even better to rewrite it correctly to make it easier to read (rename correctly the variables, create function with explicit name, ...)

kelson42:
No feedback. I close the PR.

PR #122, opened 2019-04-02
bug fix loading remote library when local file selected

jetownfeve21:
add an if to check if it's local files that is selected
` if (!local)
        emit(remoteParamsChanged());`

### DIFF
--- a/src/about.cpp
+++ b/src/about.cpp
@@ -1,8 +1,8 @@
 #include "about.h"
 #include "ui_about.h"
 
-#define _STR(X) #X
-#define STR(X)  _STR(X)
+#define _STR(...) # __VA_ARGS__
+#define STR(X) _STR(X)
 
 About::About(QWidget *parent) :
     QDialog(parent),
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -225,6 +225,8 @@ void ContentManager::setCurrentCategoryFilter(QString category)
 #define CATALOG_PORT 80
 #define CATALOG_URL "library.kiwix.org"
 void ContentManager::updateRemoteLibrary() {
+    if (m_local)
+        return;
     QUrlQuery query;
     if (m_currentLanguage != "*") {
         query.addQueryItem("lang", m_currentLanguage);


src/about.cpp
```diff
@@ -2,20 +2,18 @@
...
-    ui->setupUi(this);
-    auto htmlText = ui->aboutText->toHtml();
-    htmlText = htmlText.replace("%%VERSION%%", STR(GIT_VERSION));
-    htmlText = htmlText.replace("%%BUILD_DATE%%", STR(BUILD_DATE));
-    ui->aboutText->setHtml(htmlText);
+	ui->setupUi(this);
+	auto htmlText = ui->aboutText->toHtml();
+	htmlText = htmlText.replace("%%VERSION%%", STR(GIT_VERSION));
+	htmlText = htmlText.replace("%%BUILD_DATE%%", STR(BUILD_DATE));
+	ui->aboutText->setHtml(htmlText);
```
mgautierfr: The indentation is 4 spaces. Please do not change it.

src/about.cpp
```diff
@@ -2,20 +2,18 @@
 #include "ui_about.h"
 
 #define _STR(X) #X
-#define STR(X)  _STR(X)
+#define STR(X) _STR((X))
```
mgautierfr: This add a parentheses around the value to transform in string.
This should work better :
```
#define _STR(...) # __VA_ARGS__
#define STR(X) _STR(X)
```

src/contentmanager.cpp
```diff
@@ -31,7 +31,8 @@ void ContentManager::setLocal(bool local) {
         return;
     }
     m_local = local;
-    emit(remoteParamsChanged());
+    if (!local)
+        emit(remoteParamsChanged());
```
mgautierfr: There is several place where the signal `remoteParamsChanged` is emit.
It would be better to move the test of `m_local` in the listener of the signal (`updateRemoteLibrary`)

src/about.cpp
mgautierfr: Fixed in last forced push.

src/contentmanager.cpp
```diff
@@ -31,8 +32,7 @@ void ContentManager::setLocal(bool local) {
         return;
     }
     m_local = local;
-    if (!local)
-        emit(remoteParamsChanged());
+    emit(remoteParamsChanged());
```
mgautierfr: This change should be squash with the previous commit.

src/contentmanager.cpp
```diff
@@ -219,13 +219,19 @@ void ContentManager::setCurrentLanguage(QString language)
...
+    }
     emit(remoteParamsChanged());
 }
 
 #define CATALOG_HOST "library.kiwix.org"
 #define CATALOG_PORT 80
 #define CATALOG_URL "library.kiwix.org"
 void ContentManager::updateRemoteLibrary() {
+    if (m_local)
+        return;
```
mgautierfr: This change should be squash with the previous commit.

src/contentmanager.cpp
```diff
@@ -23,6 +23,7 @@ ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader,
     setCurrentLanguage(QLocale().name().split("_").at(0));
     connect(mp_library, &Library::booksChanged, this, [=]() {emit(this->booksChanged());});
     connect(this, &ContentManager::remoteParamsChanged, this, &ContentManager::updateRemoteLibrary);
+    connect(this, &ContentManager::remoteParamsChanged, this, &ContentManager::updateLocalLibrary);
```
mgautierfr: It is a bit strange to connect `updateLocalLibrary` slot to `remoteParamsChanged` signal.
Maybe we could rename `remoteParamsChanged` to `filterParamsChanged` to better describe what the signal is about ?

PR #125, opened 2019-04-05
Filter by category on local files

jetownfeve21:


src/library.cpp
```diff
@@ -121,6 +121,9 @@ void Library::save()
 {
     m_library.writeToFile(appendToDirectory(getDataDirectory(),"library.xml"));
     m_library.writeBookmarksToFile(appendToDirectory(getDataDirectory(), "library.bookmarks.xml"));
+    auto manipulator = LibraryManipulator(this);
+    auto manager = kiwix::Manager(&manipulator);
+    manager.readFile(appendToDirectory(getDataDirectory(),"library.xml"), false);
```
mgautierfr: This seems unnecessary. Loading something we just save should "return" the same thing.
If not, it is a bug in the kiwix-lib and it should be fixed there.

I think the best here is to simply remove this commit and open an issue to keep a trace of this change and investigate further.

src/contentmanager.cpp
```diff
@@ -212,42 +211,42 @@ void ContentManager::setCurrentLanguage(QString language)
...
-        return;
-    QUrlQuery query;
-    if (m_currentLanguage != "*") {
-        query.addQueryItem("lang", m_currentLanguage);
-    }
-    query.addQueryItem("count", QString::number(0));
-    if (m_categoryFilter != "all") {
-        query.addQueryItem("tag", m_categoryFilter);
+void ContentManager::updateLibrary() {
+    if (!m_local) {
```
mgautierfr: It is better to keep the "early exit" here. 
```cpp
if (m_local) {
    emit(booksChanged());
    return;
}
...
```
This way, all the later code is letf unchanged and we avoid unnecessary diff to review.

PR #130, opened 2019-04-16
Unresponsive desktopapp server down fix

### LINKED ISSUE #121: kiwix-desktop becomes unresponsive if library.kiwix.org is down.
We should be resilient to error on the server side.

jetownfeve21:
Fixes #121

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -13,7 +13,8 @@
 ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader, QObject *parent)
     : QObject(parent),
       mp_library(library),
-      mp_downloader(downloader)
+      mp_downloader(downloader),
+      m_remoteLibraryManager()
 {
     // mp_view will be passed to the tab who will take ownership,
     // so, we don't need to delete it.
@@ -23,6 +24,7 @@ ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader, 
     setCurrentLanguage(QLocale().name().split("_").at(0));
     connect(mp_library, &Library::booksChanged, this, [=]() {emit(this->booksChanged());});
     connect(this, &ContentManager::filterParamsChanged, this, &ContentManager::updateLibrary);
+    connect(&m_remoteLibraryManager, &OpdsRequestManager::requestReceived, this, &ContentManager::updateRemoteLibrary);
 }
 
 void ContentManager::setLocal(bool local) {
@@ -232,36 +234,22 @@ void ContentManager::setCurrentCategoryFilter(QString category)
     emit(filterParamsChanged());
 }
 
-#define CATALOG_HOST "library.kiwix.org"
-#define CATALOG_PORT 80
-#define CATALOG_URL "library.kiwix.org"
 void ContentManager::updateLibrary() {
     if (m_local) {
         emit(booksChanged());
         return;
     }
-    QUrlQuery query;
-    if (m_currentLanguage != "*") {
-        query.addQueryItem("lang", m_currentLanguage);
-    }
-    query.addQueryItem("count", QString::number(0));
-    if (m_categoryFilter != "all") {
-        query.addQueryItem("tag", m_categoryFilter);
-    }
-    QUrl url;
-    url.setScheme("http");
-    url.setHost(CATALOG_HOST);
-    url.setPort(CATALOG_PORT);
-    url.setPath("/catalog/search");
-    url.setQuery(query);
-    qInfo() << "Downloading" << url.toString(QUrl::FullyEncoded);
+    try {
+        m_remoteLibraryManager.doUpdate(m_currentLanguage, m_categoryFilter);
+    } catch (runtime_error&) {}
+}
+
+#define CATALOG_URL "library.kiwix.org"
+void ContentManager::updateRemoteLibrary(const QString& content) {
     m_remoteLibrary = kiwix::Library();
     kiwix::Manager manager(&m_remoteLibrary);
-    try {
-        auto allContent = kiwix::download(url.toString(QUrl::FullyEncoded).toStdString());
-        manager.readOpds(allContent, CATALOG_URL);
-    } catch (runtime_error&) {}
-    emit(booksChanged());
+    manager.readOpds(content.toStdString(), CATALOG_URL);
+    emit(this->booksChanged());
 }
 
 void ContentManager::setSearch(const QString &search)
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -236,10 +236,12 @@ void ContentManager::setCurrentCategoryFilter(QString category)
 
 void ContentManager::updateLibrary() {
     if (m_local) {
+        emit(pendingRequest(false));
         emit(booksChanged());
         return;
     }
     try {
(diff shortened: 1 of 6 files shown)


src/remotelibrarymanager.h
```diff
@@ -0,0 +1,32 @@
+#ifndef REMOTELIBRARYMANAGER_H
+#define REMOTELIBRARYMANAGER_H
+
+#include <QObject>
+#include <QtNetwork>
+#include <kiwix/library.h>
+#include <kiwix/manager.h>
+
+class RemoteLibraryManager : public QObject
```
mgautierfr: I would name it differently.
This class doesn't handle (manage) a remote library. It handle a request to the server to get the result of a new
request.
Something like OpdsRequestManager would be better.

We could even go further. If the class is responsible to handle request to an opds server, we could (should) move the url creation in `ContentManager::updateLibrary` into this class.

---
Or this is really a RemoteLibraryManager, and so it should handle a library. And so to have `ContentManager::updateRemoteLibrary` implemented here.

src/remotelibrarymanager.h
```diff
@@ -0,0 +1,32 @@
...
+
+public:
+    void doUpdate(const QUrl &url);
+
+private:
+    QNetworkAccessManager m_networkManager;
+    QNetworkReply *mp_reply;
+
+signals:
+    void booksChanged(const QString&);
```
mgautierfr: This signal is emit when we get a response, not when the books have changed.
So it should be name "requestReceived", "newOpdsStream", ...

src/remotelibrarymanager.cpp
```diff
@@ -0,0 +1,28 @@
...
+RemoteLibraryManager::RemoteLibraryManager()
+: mp_reply(nullptr)
+{
+}
+
+void RemoteLibraryManager::doUpdate(const QUrl &url)
+{
+    QNetworkRequest request(url);
+    if (mp_reply) {
+        mp_reply->abort();
```
mgautierfr: Should we also call `mp_reply->deleteLater()` ?

src/contentmanager.cpp
```diff
@@ -242,14 +244,18 @@ void ContentManager::updateLibrary() {
...
     url.setPath("/catalog/search");
     url.setQuery(query);
+    
     qInfo() << "Downloading" << url.toString(QUrl::FullyEncoded);
-    m_remoteLibrary = kiwix::Library();
-    kiwix::Manager manager(&m_remoteLibrary);
     try {
-        auto allContent = kiwix::download(url.toString(QUrl::FullyEncoded).toStdString());
-        manager.readOpds(allContent, CATALOG_URL);
+        m_remoteLibrary = kiwix::Library();
```
mgautierfr: This is an open question here: Should we "reset" the library when we do the request (so here), or when we receive
the response of the request (so in updateRemoteLibrary) ?

src/contentmanager.h
```diff
@@ -45,6 +45,7 @@ class ContentManager : public QObject
     void booksChanged();
     void downloadsChanged();
     void currentLangChanged();
+    void displayLoadIcon(const bool);
```
mgautierfr: I wonder if we could name it differently.
`displayLoadIcon` is more about what to do with this signal. But the signal should be about what is going on.
Something like `pendingRequest` or `Loading`. It would be to the slot of the signal to decide what to do when a request is pending (display a loading icon).

src/contentmanager.cpp
```diff
@@ -26,6 +26,7 @@ ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader,
     connect(mp_library, &Library::booksChanged, this, [=]() {emit(this->booksChanged());});
     connect(this, &ContentManager::filterParamsChanged, this, &ContentManager::updateLibrary);
     connect(&m_remoteLibraryManager, &RemoteLibraryManager::booksChanged, this, &ContentManager::updateRemoteLibrary);
+    connect(&m_remoteLibraryManager, &RemoteLibraryManager::booksChanged, this, [=]() {emit(this->displayLoadIcon(false));});
```
mgautierfr: Could this signal being emit in `ContentManager::updateRemoteLibrary` instead of creating a new connection ?

src/contentmanager.cpp
jetownfeve21: i don't see big differences between these two options.

src/remotelibrarymanager.cpp
jetownfeve21: When mp_reply->abort() is called, it emits a 'finished' signal which is connected to the slot "receivedContent" where I already call mp_reply->deleteLater() so I think not.

https://github.com/kiwix/kiwix-desktop/blob/2bf05ddb6c4a6f0b1fdb2bcd0d2b4d2f0b2630d9/src/remotelibrarymanager.cpp#L8-L28

src/remotelibrarymanager.h
jetownfeve21: Why do we use macro for the url's creation ? 

https://github.com/kiwix/kiwix-desktop/blob/a005a5cb2580d44a980a316e4a710011dceec659/src/contentmanager.cpp#L223-L225

mgautierfr: This is :
- 1. to move constants out of the code.
- 2. give some meanings to the constant.

It seems pretty obvious that in `url.setPort(80);`, `80` is the port of the request, and so probably the port of the catalog server.
But `CATALOG_HOST` and `CATALOG_URL`, despite they have the same values, are not the same things.
`CATALOG_HOST` is the host. So, no port or path.
`CATALOG_URL` is the (full) url of the catalog endpoint. It is use by the library to reconstruct absolute urls from relative ones include in the opds stream. It include at least the port. If the port is different from 80, `CATALOG_URL` should include it and be different than `CATALOG_HOST`.

If we directly put the string in the code, we lost the ideas that they are different things.

src/remotelibrarymanager.cpp
mgautierfr: We are good so.

src/contentmanager.cpp
```diff
@@ -232,36 +234,24 @@ void ContentManager::setCurrentCategoryFilter(QString category)
...
+        emit(pendingRequest(true));
+        m_remoteLibrary = kiwix::Library();
+        m_remoteLibraryManager.doUpdate(m_currentLanguage, m_categoryFilter);
     } catch (runtime_error&) {}
-    emit(booksChanged());
+}
+
+void ContentManager::updateRemoteLibrary(const QString& content) {
+    kiwix::Manager manager(&m_remoteLibrary);        
+    manager.readOpds(content.toStdString(), "library.kiwix.org");
```
mgautierfr: We should keep CATALOG_URL here. (And keep its definition in `src/contentmanagaer.cpp` instead of moving it in `src/opdsrequestmanager.cpp` where it is never used.

src/contentmanager.cpp
mgautierfr: Rethinking about it, it would prefer to se the reset of the library in `updateRemoteLibrary`.
The need for a clear library is required by the fact that the manager doesn't remove books.
This is an implementation detail and it should be an internal step of `updateRemoteLibrary`.

PR #133, opened 2019-04-18
Dynamic window title

### LINKED ISSUE #129: Fix window title (replace "MainWindow")
Kiwix Desktop Windows title is always "MainWindows - Kiwix".

It should be:
* "HTML article title - Kiwix" in case a content tab is open
* "Library - Kiwix" in case the library is open (please think that "Library" should be localised like any other part of the UI).

Quentin-Hsk:
Fixes #129


src/mainwindow.cpp
```diff
@@ -24,6 +24,8 @@ MainWindow::MainWindow(QWidget *parent) :
     connect(app->getAction(KiwixApp::AboutAction), &QAction::triggered,
             mp_about, &QDialog::show);
     connect(app, &KiwixApp::currentTitleChanged, this, [=](const QString& title) {
+        if (title.startsWith("zim://"))
+            return;
```
mgautierfr: Maybe set the title to "Library - Kiwix" in this case. So we avoid to keep the previous (wrong) title of the previous page.

PR #135, opened 2019-04-19
Delete books in a custom contextmenu

### LINKED ISSUE #97: Add "delete file" option in right-click and/or File menus
Currently there is no way to delete a file from the application's environment on Windows/beta2

jetownfeve21:
add a column at the book's list's table with "delete" button. This ui
is temporary.

On click, this button sends the id of the selected book at the
ContentManager class which gets book by this id, and erases all files
in relationship with this book on the computer.

Fixes #97

### DIFF
--- a/resources/texts/_contentManager.html
+++ b/resources/texts/_contentManager.html
@@ -74,6 +74,9 @@ function init() {
             downloadUpdaters[book.id] = setInterval(function() { getDownloadInfo(book.id); }, 1000);
           });
         },
+        eraseBook : function(book) {
+            contentManager.eraseBook(book.id);
+        },
         displayedBooks : function(books, nb) {
             var a = books.slice(0, nb);
             return a;
@@ -215,6 +218,7 @@ details:hover {
   </div>
   <div id="bookTable">
     <div class="header">
+      <span class="tablecell cell0">icone poubelle</span>
       <span class="tablecell cell0"></span>
       <span class="tablecell cell1">File name</span>
       <span class="tablecell cell2">Size</span>
@@ -225,6 +229,9 @@ details:hover {
     <div id="bookList" onscroll=scrolled(this)>
       <details v-for="book in displayedBooks(books, displayedBooksNb)" class="book">
         <summary class="tablerow">
+          <span class="tablecell cell0">
+            <button v-if="book.path" v-on:click="eraseBook(book)">Delete</button>
+          </span>
           <span class="tablecell cell0">
             <img v-if="book.faviconUrl" v-bind:src="'http://' + book.faviconUrl" />
             <img v-else-if="book.faviconMimeType" v-bind:src="'zim://' + book.id + '.favicon.meta'" />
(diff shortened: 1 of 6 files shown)


src/contentmanager.cpp
```diff
@@ -189,6 +188,22 @@ QString ContentManager::downloadBook(const QString &id)
     return QString::fromStdString(download->getDid());
 }
 
+void ContentManager::eraseBook(const QString& id)
+{
+    kiwix::Book book = mp_library->getBookById(id);
+    QString dirName = QString::fromUtf8(getDataDirectory().c_str());
+    QString fileSelection = QString::fromUtf8(book.getPath().c_str()).remove(0, getDataDirectory().size() + 1) + "*";
+    QDir dir(dirName, { fileSelection });
```
mgautierfr: Why the `{}` ?

src/contentmanager.cpp
```diff
@@ -189,6 +188,22 @@ QString ContentManager::downloadBook(const QString &id)
     return QString::fromStdString(download->getDid());
 }
 
+void ContentManager::eraseBook(const QString& id)
+{
+    kiwix::Book book = mp_library->getBookById(id);
+    QString dirName = QString::fromUtf8(getDataDirectory().c_str());
+    QString fileSelection = QString::fromUtf8(book.getPath().c_str()).remove(0, getDataDirectory().size() + 1) + "*";
```
mgautierfr: I find this line a bit difficult to read. Maybe split it in several lines ?

mgautierfr: We should be careful here.
If a user open a zim file in its own directory, the file is added to the library but it is not copied in the data directory.
Just removing the size of the data directory from the file path may lead to strange results.
Maybe we could use `getLastPathElement` from kiwix-lib to get the filename.
If the file is not in the data directory, the `dir.entryList()` will be empty and we will remove nothing. And so it should be ok.

src/contentmanager.cpp
```diff
@@ -189,6 +188,22 @@ QString ContentManager::downloadBook(const QString &id)
...
+    QDir dir(dirName, { fileSelection });
+    for(const QString& filename: dir.entryList()) {
+        dir.remove(filename);
+        qInfo() << "fichier "<<filename;
+    }
+    mp_library->removeBookFromLibraryById(id);
+    mp_library->save();
+    emit(mp_library->booksChanged());
+    emit(booksChanged());
```
mgautierfr: There is already a slot connect to `mp_library->booksChanged()` that emit `bookChanged()`. So no need to emit twice.

resources/texts/_contentManager.html
```diff
@@ -74,13 +75,16 @@
             downloadUpdaters[book.id] = setInterval(function() { getDownloadInfo(book.id); }, 1000);
           });
         },
-        eraseBook : function(book) {
-            contentManager.eraseBook(book.id);
+        eraseBook : function(id) {
```
mgautierfr: Maybe rename the function to `eraseBookById`.
There is no type checking in js. `eraseBook` imply to the argument is a book, not an id.
And nothing will warn us if we do something wrong... until the user try to erase the book.

PR #137, opened 2019-04-29
pause and resume button

### LINKED ISSUE #53: Add Stop/Pause button(s) to download
Probably right next to the download counter on the right of the screen?
image

jetownfeve21:
It needs that the PR that add pause and unpause functions for aria2 is merged (https://github.com/kiwix/kiwix-lib/pull/218/files).

It adds resume and pause button that call pause and unpause functions of aria2.

Fixes #53

### DIFF
--- a/resources/texts/_contentManager.html
+++ b/resources/texts/_contentManager.html
@@ -82,6 +82,8 @@ function init() {
         },
         downloadBook : function(book) {
           contentManager.downloadBook(book.id, function(did)  {
+            if (did.length == 0)
+                return;
             book.downloadId = did;
             downloadUpdaters[book.id] = setInterval(function() { getDownloadInfo(book.id); }, 1000);
           });
--- a/resources/texts/_contentManager.html
+++ b/resources/texts/_contentManager.html
@@ -48,6 +48,8 @@ function getDownloadInfo(id) {
     d = createDict(DOWNLOAD_KEYS, values);
     if (d.status == "completed") {
       clearInterval(downloadUpdaters[id]);
+      Vue.delete(app.downloads, id);
+      return;
     }
     Vue.set(app.downloads, id, createDict(DOWNLOAD_KEYS, values));
   });
@@ -86,6 +88,12 @@ function init() {
         },
         eraseBook : function(book) {
             contentManager.eraseBook(book.id);
+        },
+        pauseBook : function(book) {
+          contentManager.pauseBook(book.id);
+        },
+        resumeBook : function(book) {
+          contentManager.resumeBook(book.id);
         },
         displayedBooks : function(books, nb) {
             var a = books.slice(0, nb);
@@ -364,8 +372,10 @@ details:hover {
                 {{ niceBytes(downloads[book.id].completedLength) }} / {{ niceBytes(downloads[book.id].totalLength) }}
               </span>
             </template>
-            <button v-else-if="book.path" v-on:click="openBook(book)">Open</button>
-            <button v-else v-on:click="downloadBook(book)">Download</button>
+            <button v-if="book.path" v-on:click="openBook(book)">Open</button>
+            <button v-if="!book.path && !downloads[book.id]" v-on:click="downloadBook(book)">Download</button>
+            <button v-if="downloads[book.id] && downloads[book.id].status == 'active'" v-on:click="pauseBook(book)">Pause</button>
+            <button v-if="downloads[book.id] && downloads[book.id].status == 'paused'" v-on:click="resumeBook(book)">Resume</button>
           </span>
         </summary>
         <p class="content">
--- a/resources/texts/_contentManager.html
+++ b/resources/texts/_contentManager.html
@@ -95,6 +95,11 @@ function init() {
         resumeBook : function(book) {
           contentManager.resumeBook(book.id);
         },
+        cancelBook : function(book) {
+            contentManager.cancelBook(book.id);
+            clearInterval(downloadUpdaters[book.id]);
+            Vue.delete(app.downloads, book.id);
+        },
         displayedBooks : function(books, nb) {
             var a = books.slice(0, nb);
             return a;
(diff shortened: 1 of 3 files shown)


src/contentmanager.cpp
```diff
@@ -189,6 +188,63 @@ QString ContentManager::downloadBook(const QString &id)
     return QString::fromStdString(download->getDid());
 }
 
+void ContentManager::pauseBook(const QString& id)
+{
+    qInfo() << "pause";
+    auto& b = mp_library->getBookById(id);
+    kiwix::Download* download;
+    download = mp_downloader->getDownload(b.getDownloadId());
+    download->pauseDownload();
```
mgautierfr: `auto download = mp_downloader->getDownload(b.getDownloadId());` ?

The `auto` keyword avoid us the write the exact type. If at a moment we change the signature of `getDowload` to return something else (as a smart pointer), we don't have to change this code.
And I prefer to initialize variable directly in the same line it is declared. It may avoid uninitialized variable in the future (refactoring, wrong copy/paste, ...)

src/contentmanager.cpp
```diff
@@ -189,6 +189,24 @@ QString ContentManager::downloadBook(const QString &id)
     return QString::fromStdString(download->getDid());
 }
 
+void ContentManager::pauseBook(const QString& id)
+{
+    qInfo() << "pause";
+    auto& b = mp_library->getBookById(id);
+    kiwix::Download* download;
+    download = mp_downloader->getDownload(b.getDownloadId());
```
mgautierfr: We need to call `d->updateStatus(true);`
The `getDownload` method do not call it for us.

And as we are asynchrone here, the download may be complete before we get there (and because we are refreshing the download status only every second). But I'm not sure it is a problem, probably aria will do nothing and keep the download as completed and next refresh will show the correct status.

resources/texts/_contentManager.html
```diff
@@ -367,6 +373,12 @@
             <button v-else-if="book.path" v-on:click="openBook(book)">Open</button>
             <button v-else v-on:click="downloadBook(book)">Download</button>
           </span>
+          <span v-if="downloads[book.id] && downloads[book.id].status == 'active'" class="tablecell cell5">
+            <button v-on:click="pauseBook(book)">Pause</button>
+          </span>
+          <span v-if="downloads[book.id] && downloads[book.id].status == 'paused'" class="tablecell cell5">
+                <button v-on:click="resumeBook(book)">Resume</button>
+              </span>
```
mgautierfr: We should add those button **inside** the already created `span` (and even better, inside the `<template v-if="book.downloadId">`) .

jetownfeve21: image


The button moves according to the window's width, I think that's it's better that this kind of button have a fixed position.

image

Why don't create another span specific to all buttons ?

resources/texts/_contentManager.html
```diff
@@ -359,13 +374,20 @@
...
-            <button v-else v-on:click="downloadBook(book)">Download</button>
+          </span>
+          <span class="tablecell cell5">
+            <button v-if="book.path" v-on:click="openBook(book)">Open</button>
+            <button v-if="!book.path && !downloads[book.id]" v-on:click="downloadBook(book)">Download</button>
+            <button v-if="downloads[book.id] && downloads[book.id].status == 'active'" v-on:click="pauseBook(book)">Pause</button>
+            <button v-if="downloads[book.id] && downloads[book.id].status == 'paused'" v-on:click="resumeBook(book)">Resume</button>
+          </span>
+          <span v-if="downloads[book.id]" class="tablecell cell5">
+            <button v-on:click="cancelBook(book)">Cancel</button>
```
mgautierfr: The `div` "bookTable" is a table.It as 6 columns.
The header of the table is declared in the `header div`.
Each `span` represent one cell of the table. So it can't be two (or three) `span` with the class `cell5`.
Here, there are the span at line 376, the span at line 383 and (if a downoald is present) the span at line 389.
All the content of those three span should be merge in one.
Instead of having something like :
```html
<span class="tablecell cell5">
  <template v-if="downloads[book.id]" > ...</template>
</span>
<span class="tablecell cell5">
  <button v-if "book.path"> ...</button>
  <button v-if "!book.path"> ...</button>
  ...
</span>
<span v-if="downloads[book.id]" class="tablecell cell5">
  <button v-on:click="cancelBook(book)">Cancel</button>
</span>
```

We should have :
```html
<span class="tablecell cell5">
  <template v-if="downloads[book.id]" >
    ...
    <button v-on:click="cancelBook(book)">Cancel</button>
  </template>
  <button v-if "book.path"> ...</button>
  <button v-if "!book.path"> ...</button>
  ...
</span>
```

src/contentmanager.cpp
```diff
@@ -189,18 +193,51 @@ QString ContentManager::downloadBook(const QString &id)
...
     QDir dir(dirName, fileSelection);
     for(const QString& filename: dir.entryList()) {
         dir.remove(filename);
     }
+}
+
+void ContentManager::eraseBook(const QString& id)
+{
+    kiwix::Book book = mp_library->getBookById(id);
+    QString fileSelection = QString::fromUtf8(getLastPathElement(book.getPath()).c_str()) + "*";
```
mgautierfr: This is not a `fileSelection`, this is a `bookPath` or `fileToRemove` or ...

src/contentmanager.cpp
```diff
@@ -189,18 +193,51 @@ QString ContentManager::downloadBook(const QString &id)
...
+void ContentManager::cancelBook(const QString& id)
+{
+    auto& b = mp_library->getBookById(id);
+    auto download = mp_downloader->getDownload(b.getDownloadId());
+    QString fileSelection = QString::fromUtf8(getLastPathElement(download->getPath()).c_str()) + "*";
+    eraseBookFilesFromComputer(fileSelection);
     mp_library->removeBookFromLibraryById(id);
     mp_library->save();
     emit(mp_library->booksChanged());
+    download->cancelDownload();
```
mgautierfr: The cancel should probably be called before we remote the file from the fs or update the view.
Else, we could have a race condition where aria2 try to update the file on the fs while we are deleting it.

jetownfeve21:
can this branch be merged now that (https://github.com/kiwix/kiwix-lib/pull/218/files) has been merged ? @mgautierfr

mgautierfr:
I've just made a review before your comment :)

jetownfeve21:
I applied changes you requested @mgautierfr

mgautierfr:
Hey! It's good !

PR #138, opened 2019-05-06
Split contentmanager.html

jetownfeve21:
branch based on the stop-download branch

I don't really think that it's a clean option to split this. May I have your opinion ? @mgautierfr .


jetownfeve21:
I imported the css file with "<style src='qrc:///etc...'>" instead of <link rel="stylesheet" type="text/css" href="qrc:///css/_contentManager.css"/> and that broke everything.

PR #144, opened 2019-05-09
add alt+home shorcut and CTRL+mouse wheel zoom

### LINKED ISSUE #141: Implement ALT+Home keyboard shortcut to open the welcome page of a ZIM
Pressin ALT+Home should:
* open the welcome page of the ZIM file in the current content tab (if there is a tab with content open)
* do nothing otherwise

See also #58

### LINKED ISSUE #146: CTRL+mouse wheel should zoom (or dezoom)
Currently it does nothing

jetownfeve21:
When a webview is created, a QShortcut is set and connected to the slot
"openHomePage" that loads the home page thanks to the ZimId.

fixes #141 
fixes #146

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -270,6 +270,9 @@ void KiwixApp::createAction()
     connect(mpa_actions[RandomArticleAction], &QAction::triggered,
             this, [=]() { this->openRandomUrl(false); });
 
+    CREATE_ACTION(OpenHomePageAction, tr("Home page"));
+    SET_SHORTCUT(OpenHomePageAction, QKeySequence(Qt::ALT + Qt::Key_Home));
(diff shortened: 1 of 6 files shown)


src/mainmenu.cpp
```diff
@@ -12,6 +12,7 @@ MainMenu::MainMenu(QWidget *parent) :
 
     ADD_ACTION(KiwixServeAction);
     ADD_ACTION(RandomArticleAction);
+    ADD_ACTION(OpenHomePageAction);
```
mgautierfr: I would prefer to not add the button to the menu.

- It is not in the design
- The button is not is not align.

Capture d’écran du 2019-06-12 10-21-08

It is far simpler to simply add the action to the main window.

PR #145, opened 2019-05-10
kiwix-serve integration in kiwix-dekstop

### LINKED ISSUE #85: Integrate kiwix-serve to kiwix-desktop
So people can start/stop the kiwix-serve HTTP daemon in a few click (like in Kiwix Desktop 1.0)

jetownfeve21:
needs a merge of this pr before https://github.com/kiwix/kiwix-lib/pull/221

Display the "local kiwix server" option in the menu
This option opens a Qdialog with a "start kiwix server" button that set
the Qdialog's content and calls the "runServer()" method of Kiwixapp which
use its mp_server member that wraps kiwix-lib methods to execute the
kiwix-serve binary

Fixes #85

### DIFF
--- a/src/localkiwixserver.cpp
+++ b/src/localkiwixserver.cpp
@@ -1,0 +1,81 @@
+#include "localkiwixserver.h"
+#include "ui_localkiwixserver.h"
+#include "kiwixapp.h"
+#include <QDesktopServices>
+#include <QMessageBox>
+#include <thread>
+
+LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
+    QDialog(parent),
+    ui(new Ui::LocalKiwixServer)
+{
+    ui->setupUi(this);
+
+    QFile styleFile(":/css/localServer.css");
+    styleFile.open(QIODevice::ReadOnly);
+    auto byteContent = styleFile.readAll();
+    styleFile.close();
+    QString style(byteContent);
+    setStyleSheet(style);
+
+    mp_server = KiwixApp::instance()->getLocalServer();
+    m_port = mp_server->getPort();
+
+    connect(ui->KiwixServerButton, SIGNAL(clicked()), this, SLOT(runOrStopServer()));
+    connect(ui->OpenInBrowserButton, SIGNAL(clicked()), this, SLOT(openInBrowser()));
+    const QHostAddress &localhost = QHostAddress(QHostAddress::LocalHost);
+    for (const QHostAddress &address: QNetworkInterface::allAddresses()) {
+        if (address.protocol() == QAbstractSocket::IPv4Protocol && address != localhost) {
+            m_ipAddress = address.toString();
+            break;
+        }
+    }
+    ui->OpenInBrowserButton->setVisible(false);
+    ui->IpAddress->setVisible(false);
+}
+
+LocalKiwixServer::~LocalKiwixServer()
+{
+    delete ui;
+}
+
+void LocalKiwixServer::openInBrowser()
+{
+    QUrl url;
+    url.setScheme("http");
+    url.setHost(m_ipAddress);
+    url.setPort(m_port);
+    QDesktopServices::openUrl(url);
+}
+
+void LocalKiwixServer::runOrStopServer()
+{
+    if (!m_active) {
+        mp_server->run();
+        ui->IpAddress->setText(m_ipAddress + ":" + QString::number(m_port));
+        std::this_thread::sleep_for(std::chrono::milliseconds(500));
+        if (!mp_server->isRunning()) {
+            QMessageBox messageBox;
+            messageBox.critical(0,"Error","An error has occured !");
+            return;
+        }
+        m_active = true;
+    } else {
+        mp_server->shutDown();
+        m_active = false;
+    }
+
+    if (m_active) {
+        ui->KiwixServerButton->setText("Stop Kiwix Server");
+        QUrl url("qrc:/texts/kiwixServerRunningText.html");
+        ui->KiwixServerText->setSource(url);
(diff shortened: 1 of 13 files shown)


mgautierfr:
It seems that we never check that the server is actually running.
So if for some reason we cannot launch the server, we still propose to the user to connect to it but it doesn't work. We should check that the server is correctly running and display a message if we cannot.

jetownfeve21:
Something like what is done with aria2 ?
https://github.com/kiwix/kiwix-desktop/blob/af3d00d20078612e66de574370f67992f37b5fa9/src/kiwixapp.cpp#L15-L26

mgautierfr:
This work only for aria because we try to connect to aria and throw an exception if we cannot connect.
(https://github.com/kiwix/kiwix-lib/blob/master/src/aria2.cpp#L83-L94)

We should use the isRunning method of the subprocess class.

jetownfeve21:
need this PR https://github.com/kiwix/kiwix-lib/pull/233 to be merged before

src/localkiwixserver.cpp
```diff
@@ -0,0 +1,71 @@
...
+        return;
+    }
+    m_active = !m_active;
+    if (m_active) {
+        KiwixApp::instance()->runServer();
+        ui->KiwixServerButton->setText("Stop Kiwix Server");
+        QUrl url("qrc:/texts/kiwixServerRunningText.html");
+        ui->KiwixServerText->setSource(url);
+        ui->OpenInBrowserButton->setVisible(true);
+        ui->IpAddress->setVisible(true);
```
mgautierfr: There is no information about the port visible anywhere.

src/localkiwixserver.cpp
```diff
@@ -0,0 +1,71 @@
...
+void LocalKiwixServer::runOrStopServer()
+{
+    if (KiwixApp::instance()->isLocalServerRunning()) {
+        QMessageBox messageBox;
+        messageBox.critical(0,"Error","An error has occured !");
+        return;
+    }
+    m_active = !m_active;
+    if (m_active) {
+        KiwixApp::instance()->runServer();
```
mgautierfr: We should check here that we have correctly launch the server.
If kiwix-serve is not present, the serve is not launch (normal) but everything seems to work correctly. The "open in browser" just open the browser using a address where nothing is served.

src/localkiwixserver.cpp
```diff
@@ -0,0 +1,71 @@
...
+    QDesktopServices::openUrl(url);
+}
+
+void LocalKiwixServer::runOrStopServer()
+{
+    if (KiwixApp::instance()->isLocalServerRunning()) {
+        QMessageBox messageBox;
+        messageBox.critical(0,"Error","An error has occured !");
+        return;
+    }
```
mgautierfr: Why this test ? We cannot stop the server now.

src/localkiwixserver.cpp
kelson42: @jetownfeve21 This is really important, please secure it is similar (or better) than with kiwix-xulrunner.

kelson42: @jetownfeve21 The behaviour if the kiwix-serve port is busy should be also clear.

jetownfeve21:
@mgautierfr I think that there is a problem with the UnixImpl::isRunning() method : 
https://github.com/kiwix/kiwix-lib/blob/e740a511c6ebf61c055e131dbbf1cfaf2858bc2e/src/subprocess_unix.cpp#L61-L78

In the case where execvp fails I'm not sure that the parent catchs the exit -1 of the child and then set the m_running member to false

mgautierfr:
No, it is ok.
The child is created with the fork. So if the execvp fails, the child exit and the parent is notified (it catches the signal using `waitpid` in `waitForPID`.
There was a bug about `waitpid` but I have fixed it here (https://github.com/kiwix/kiwix-lib/pull/232)

jetownfeve21:
I removed the kiwix-serve binary to test errors but when I try to launch the local kiwix-server with kiwix-desktop the isRunning method returns me true

mgautierfr:
It seems that the child process is never exited. It seems that a QT atexit handler blocks somewhere because of mutex and the process never exits.

It is strange because when I tested it yesterday the process correctly exited.

mgautierfr:
This should fix the problem https://github.com/kiwix/kiwix-lib/pull/235

jetownfeve21:
even with this change https://github.com/kiwix/kiwix-lib/pull/235 the child process is too "slow" to exit before the parent process use the isRunning method. 
If I add a sleep of 500 milliseconds just after the launch of the server to let the child process exits if he needs is it ok or too dirty ?

mgautierfr:
It is ok :)

jetownfeve21:
need this PR to be merged https://github.com/kiwix/kiwix-lib/pull/236

kelson42:
@jetownfeve21 @mgautierfr What is the status here?

mgautierfr:
Last needed PR in kiwix-lib (kiwix/kiwix-lib#236) has been merged.
I'm agree with the code of this PR.

> The behaviour if the kiwix-serve port is busy should be also clear.

@kelson42, please not that the port will be displayed but not configurable. We need to implement a configuration system / display page.
If the port is busy, kiwix-serve will fail to launch and the kiwix-desktop will display an error message (a generic one because we cannot (for now?) know why).

We will still need PR kiwix/kiwix-build#342 to be merged.

kelson42:
@mgautierfr Good, at least the behaviour is clear for the end user. We need then to implement #200

mgautierfr:
Time to merge this.

PR #155, opened 2019-05-15
add a confirmation message before deleting

jetownfeve21:


mgautierfr:
Would be nice to also have a confirmation message on "cancel download".

PR #158, opened 2019-05-21
don't open new tab

### LINKED ISSUE #81: Stay in current tab when opening a new article
I think it would be better to stay in the current tab when you search for a new article or try a random article. At least I am not used to new tabs opening all the time.

However, if you make this change, then you also have to add a small "tab" to the right of the last article tab with a plus sign or something for creating a new tab, as it is in many browsers and other programs.

jetownfeve21:
Fix #81


kelson42:
@jetownfeve21 Could you please also fix this behaviour for the random article (see ticket)?

Issue #80, opened 2018-12-12
Open articles by pressing enter


jargony:
If you have typed a correct article title it should be possible to press enter to load the page. If I write the full title, e.g. "united states", and then press enter, I would assume the article will open. It feels like a strange behavior that the search bar is just emptied and one stays on the old page. It is often obvious what the title is and it is much faster to type the entire title then to wait for the suggestions.

kelson42:
Ticket is probably related to the implementation of the fulltext search result page. On 1.0, the behaviour was, that if the article can be found directly then it open it, otherwise it opens the fulltext result page.

jargony:
That feature sounds like what I was thinking about. Like how online wikipedia behaves too.

kelson42:
I have open a separate ticket for the fullltext search https://github.com/kiwix/kiwix-desktop/issues/159

PR #161, opened 2019-05-22
click on a popup's suggestion opens article

jetownfeve21:
#80

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -73,6 +73,7 @@ SearchBar::SearchBar(QWidget *parent) :
 #endif
     connect(KiwixApp::instance(), &KiwixApp::currentTitleChanged,
             this, &SearchBar::on_currentTitleChanged);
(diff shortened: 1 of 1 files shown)


mgautierfr:
I'm not sure it fix the issue #80 
The problem is that if a user enter a text and press enter, the page is not opened.
While opening the page if the user clicks on the suggestion is a good thing, it seems not related to #80.

From what I see, it seems that the page is opened on "enter" if the suggestion popup has the focus (see the small border inside the popup) and the search entry is cleaned if the popup has no focus.

Depending of how the user navigate in the pages (two searchs in a row vs a search, go the to open page, do a second search), the focus is correctly set or not.

It may be related to https://github.com/kiwix/kiwix-desktop/blob/master/src/searchbar.cpp#L67-L70 I did take the time to investigate further.

jetownfeve21:
The improvement of this PR is integrated in this other PR https://github.com/kiwix/kiwix-desktop/pull/166 , so this PR  is now useless

PR #164, opened 2019-05-27
choose the library.xml file in the binary's directory if it exists

jetownfeve21:
wip

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -27,6 +27,7 @@ kiwix::Downloader* createDownloader() {
 
 KiwixApp::KiwixApp(int& argc, char *argv[])
     : QApplication(argc, argv),
+      m_libraryDirectory(findLibraryDirectory()),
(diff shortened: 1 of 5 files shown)


jetownfeve21:
@mgautierfr 
I check if there is a library.xml in the binary's directory and set a member with that path instead of reading many library.xml files to avoid the save of a non corresponding file at the exit of the app. Is it good ?

src/kiwixapp.cpp
```diff
@@ -109,6 +110,17 @@ KiwixApp::~KiwixApp()
     delete mp_mainWindow;
 }
 
+QString KiwixApp::findLibraryPath()
+{
+    QString currentPathLibrary = QString::fromStdString(appendToDirectory(getCurrentDirectory(),"library.xml"));
```
mgautierfr: This should be `getExecutablePath`, not `getCurrentDirectory`.

PR #166, opened 2019-05-31
Auto completion search bar

### LINKED ISSUE #80: Open articles by pressing enter

jetownfeve21:
fix #80 

open article by pressing enter : 
Set the completer case insensitive to match with the suggestion's list
Re-use the connect with openCompletion method that works fine to handle mouse
click and enter key press to open article
Remove the openTitle method that is no longer usefull.

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -65,13 +65,8 @@ SearchBar::SearchBar(QWidget *parent) :
     m_completer.setCaseSensitivity(Qt::CaseInsensitive);
     setCompleter(&m_completer);
     connect(this, &QLineEdit::textEdited, this, &SearchBar::updateCompletion);
-#if 0 //The `activated` signal seems to not be emitted if user navigate in the page.
-      // Something is broken here .. :/
     connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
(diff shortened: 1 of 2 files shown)


src/searchbar.cpp
```diff
@@ -63,16 +63,18 @@ SearchBar::SearchBar(QWidget *parent) :
     setPlaceholderText(tr("Search"));
     m_completer.setCompletionMode(QCompleter::UnfilteredPopupCompletion);
     setCompleter(&m_completer);
+    m_completer.setCaseSensitivity(Qt::CaseInsensitive);
```
jetownfeve21: @mgautierfr 
I think that it's easier to set the QCompleter handle the focus on the right row setting the case insensitive instead of rewrite the focus of the different widgets.
Even if the app has a custom suggestion's list engine, say to the QCompleter to match with the case insensitive doesn't seem inconsistent to me

mgautierfr:
Good catch with the "sensitive" setting.

PR #169, opened 2019-06-06
add new tab button with basic css

### LINKED ISSUE #36: New tab button is missing.

jetownfeve21:
add a new tab with a button inside at the start of the app. When this button
is pressed, it inserts a new tab just before this tab.
To avoid that this tab/button can be selected when a tab is closed, the
SelectionBehaviorOnRemove is set to select the left tab if it's the most
right before the tab/button and the right if not

Fix #36

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -4,6 +4,7 @@
 #include <QAction>
 #include <QTimer>
 #include <QWebEnginePage>
+#include <QToolButton>
 
 #define QUITIFNULL(VIEW) if (nullptr==(VIEW)) { return; }
 #define QUITIFNOTCURRENT(VIEW) if((VIEW)!=currentWidget()) {return;}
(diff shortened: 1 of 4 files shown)


src/tabbar.cpp
```diff
@@ -77,6 +78,19 @@ void TabBar::setContentManagerView(ContentManagerView* view)
     setTabButton(0, RightSide, nullptr);
 }
 
+void TabBar::setNewTabButton()
+{
+    mp_stackedWidget->addWidget(new QWebEngineView());
+    QToolButton *tb = new QToolButton();
+    tb->setText("+");
+    connect(tb, &QAbstractButton::pressed, this, [=]() {  KiwixApp::instance()->getAction(KiwixApp::NewTabAction)->activate(QAction::Trigger); });
```
mgautierfr: Maybe the `setDefaultAction` of a `QToolButton` could be used here.

src/tabbar.cpp
```diff
@@ -77,6 +78,19 @@ void TabBar::setContentManagerView(ContentManagerView* view)
     setTabButton(0, RightSide, nullptr);
 }
 
+void TabBar::setNewTabButton()
+{
+    mp_stackedWidget->addWidget(new QWebEngineView());
```
mgautierfr: If the "tab +"  will be never selected, we don't need to add  a widget.

jetownfeve21:
With a button after the tabbar, we will have to handle its position to place it just to the right of the opened tabs isn't it ?

mgautierfr:
We are good. Having a separated button would make more sense but it works well as it is now.

PR #173, opened 2019-06-07
Fix random crash

### LINKED ISSUE #160: kiwix-desktop sometimes crashes when opening a zim file from file explorer

jetownfeve21:
WIP

fix #160

### DIFF
--- a/src/urlschemehandler.cpp
+++ b/src/urlschemehandler.cpp
@@ -49,7 +49,7 @@ UrlSchemeHandler::handleContentRequest(QWebEngineUrlRequestJob *request)
     BlobBuffer* buffer = new BlobBuffer(entry.getBlob());
     auto mimeType = QByteArray::fromStdString(entry.getMimetype());
     mimeType = mimeType.split(';')[0];
-    connect(buffer, &QIODevice::aboutToClose, buffer, &QObject::deleteLater);
+    connect(request, &QObject::destroyed, buffer, &QObject::deleteLater);
     request->reply(mimeType, buffer);
 }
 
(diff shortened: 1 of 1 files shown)


src/readinglistbar.cpp
```diff
@@ -55,5 +55,5 @@ void ReadingListBar::on_itemActivated(QListWidgetItem* item)
     url.setScheme("zim");
     url.setHost(QString::fromStdString(bookmark.getBookId())+".zim");
     url.setPath(QString::fromStdString(bookmark.getUrl()));
-    KiwixApp::instance()->openUrl(url);
+    KiwixApp::instance()->openUrl(url, false);
```
jetownfeve21: only to facilitate all the manipulations to reproduce the crash

src/urlschemehandler.cpp
```diff
@@ -49,7 +49,8 @@ UrlSchemeHandler::handleContentRequest(QWebEngineUrlRequestJob *request)
     BlobBuffer* buffer = new BlobBuffer(entry.getBlob());
     auto mimeType = QByteArray::fromStdString(entry.getMimetype());
     mimeType = mimeType.split(';')[0];
-    connect(buffer, &QIODevice::aboutToClose, buffer, &QObject::deleteLater);
+    // connect(buffer, &QIODevice::aboutToClose, buffer, &QObject::deleteLater);
+    connect(request, &QObject::destroyed, buffer, &QObject::deleteLater);
```
jetownfeve21: it seems that the deletion of the buffer is the cause of the random crash
According to the doc, it can be deleted like this.

image

https://doc.qt.io/qt-5/qwebengineurlrequestjob.html#reply

src/webview.cpp
```diff
@@ -14,10 +14,21 @@ WebView::WebView(QWidget *parent)
...
+    connect(this, &QWebEngineView::loadFinished, this, [=](bool test) {
+        if (test) {
+            qInfo() << "good";
+        } else {
+            qInfo() << "not good";
+        }
+    });
+    connect(this, &QWebEngineView::loadStarted, this, [=]() {
+            qInfo() << "load start";
+    });
```
jetownfeve21: only debug

mgautierfr:
Good catch

PR #174, opened 2019-06-07
add alert if there is not enough storage to download

### LINKED ISSUE #171: Software should check about free disk space
Before downloading...

jetownfeve21:
Check the available storage before the start of a download thanks to
QStorageInfo::bytesAvailable(), if the book size is bigger than this value
an alert message appears.

fix #171


mgautierfr:
I am not a big fan of returning a string as an error, but we don't have a lot to do here.

Issue #119, opened 2019-03-30
Category "other" does not return any result


kelson42:
It should return all the other which are in none of the other categories.

mgautierfr:
We should make this list dynamic and get the real categories from a request to the server.
There is no real sense to "other" when you filter content by tags.

PR #175, opened 2019-06-13
fix local category other

jetownfeve21:
need this pr to be merged https://github.com/kiwix/kiwix-lib/pull/229

Use a new api Library::listBookIds. If the "other" category is selected, set
the "other" parameter to true and send the complete list of category that is
displayed in the sidemenu.

fix kiwix/kiwix-desktop#119

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -2,6 +2,7 @@
 #include "kiwixapp.h"
 
 #include <kiwix/manager.h>
+#include "kiwixapp.h"
 
 #include <QtDebug>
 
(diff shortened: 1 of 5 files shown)


kelson42:
This PR should be close IMO, it does not do what it should do.

src/library.cpp
```diff
@@ -94,15 +95,24 @@ QStringList Library::listBookIds(const QString &query, const QString &categoryFi
...
-    if (categoryFilter != "all") {
+    if (categoryFilter != "all" && categoryFilter != "other") {
         tags.push_back(categoryFilter.toStdString());
     }
+    bool other = false;
+    if (categoryFilter == "other") {
+        other = true;
+        auto categoryList = KiwixApp::instance()->getMainWindow()->getSideContentManager()->getCategoryList();
+        for (auto& category: categoryList) {
+            tags.push_back(category.toLower().toStdString());
```
mgautierfr: We should not add the tag "other" here. Else we may never be able to have book that have the "other" tag.

jetownfeve21:
@mgautierfr 
I started to implement the filtering with your new api, it works great.
For the remote side (kiwix-tools), which list of category do I use for the rejected tags ?

mgautierfr:
> For the remote side (kiwix-tools), which list of category do I use for the rejected tags ?

The same :)
But we need to update the search API of kiwix-serve to list books. We should add a new request argument (`notag`) that will be used by kiwix-serve to list books that have no tags listed in `notag` (using the `excludeTags` of Filter) See https://github.com/kiwix/kiwix-tools/blob/master/src/server/kiwix-serve.cpp#L700-L729 or better with the new api (https://github.com/kiwix/kiwix-tools/pull/302)

jetownfeve21:
@mgautierfr 
I prefer to make another PR for the remote side

PR #178, opened 2019-06-17
Fix fullscreen icon position in menu

### LINKED ISSUE #176: Bad "fullscreen" position icon in contextual menu
The icon is behind the text, see screenshot:
scr

Tested on GNU/Linux with beta5

jetownfeve21:
The "setCheckable(true)" method adds a check mark in the menu that shifts
the fullscreen icon.
To avoid this behaviour, the action is not checkable anymore. When it is
triggered, it uses the iconText property to display either or those.

fix #176

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -319,14 +319,15 @@ void KiwixApp::createAction()
 
     CREATE_ACTION_ICON(ToggleFullscreenAction, "full-screen-enter", tr("Set fullScreen"));
     SET_SHORTCUT(ToggleFullscreenAction, QKeySequence::FullScreen);
-    connect(mpa_actions[ToggleFullscreenAction], &QAction::toggled,
-            this, [=](bool checked) {
+    connect(mpa_actions[ToggleFullscreenAction], &QAction::triggered,
+            this, [=]() {
         auto action = mpa_actions[ToggleFullscreenAction];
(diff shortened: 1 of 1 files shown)


jetownfeve21:
I use the IconText property like a boolean because I didn't find how to hide the check mark.

kelson42:
@jetownfeve21 Thank you for having spontaneously a look at this. That said the solution is a bit *hacky*... not sure it will pass the review ;)

mgautierfr:
> not sure it will pass the review ;)

Yes :)

I think this is a issue in qt 5.12 itself (see https://bugreports.qt.io/browse/QTBUG-74655)
I recommend to do nothing on our side to fix this and wait that qt fixes the bug (or that we move back to 5.11).

Independently of the issue, it seems to me that the code given in the bug example program does better to change the check/uncheck status icon. We may use it (without urgency, our code works and it will not fix the issue anyway)

kelson42:
@mgautierfr Good to me. Lets wait a few months to see if the problem will be fixed on Qt side. Don't forget to vote/follow on the upstream ticket if you can.

kelson42:
The is a bug in QT, see https://bugreports.qt.io/browse/QTBUG-74655). This will should be fixed in Qt 5.12.5 to be release end of 08/2019. Closing the PR.

PR #180, opened 2019-06-18
Fix bug signal activated not emitted

jetownfeve21:
Qt seems to disconnect every signal handlers (signal emitted by QCompleter)
of QLineEdit (the SearchBar) every time the QLineEdit loses the focus,
so it has to be connected every time it gains the focus with the mouse.

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -73,8 +73,6 @@ SearchBar::SearchBar(QWidget *parent) :
     m_completer.popup()->setStyleSheet(style);
 
     connect(this, &QLineEdit::textEdited, this, &SearchBar::updateCompletion);
-    connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
-            this, &SearchBar::openCompletion);
     connect(KiwixApp::instance(), &KiwixApp::currentTitleChanged,
             this, &SearchBar::on_currentTitleChanged);
 }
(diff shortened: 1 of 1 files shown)


jetownfeve21:
@mgautierfr 
The bug you have discovered about the QCompleter's signal "activated" here
https://github.com/kiwix/kiwix-desktop/blob/18d123bcfbbb8ae6b7de68061d10fb191cbf9fe2/src/searchbar.cpp#L68-L74

I thought that it was fixed but I was wrong. As it says here https://stackoverflow.com/questions/47102602/qcompleteractivated-disconnected-after-losing-and-gaining-focus QLineEdit seems to disconnect its signal handlers every time it loses the focus.

This PR has to be reviewed/merged/integrated to the release version as soon as possible to make the searchbar work again

src/searchbar.cpp
```diff
@@ -87,8 +85,11 @@ void SearchBar::on_currentTitleChanged(const QString& title)
 
 void SearchBar::focusInEvent( QFocusEvent* event)
 {
-    if (event->reason() == Qt::MouseFocusReason)
+    if (event->reason() == Qt::MouseFocusReason) {
         clear();
+        connect(&m_completer, static_cast<void(QCompleter::*)(const QModelIndex &)>(&QCompleter::activated),
```
mgautierfr: Why changing the cast ?

jetownfeve21: In the Qt documentation, it's the cast they used
image

mgautierfr: There are different version of connect.
The one used in the documentation takes a function (https://doc.qt.io/qt-5/qobject.html#connect-1)
I prefer to use the one that use a "PointerToMemberFunction" (https://doc.qt.io/qt-5/qobject.html#connect-3). I find the code easier to read.
I'm not sure of how we should strictly follow the note (ie what happens if we don't use the function pointer syntax ?) 
What is the documentation you use ? https://doc.qt.io/qt-5/qcompleter.html#activated-1 gives another example.

jetownfeve21: Indeed it was the qt 4.8

mgautierfr:
We are good

PR #181, opened 2019-06-19
remove category wikispecies

### LINKED ISSUE #113: wikispecies category filtering fails
I get no result

jetownfeve21:
fix #113

### DIFF
--- a/src/contentmanagerside.cpp
+++ b/src/contentmanagerside.cpp
@@ -205,19 +205,19 @@ ContentManagerSide::ContentManagerSide(QWidget *parent) :
 
     for (auto category: {
          "All",
-         "Gutenberg",
+         "Gutenberg", // Gutenberg book has wrong tag
          "Other",
-         "Phet",
+         "Phet", // Phet books have no tags
          "Psiram",
          "Stack Exchange",
-         "Ted",
+         "Ted",  // Ted books have wrong tags
          "Vikidia",
-         "Wikibooks",
+         "Wikibooks", // wikibooks have no tags
          "Wikinews",
          "Wikipedia",
          "Wikiquote",
          "Wikisource",
-         "Wikispecies",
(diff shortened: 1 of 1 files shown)


kelson42:
@jetownfeve21 The ticket ask to remove "Wikispecies" nothing else. The other one should display multiple content. The problem here is why it does not!

jetownfeve21:
You asked me to remove Wikispecies because it displays nothing, this is due to a wrong tag assignment, and the others that I removed are in the same case. I don't know how the tags are assigned

kelson42:
@jetownfeve21 The main reason to remove the Wikispecies category is not because this is empty (this is a bug) but because even if it would work it would display only one content... so better leave it in the "Other" category. For the the other category you have removed in this PR, this is not the case, they should be kept and display multiple items. The question is why not? What do you mean exactly with "wrong tag assignment"? Do the ZIM metadata are wrong? If "yes" how are they wrong?

jetownfeve21:
Gutenberg has only one content and there is no tags to clearly identify it as a gutenberg content
image

jetownfeve21:
Same for the phet category, there is no tags at all
image

jetownfeve21:
Ted has content but the files don't have tags
image

jetownfeve21:
No tags for wikibooks
image

jetownfeve21:
Wikispecies content have the tag "wikimedia"
image

mgautierfr:
I'm partisan to fix the tags on the zim files. Here we are just adding special case in the client to fix bugs in another that our projects.
I prefer to create a small tool changing the tags in zim files instead of adding hacks in the client.

mgautierfr:
@kelson42 This PR just remove wikispecies category.
Other categories are keep even if they return no result.
Are we agree ? If yes, feel free to merge it :)

PR #183, opened 2019-06-19
Pause book's download while dialog to cancel is open

### LINKED ISSUE #192: Cancel dialog doesn't pause the download
When the download of a book finishes while the cancel's dialog is opened
and then that the user confirms the deletion of the book, the app crashes

jetownfeve21:
When the download of a book finishes while the cancel's dialog is opened
and then that the user confirms the deletion of the book, the app crashes
because aria2 can't find the active download (which is already complete).
Pause the download before opening the dialog and check the download's status
before cancel it avoid this crash.

fix #192

### DIFF
--- a/resources/js/_contentManager.js
+++ b/resources/js/_contentManager.js
@@ -100,10 +100,13 @@ function init() {
             }
         },
         cancelBook : function(book) {
+            contentManager.pauseBook(book.id);
             if (confirm("Are you sure you want to abort the download of '" + book.title + "' ?")) {
                 contentManager.cancelBook(book.id);
                 clearInterval(downloadUpdaters[book.id]);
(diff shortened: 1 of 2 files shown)


kelson42:
@jetownfeve21 Thank you for this good catch and having explained the problem properly. In general I prefer to see a proper bug ticket open (with the description of the problem) and then a linked PR. This is clearer.

mgautierfr:
> In general I prefer to see a proper bug ticket open (with the description of the problem) and then a linked PR. This is clearer.

I do not agree with this. This is a multiplication of actions and informations in different places.
If the PR explains correctly what it fixes, this is enough for me.

But, this is a global discussion, not specific for this PR.
The PR is good. I'm merging it.

PR #184, opened 2019-06-20
Refresh only one book instead of all the library

### LINKED ISSUE #134: remote library isn't refreshed after starting many downloads
After starting a download with "stack" set with the search bar, there in no books displayed anymore.
But it doesn't appened always, maybe it depends on the number of books to display, or number of simultaneous download.

image

jetownfeve21:
These changes allow many improvements :
- they fix the infinite refresh of the library due to the multiple
"bookChanged" signal emitted when several operations on books were
performed simultaneously (launch several downloads) fix #134 
- the scroll isn't reset to the top of the list when starting a download at
the bottom of the list
- (better optimization ?)


mgautierfr:
We are good

PR #199, opened 2019-06-26
Removed RPATH from compiled binaries.

xvitaly:
Rpath is not needed to run installed binaries. Most of GNU/Linux distributions strictly forbid it. Already removed from Kiwix Desktop's dependencies.

See also:

 * https://lintian.debian.org/tags/binary-or-shlib-defines-rpath.html
 * https://wiki.debian.org/RpathIssue
 * https://docs.fedoraproject.org/en-US/packaging-guidelines/#_beware_of_rpath

### DIFF
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -105,8 +105,6 @@ static {
 }
 
 unix {
-  QMAKE_LFLAGS += "-Wl,-rpath,\'\$$ORIGIN/../lib64\ '"
-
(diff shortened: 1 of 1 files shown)


kelson42:
@mgautierfr I thought we don't use rpath anymore?!

mgautierfr:
> @mgautierfr I thought we don't use rpath anymore?!

In other projects. I miss kiwix-desktop.

Thanks @xvitaly

Issue #193, opened 2019-06-24
Rework how we handle categories.


mgautierfr:
I dislike how we handle the categories.
There is no real separation between what is displayed and what is done.
So we end with silly and special case hacks to handle "all", "other", "stack exchange".
And we can be pretty sure that other special case will appear (without speaking of translation problem)

We should think how we handle this and make a better system.

kelson42:
@mgautierfr You have proposed a dynamic system (not sure if we have a system here) and I think we could think about that... but what I don't understand is why we don't have now - in the current system - a simple mapping table with tag <-> label key? That seems simple and that is what is done in Kiwix iOS.

mgautierfr:
>  but what I don't understand is why we don't have now - in the current system - a simple mapping table with tag <-> label key

Because we have to implement it. And maybe because we never finish this discussion https://github.com/kiwix/kiwix-desktop/issues/118
And we decided to go "the simpler (short therm) way"

rgaudin:
Please follow-up on [kiwix-tools#317](https://github.com/kiwix/kiwix-tools/issues/317)

PR #201, opened 2019-06-28
rename "Stack Exchange" to "Stackexchange" when getting category list

jetownfeve21:
rename "Stack Exchange" to "Stackexchange" when getting category list otherwise the "other" category include stackexchange file


kelson42:
I hardly believe this will fix the problem correctly. Here the problem is that we have a label which has to be different from the tag.

mgautierfr:
See #193

PR #202, opened 2019-06-28
add "notag" parameter to the search api

jetownfeve21:
The notag parameter is filled with all categories from the category list
separated by a ";"

Need this https://github.com/kiwix/kiwix-tools/pull/305 to work.

### DIFF
--- a/src/opdsrequestmanager.cpp
+++ b/src/opdsrequestmanager.cpp
@@ -1,4 +1,5 @@
 #include "opdsrequestmanager.h"
+#include "kiwixapp.h"
 
 OpdsRequestManager::OpdsRequestManager()
 : mp_reply(nullptr)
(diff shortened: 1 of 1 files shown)


src/opdsrequestmanager.cpp
```diff
@@ -14,9 +15,23 @@ void OpdsRequestManager::doUpdate(const QString& currentLanguage, const QString&
...
+
+    if (categoryFilter == "other") {
+        auto categoryList = KiwixApp::instance()->getMainWindow()->getSideContentManager()->getCategoryList();
+        QString excludeTags;
+        for (int i = 0; i < categoryList.size(); i++) {
+            if (categoryList[i] != "Other") {
+                excludeTags += (i != 0) ? ";" : "";
+                excludeTags += categoryList[i].toLower();
+            }
+        }
```
mgautierfr: `QStringList` (the type of categoryList) already has a function to join string together (https://doc.qt.io/qt-5/qstringlist.html#join)

`excludeTags = categoryList.join(";");` should work

PR #216, opened 2019-07-31
Fix compilation instructions

kelson42:
This is what I needed to do to get things working on my Ubuntu 18.04.

### DIFF
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -151,6 +151,6 @@ RESOURCES += \
     resources/style.qrc
 
 unix {
-    system($$QMAKE_LUPDATE -locations relative -no-ui-lines $$_PRO_FILE_)
-    system($$QMAKE_LRELEASE $$_PRO_FILE_)
+    system(lupdate -locations relative -no-ui-lines $$_PRO_FILE_)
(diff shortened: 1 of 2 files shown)


kiwix-desktop.pro
```diff
@@ -150,6 +150,9 @@ RESOURCES += \
     resources/contentmanager.qrc \
     resources/style.qrc
 
+QMAKE_LUPDATE = /usr/bin/lupdate
+QMAKE_LRELEASE = /usr/bin/lrelease
```
mgautierfr: I don't know why we use the variable QMAKE_LUPDATE and QMAKE_LRELEASE.
This come from this PR #25 and as @Ordissimo said itself, it is more accustomed to gettext than Qt. As I am.

If we don't need it (define it just to use it later), it may be better to directly use the binary name late.
ie : `system(lupdate -locations relative ..)`
And this way, we would be more tolerant to tools install elsewhere than in `/usr/bin`.

kelson42: OK. Had somehow the expectation to improve that line to apply the hardcoded path only if undefined... but not sure about which language is exactly used in the `.pro` file.

PR #220, opened 2019-08-06
delete a book closes all tabs opened of it

### LINKED ISSUE #210: Deleting a ZIM file should close tabs open of this ZIM file
Currently the tabs previously open on this ZIM file are still open.

jetownfeve21:
Before deleting a book, check all the tabs opened (except the first wich is
the home tab and the last wich is the "new tab" button) and close those who
have the same id of the book.

fix #210

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -218,6 +218,16 @@ void ContentManager::eraseBookFilesFromComputer(const QString fileToRemove)
 
 void ContentManager::eraseBook(const QString& id)
 {
+    auto tabBar = KiwixApp::instance()->getTabWidget();
+    int i = 1;
+    while (i < tabBar->count() - 1) {
+        WebView* webView = tabBar->widget(i);
+        if (webView->zimId() == id) {
+            tabBar->closeTab(i);
+        } else {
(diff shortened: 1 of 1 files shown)


src/contentmanager.cpp
```diff
@@ -213,6 +213,15 @@ void ContentManager::eraseBookFilesFromComputer(const QString fileToRemove)
 
 void ContentManager::eraseBook(const QString& id)
 {
+    auto tabBar = KiwixApp::instance()->getTabWidget();
+    for (int i = 1; i < tabBar->count() - 1; i++)
+    {
+        WebView* webView = tabBar->widget(i);
+        if (webView->zimId() == id) {
+            tabBar->closeTab(i);
+            i--;
```
mgautierfr: Hum, I understand why you decrement `i` (the `tabBar` count change when you remove a tab).
But I would prefer a while here (instead of decrement and increment).
```
int i = 1;
while (i < tasBar->count() - 1) {
  WebView* webView = tabBar->widget(i);
  if (webView->zimId() == id) {
    tabBar->closeTab(i);
  } else {
    i++;
  }
}
```

PR #223, opened 2019-08-08
scroll wheel click closes tab

### LINKED ISSUE #79: Close tabs with the scroll wheel
It should be possible to close tabs with the scroll wheel. It is possible in most browsers and many other programs so it feels natural to do it.

jetownfeve21:
add an event listener on the mouse click event, if it's the middle button
(the scroll wheel click) it closes the tab on which the cursor is.
In the closeTab() method, add a security to prevent closing the "new tab
button" tab.

fix #79


mgautierfr:
The security to prevent closing the "new tab button" could have been in a separated commit. But that's ok.

PR #228, opened 2019-08-15
Add save page option

### LINKED ISSUE #77: Implement "Save page"
It is not possible to save pages in Kiwix now. When you click "Save page" nothing happens.

jetownfeve21:
open a file dialog to choose the path of the file which is used by the
save() method. There is no css and images.

fix #77

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -156,6 +156,16 @@ void KiwixApp::openZimFile(const QString &zimfile)
     openUrl(QUrl("zim://"+zimId+".zim/"));
 }
 
+void KiwixApp::savePage()
+{
+    if(!mp_tabWidget->currentWidget())
+        return;
+    QString fileName = QFileDialog::getSaveFileName(getMainWindow(),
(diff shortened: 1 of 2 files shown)


mgautierfr:
You are using the flag `QWebEngineDownloadItem::CompleteHtmlSaveFormat` when you save the page and the QT documentation says that it will save the resources.
Does the commit's comment ("There is no css and images.") is still valid ?

mgautierfr:
I think that the main problem with the images and the resources is that we use the `zim://` as url scheme to intercept the request and serve the content from the zim file. (And forward the request to the native browser if this is something else than a `zim://...` request).
The save mechanism of Qt must be lost with this custom scheme and don't save the images and other resources.

mgautierfr:
And this PR implement the save action from the "top bar" menu. But the "Save page" from the popup menu still does nothing.

jetownfeve21:
@kelson42 
Do I keep working on it ?

kelson42:
I think this PR needs to be closed without been merged. Too complicated for now to get it work properly.

PR #229, opened 2019-08-15
don't display zim id in searchbar

### LINKED ISSUE #218: Kiwix Desktop should not print the ZIM id in the search bar
Kiwix Desktop should not print in the searchbox the ZIM ID - never - because this has no meaning for the end user.

image

jetownfeve21:
Display the title if it doesn't start with "zim://"

fix #218

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -81,7 +81,11 @@ SearchBar::SearchBar(QWidget *parent) :
 
 void SearchBar::on_currentTitleChanged(const QString& title)
 {
-    setText(title);
+    if (!title.startsWith("zim://")) {
+        setText(title);
+    } else {
+        setText("");
+    }
     m_button.set_searchMode(title.isEmpty());
 }
 


src/searchbar.cpp
```diff
@@ -79,7 +79,9 @@ SearchBar::SearchBar(QWidget *parent) :
 
 void SearchBar::on_currentTitleChanged(const QString& title)
 {
-    setText(title);
+    if (!title.startsWith("zim://")) {
+        setText(title);
+    }
```
mgautierfr: We should display an empty string (or something else) if title doesn't start with "zim://". This is what we do in the search bar (https://github.com/kiwix/kiwix-desktop/blob/master/src/tabbar.cpp#L147-L153)

The `url.path()` is better than nothing, but if the `url` is empty (or `/`) it would be even better to display the zim name. (But it should be in another issue/pr).

kelson42: @mgautierfr I'm not so sure how @jetownfeve21 applies in real-life... If we keep it like it is, are we sure something will go wrong?

mgautierfr: The problem is that if we do nothing, we will keep the title of the previous page.

jetownfeve21: So for now I display an empty string and maybe in another pr it will be the zim name or something else ?

mgautierfr: yes

PR #231, opened 2019-08-20
hidden tags aren't display

### LINKED ISSUE #227: Hidden tags should not be displayed
Tags starting with '_'

jetownfeve21:
Split the tags into a stringlist to remove tags that start with "_" and then
join them into one string with a space as separator.
Remove the replacement of the commoda points by spaces from the js file
because we do not need it anymore.

fix #227

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -68,7 +68,6 @@ QStringList ContentManager::getBookInfos(QString id, const QStringList &keys)
         ADD_V("date", getDate);
         ADD_V("url", getUrl);
         ADD_V("name", getName);
-        ADD_V("tags", getTags);
         ADD_V("origId", getOrigId);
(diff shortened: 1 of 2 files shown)


src/contentmanager.cpp
```diff
@@ -86,6 +85,12 @@ QStringList ContentManager::getBookInfos(QString id, const QStringList &keys)
         if (key == "mediaCount") {
             values.append(QString::number(b->getMediaCount()));
         }
+        if (key == "tags") {
+            QStringList tagList = QString::fromStdString(b->getTags()).split(';');
+            tagList.replaceInStrings(QRegExp("^_.*$"), "");
```
mgautierfr: `filter` would have been better here. Because this is what we do : Remove tags starting by `_`.

Replace the tag by an empty string :
- Is more complex to understand what we do.
- Add an extra space around the "hidden" tag.

PR #233, opened 2019-08-22
remove status bar

### LINKED ISSUE #71: Remove status bar
The status bar at the far bottom is always empty. So far I can see with the new UI design, there should not be any status bar at all. @schoenbaechler You confirm?

@mgautierfr depending a bit what @schoenbaechler means, but I would remove it or keep it in sync: display properly the url which is pointed to know where a link point to.

jetownfeve21:
fix #71

### DIFF
--- a/ui/mainwindow.ui
+++ b/ui/mainwindow.ui
@@ -100,7 +100,6 @@
     <bool>false</bool>
    </attribute>
   </widget>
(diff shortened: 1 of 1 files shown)


jetownfeve21:
I hide the statusbar in this pr #236  so this pr is no longer usefull. I close it

PR #236, opened 2019-08-23
Tooltip appears when a link has to be displayed

### LINKED ISSUE #71: Remove status bar

### LINKED ISSUE #209: URL is not written in the bottom bar
By being over the link/clicking on it.

jetownfeve21:
The status bar is hidden at the start of the app, then when the mouse is
over a link, the status bar appears with the url displayed.
The event "linkHovered" returns an empty string when there is no url, so
the status bar become hidden when it happens

fix #209

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -5,6 +5,7 @@
 #include <QTimer>
 #include <QWebEnginePage>
 #include <QToolButton>
+#include <QToolTip>
 
 #define QUITIFNULL(VIEW) if (nullptr==(VIEW)) { return; }
 #define QUITIFNOTCURRENT(VIEW) if((VIEW)!=currentWidget()) {return;}
@@ -121,6 +122,21 @@ WebView* TabBar::createNewTab(bool setCurrent)
             [=]() {
                 QUITIFNOTCURRENT(webView);
                 emit webActionEnabledChanged(QWebEnginePage::Forward, webView->isWebActionEnabled(QWebEnginePage::Forward));
+            });
+    connect(webView->page(), &QWebEnginePage::linkHovered, this, 
+            [=](const QString& url) {
+                auto tabbar = KiwixApp::instance()->getTabWidget();
+                if (url.isEmpty()) {
+                    QToolTip::hideText();
+                } else {
+                    auto link = url;
+                    if (url.startsWith("zim://")) {
+                        link = QUrl(url).path();
+                    }
+                    auto height = tabbar->currentWidget()->height() + 1;
+                    auto pos = tabbar->mapToGlobal(QPoint(-3, height));
+                    QToolTip::showText(pos, link);
+                }
             });
     // Ownership of webview is passed to the tabWidget
     auto index = count() - 1;
--- a/src/webview.cpp
+++ b/src/webview.cpp
@@ -5,6 +5,7 @@
 #include <iostream>
 #include "kiwixapp.h"
 #include "webpage.h"
+#include <QToolTip>
 
 WebView::WebView(QWidget *parent)
     : QWebEngineView(parent)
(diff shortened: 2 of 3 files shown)


kelson42:
@jetownfeve21 It should appears as an overlay like in FF or Chrome, if the status bar just appears and disappear depending if we are over a link or not, this will be horrible. Could you make a small screencast with Peek please?

jetownfeve21:
Peek 2019-08-24 11-34

jetownfeve21:
@kelson42 
Maybe I can improve the animation and the style of the status bar

kelson42:
@jetownfeve21 This is not bad, but if you could secure it looks like in a browser, this would be better. Maybe using another approach, like creating a custom tooltip, would be easier.

kelson42:
@jetownfeve21 Maybe with http://www.kuqin.com/qtdocument/qtooltip.html

jetownfeve21:
The event "linkHovered" is used to display a tooltip message with the url of the
link.
It is positioned at the bottom left corner of the webview.

src/tabbar.cpp
```diff
@@ -122,6 +123,17 @@ WebView* TabBar::createNewTab(bool setCurrent)
...
                 emit webActionEnabledChanged(QWebEnginePage::Forward, webView->isWebActionEnabled(QWebEnginePage::Forward));
             });
+    connect(webView->page(), &QWebEnginePage::linkHovered, this, 
+            [=](const QString& url) {
+                auto tabbar = KiwixApp::instance()->getTabWidget();
+                if (url.isEmpty()) {
+                    QToolTip::hideText();
+                } else {
+                    auto height = tabbar->currentWidget()->height() + 1;
+                    auto pos = tabbar->mapToGlobal(QPoint(-3, height));
```
jetownfeve21: I have to adjust the position of the tooltip here to correctly fit the side of the window when there will be no more status bar (if it's still relevant https://github.com/kiwix/kiwix-desktop/issues/71).

mgautierfr: I think #71 is still relevant. Please do a PR for this (you've pushed the branch but not created a pr)

jetownfeve21: I already created a pr for this but I closed it when I thought that I have just to have it. I don't know if it's possible to reopen this pr https://github.com/kiwix/kiwix-desktop/pull/233 ?

mgautierfr: Are you hiding the statusbar here (as said in #71) ?

jetownfeve21: not anymore

mgautierfr:
Should we display only the path ("foo.html") or the full url ("zim://<long_id>.zim/foo.hmtl") ?

kelson42:
Good question.. but "/foo.html" - so absolut path - otherwise "https://" if external link.

src/tabbar.cpp
```diff
@@ -122,6 +123,21 @@ WebView* TabBar::createNewTab(bool setCurrent)
...
+    connect(webView->page(), &QWebEnginePage::linkHovered, this, 
+            [=](const QString& url) {
+                auto tabbar = KiwixApp::instance()->getTabWidget();
+                if (url.isEmpty()) {
+                    QToolTip::hideText();
+                } else {
+                    auto link = url;
+                    if (url.startsWith("zim://")) {
+                        link.remove(0, url.indexOf(".zim")+ 4);
+                    }
```
mgautierfr: Would be better to convert this to a QUrl and use the scheme and path method :
https://doc.qt.io/Qt-5/qurl.html#path

PR #241, opened 2019-09-03
Small CSS/icon fixes around top bar

kelson42:


mgautierfr:
Seems good to me. I'm against the CodeFactor warning, it seems more coherent to have the 0 with the unit when all the other values have a unit.

BTW, next time, some screenshot showing what is fixed would be helpful :)

(I've just rebased on top of master).

PR #242, opened 2019-09-05
open a filedialog to download files

### LINKED ISSUE #26: Open a download filepicker if content can not be displayed by the qtwebengine
For example, with the Gutenberg ZIM file, we have a lot of EPUB. Currently, clicking on an EPUB makes nothing. A download/open filepicker (of the operating system) should be displayed instead.

jetownfeve21:
In the UrlSchemeHandler::handleContentRequest method if the mime type indicates an epub file it opens a filedialog to choose the path (and the filename).
If the filename chosen doesn't end with ".epub" it appends it at the end. The data is then written in this file.

fix #26 

This pr only handles epub. For every other cases that will not display a filepicker in the future, the issue has to be reopen to add this specific case.

### DIFF
--- a/src/webpage.cpp
+++ b/src/webpage.cpp
@@ -1,11 +1,14 @@
 #include "webpage.h"
 
 #include <QDesktopServices>
+#include <QFileDialog>
+#include "kiwixapp.h"
+#include <QWebEngineProfile>
 
 WebPage::WebPage(QObject *parent) :
     QWebEnginePage(parent)
 {
-
+    connect(QWebEngineProfile::defaultProfile(), &QWebEngineProfile::downloadRequested, this, &WebPage::startDownload);
 }
 
 bool WebPage::acceptNavigationRequest(const QUrl &url, QWebEnginePage::NavigationType type, bool isMainFrame)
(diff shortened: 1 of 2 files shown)


mgautierfr:
It seems that Qt already has functionality to handle download (https://doc.qt.io/qt-5/qwebengineprofile.html#downloadRequested, https://doc.qt.io/qt-5/qwebenginedownloaditem.html, https://doc.qt.io/qt-5/qwebenginepage.html#download). It would be better to use it.

From a quick reading of the documentation, it seems that we should call `QWebEnginePage::download` from the `acceptNavigationRequest` and then correctly handle the `downloadRequested` signal. I think that on the `UrlSchemeHandle` there will be nothing to do.

src/webpage.cpp
```diff
@@ -14,6 +17,19 @@ bool WebPage::acceptNavigationRequest(const QUrl &url, QWebEnginePage::Navigatio
         QDesktopServices::openUrl(url);
         return false;
     }
+    if (url.url().endsWith(".epub")) {
```
mgautierfr: It seems that you don't even have to launch the download yourself.

Qt starts the download itself.
You have to display the file selection dialog box in `startDownload` and set the path with `download->setPath`.

And be careful to refuse the download if the user "cancel" the dialog box. Else Qt still download the file using the default path.

mgautierfr: And it would work also on other files than `epub`.

mgautierfr:
It works (even if we still have some change to do).
However, there is no user feedback on the downloading (progress, finish, ...). And we have no design for this.
(We should open a issue and see later. For now, just have the file downloaded is enough)

kelson42:
@mgautierfr Do we really need a quality feedback? All of this is local and should be almost instantaneous? An error should only be displayed IMO of the writing to the disk just fails (or maybe this is what you meant?).

mgautierfr:
We don't have any feedback at all for now. Maybe a simple dialog box says that download is complete/has fail is enough.

PR #244, opened 2019-09-10
[URGENT] quick fix tabbar navigation

jetownfeve21:
Add the execution of normal mousePressEvent() if it's not the middlebutton.
Without this the navigation between tab is impossible.


mgautierfr:
Ho :flushed: silly bug indeed.

PR #258, opened 2019-09-18
Fix handleMetaRequest.

### LINKED ISSUE #239: Broken icons in local library
With `kiwix-desktop_x86_64_2.0-rc1-1.appimage` and the content have been downloaded with the Flathub version of Kiwix Desktop.

image

Remark: After better testing, it seems unrelated to the other Flathub version. the `.appimage` seems unable to download content and display the icon properly. The icon of contents in local library are always broken.

mgautierfr:
The main fix is about to return after we reply the request and not do a
`request->fail` just after.
The try/catch is just here be safety.

It seems that depending of the version of Qt, doing a `reply` followed by
a `fail`, the request is replied (as with Qt packaged in fedora) or
failed (as with Qt packaged in ubuntu).

This should fix #239

### DIFF
--- a/src/urlschemehandler.cpp
+++ b/src/urlschemehandler.cpp
@@ -62,15 +62,18 @@ UrlSchemeHandler::handleMetaRequest(QWebEngineUrlRequestJob* request)
     auto zimId = parts[0];
     auto metaName = parts[1];
 
-    auto library = KiwixApp::instance()->getLibrary();
-    auto book = library->getBookById(zimId);
     if (metaName == "favicon") {
-        std::string content= book.getFavicon();
-        std::string mimeType = book.getFaviconMimeType();
-        QBuffer* buffer = new QBuffer;
-        buffer->setData(content.data(), content.size());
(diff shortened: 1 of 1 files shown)


src/urlschemehandler.cpp
```diff
@@ -65,12 +65,17 @@ UrlSchemeHandler::handleMetaRequest(QWebEngineUrlRequestJob* request)
...
     if (metaName == "favicon") {
-        std::string content= book.getFavicon();
-        std::string mimeType = book.getFaviconMimeType();
-        QBuffer* buffer = new QBuffer;
-        buffer->setData(content.data(), content.size());
-        connect(request, &QObject::destroyed, buffer, &QObject::deleteLater);
-        request->reply(QByteArray::fromStdString(mimeType), buffer);
+        try {
+          auto library = KiwixApp::instance()->getLibrary();
+          auto book = library->getBookById(zimId);
```
jetownfeve21: The variables "library" and "book" are already set line 65 and 66 aren't they ?

mgautierfr: Yes, good catch.

PR #263, opened 2019-09-24
Fix reading list

### LINKED ISSUE #252: [REGRESSION] Bookmarks/Reading list items are totally broken
I can not get it work at all anymore. Adding a bookmark seems to work but each time I click on it, the article is not open anymore.

jetownfeve21:
fix some bugs of the reading list :

when a zim is deleted the bookmarks related to it are still save in the library.bookmark.xml file, so at the start of kiwix-desktop the local library object contains "orphan" bookmarks. These bookmarks are not displayed in the readinglist and they are not considered in the ReadingListBar::on_itemActivated function

fix #252


src/readinglistbar.cpp
```diff
@@ -31,7 +31,7 @@ void ReadingListBar::setupList()
     listWidget->clear();
     for(auto& bookmark:bookmarks) {
         auto reader = library->getReader(QString::fromStdString(bookmark.getBookId()));
-        if (reader == nullptr)
+        if (reader == nullptr || !library->getBookIds().contains(QString::fromStdString(bookmark.getBookId())))
```
mgautierfr: When it is possible that the library find a reader for the book but in the same time the library doesn't contain the book ?

src/readinglistbar.cpp
```diff
@@ -50,7 +50,17 @@ void ReadingListBar::setupList()
...
-    auto bookmark = KiwixApp::instance()->getLibrary()->getBookmarks().at(index);
+    auto library = KiwixApp::instance()->getLibrary();
+    auto bookmarks = library->getBookmarks();
+    QVector<kiwix::Bookmark> validBookmarks;
+    auto booksId = library->getBookIds();
+    for(auto& bookmark:bookmarks) {
+        if (booksId.contains(QString::fromStdString(bookmark.getBookId()))) {
+            validBookmarks.append(bookmark);
+        }
+    }
```
mgautierfr: I'm not sure we should have this in kiwix-desktop.
This is to the kiwix-lib to be sure that the content is coherent.
We could add a new argument to `kiwix::Library::getBookmarks` to only list valid bookmark. This way kiwix-desktop doesn't nothing special, when populating the list or reacting to a item activation.

src/readinglistbar.cpp
jetownfeve21: After removing a zim

mgautierfr: But the `getReader` should return `nullptr` in this case no ?

jetownfeve21: I think it should yes :)
I am still able to access to deleted zim with the readinglist (without this pr) is it related to this reader that is not null ?

mgautierfr: This is probably because the reader is still in the `m_readersMap` of the library.
We should remove it from the reader when we remove the book.

mgautierfr: And the new version of kiwix-lib also has an internal map of readers (I've add it for the integration of the server in the kiwix-lib). We should use it in kiwix-desktop instead of having a separated map.
But there is the same bug in kiwix-lib, the reader is not removed from the map when we remove the book.

jetownfeve21:
needs https://github.com/kiwix/kiwix-lib/pull/288 to work

PR #336, opened 2019-12-19
add margin for left and right icon in toolbar

### LINKED ISSUE #283: Menu bar left and rigth icon margins to window border is too small
See here for the left icon:
image

Here for the right icon:
image

The margins (left for the left icon, and right for the right icon) are only a few pixels, they should be the same as top and bottom.

jetownfeve21:
fix #283

### DIFF
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -69,6 +69,15 @@ TopWidget QToolButton::menu-arrow {
     image: none;
 }
 
+TopWidget QToolButton#backButton {
+    margin-left: 6px;
+}
+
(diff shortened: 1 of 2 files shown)


resources/css/style.css
```diff
@@ -69,6 +69,15 @@ TopWidget QToolButton::menu-arrow {
     image: none;
 }
 
+TopWidget QToolButton#backButton {
+    margin-left: 6px;
+}
+
```
kelson42: Why an additional line here?

PR #342, opened 2020-01-06
close every empty tab when opening a external link

### LINKED ISSUE #326: Open in a new tab does not work fine with external links

jetownfeve21:
fix #326 

Probably not the best way to do it. I wanted to prevent the application from opening a tab instead of close it manually but didn't find a way to do it


kelson42:
@jetownfeve21 Like I said during the meeting, this has to be solved properly. Opening and closing the tab is not a proper solution to me.

jetownfeve21:
I close this PR because [this PR](https://github.com/kiwix/kiwix-desktop/pull/386) is better

PR #343, opened 2020-01-07
closing a tab keeps the scroll position of the right tab

### LINKED ISSUE #329: Closing a page auto scrolls the page on the right to top
Suppose tab1 with page1 and tab2 with page2. Tab2 is on the right of Tab1, when closing page1, page 2 is auto scrolled to the top. Page 2 should have keep its scroll state.

jetownfeve21:
QTabBar::setSelectionBehaviorOnRemove seems to reset the scroll position of
the selected tab, set the current index manually keeps it.

fix #329

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -252,11 +252,10 @@ void TabBar::closeTab(int index)
 
 void TabBar::setSelectionBehaviorOnRemove(int index)
 {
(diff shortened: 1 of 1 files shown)


kelson42:
@jetownfeve21 Seems to work, even if don't really understand the code.

PR #345, opened 2020-01-08
close tab button style improvement

### LINKED ISSUE #284: Tab close button is not like in UI design mockup
The button has two problems:
* It is far too small
* There is not background appearing if the mouse is over it (so it is not clear, it is clickable)

In the picture you see on the top the UI design mockup (how it should be) and below a screenshot I made (how it is now):
image

jetownfeve21:
bigger icon and background when mouse over it

fix #284


kelson42:
@jetownfeve21 The size is OK now, but:
* The color/thickness is not ok. Have a look to the mockup
* The hover effect is not OK, it should be like in the other buttons

PR #346, opened 2020-01-08
bigger menu icon

### LINKED ISSUE #281: Menu logos should be bigger
They are almost twice to small in comparison to the UI design mockups

Here is an example I have taken with the "Settings" menu entry. The first one is the UI design mockup, the second is issue from a screenshot I have made of my Kiwix Desktop.

image

jetownfeve21:
apply a custom proxystyle for the menu widget to change the icon size

fix #281

### DIFF
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -96,7 +96,8 @@ HEADERS += \
     src/opdsrequestmanager.h \
     src/localkiwixserver.h \
     src/fullscreenwindow.h \
-    src/fullscreennotification.h
+    src/fullscreennotification.h \
+    src/menuproxystyle.h
 
 FORMS += \
     ui/mainwindow.ui \
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -88,7 +88,7 @@ QMenu::item {
     min-height: 40px;
     max-height: 40px;
     border: 1px solid transparent;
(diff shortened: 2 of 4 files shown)


kelson42:
@jetownfeve21 Any news on that one? The solution looks more like a workaround to me.

jetownfeve21:
@kelson42 It seems that the overlapping problem of the appimage comes from the Qt version that the appimage uses which is Qt 5.9

kelson42:
I use Qt 5.9.
```
$ qmake --version
QMake version 3.1
Using Qt version 5.9.5 in /usr/lib/x86_64-linux-gnu
```

@mgautierfr @jetownfeve21 Same as #176 ?

mgautierfr:
Yes, We have to move to a more recent version of Qt. See kiwix/kiwix-build#403

kelson42:
@jetownfeve21 Then, I will review again your PR and if OK (beside this) I will merge it. Flatpak does already use 5.13 and Appveyor 5.12... so we should not be strongly impacted. The problem with appimage is well known and I don't want to stop things because of it.

mgautierfr:
> Flatpak does already use 5.13

No, it use 5.12

kelson42:
@jetownfeve21 Surprisingly it displays now properly here on my machine. Looks really better. Was worth the effort.

PR #347, opened 2020-01-10
Find in page is now a bottom bar

### LINKED ISSUE #325: Cancel "Find in page" highlighting
Now when search in a loaded page, the searched key word will be highlighted, but there is no way to remove the highlights. Maybe there need to be a difference between hide-search-window and cancel-search-clear-highlights.

### LINKED ISSUE #327: Mutate "Find in Page" to a bottom bar
Current implementation of "Find in Page" displays it formular as a sidebar.
image

This takes a lot of space for "nothing" because most of the sidebar is empty. We don't have a UI mockup for that I think. Therefore I would recommend to implement it like in Firefox or Chrome, as a bottom bar.

image

@jetownfeve21 Does that make sense?

jetownfeve21:
fix #325 
fix #327


jetownfeve21:
@kelson42 I fixed the resize problem but for the top-border it seems complicated to reduce it.

kelson42:
@jetownfeve21 I don't have asked to reduce it. I have asked to (1) not make it resizeable (2) put a border with the right color. Which part is a problem?

jetownfeve21:
image

I didn't succeed to make a top-border with the css, so I made a widget of 1px height but there is still a small margin on top of it

kelson42:
> I didn't succeed to make a top-border with the css, so I made a widget of 1px height but there is still a small margin on top of it

@jetownfeve21 This can not be the right approach but I can not tell you more... If you feel stuck and @mgautierfr can not help you, then please ask in a forum, open a ticket. Link everything in our ticket please.

kelson42:
@jetownfeve21 Do you have open an upstream ticket? Ask Matthieu?

src/tocbar.cpp
```diff
@@ -47,17 +50,36 @@ void TocSideBar::findNext()
...
+    close();
 }
+
+void TocBar::findCloseByIndex(int index)
+{
+    auto widget = KiwixApp::instance()->getTabWidget()->widget(index);
+    if (!widget)
+        return;
+    auto page = widget->page();
+    page->findText("");
```
mgautierfr: I'm not sure we should clear the search when we change tab (or display the contentmanager).
When the user come back to the tab, the search should be there.
We could save the searched text in the page (in our `WebPage`). When the current tab is change, the searchbar should check if there is a search text for the page and set itself to be visible if needed.

jetownfeve21:
@mgautierfr I don't succeed to add a top border properly like it is asked, do you have an idea ?

jetownfeve21:
@kelson42 I finally succeed to add a proper top border

PR #349, opened 2020-01-13
fix searchbar that doesn't search after switching from another app

### LINKED ISSUE #287: Search bar doens't search after copy/cut text from other programs
Finally found the key: The problem only appear when:
(1) Be sure the cursor is in the search bar and is blinking, ready for input
(2) Switch to another program then switch back. When you switch back, be sure you click the item in Task Bar, not click the Kiwix UI, since if the cursor go away from the search bar, the problem will be fixed as described bellow.
(3) Continue to input in the search bar (not needed if the input is already completed before), then press enter or click the item in the live search drop down, now it won't show the selected page anymore,  whatever you continue to input and action.

The problem will disappear if you click other places in Kiwix like the page content and let the cursor go away from the search bar. Then when the focus goes back to the search bar again, the search bar will be cleared automatically , now the problem is fixed. If you do the above again, the problem will appear again. Seems that when focus go back to the search bar, some reinitialization code reset the corrupted status when you switch to another program while the focus is in the search bar.

If you click different tab headers while the focus is in the search bar, the problem won't be fixed since the focus is keeped in the search bar during this process. This also mean that the it is not the focus, but the reinitialization of the search bar that fix the problem.

jetownfeve21:
Same problem as this problem https://github.com/kiwix/kiwix-desktop/pull/180

The openCompletion has to be connected every time the searchbar gains the focus.

fix #287

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -93,6 +93,9 @@ void SearchBar::focusInEvent( QFocusEvent* event)
 {
     if (event->reason() == Qt::MouseFocusReason) {
         clear();
+    }
+    if (event->reason() == Qt::ActiveWindowFocusReason ||
+        event->reason() == Qt::MouseFocusReason) {
         connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
         this, &SearchBar::openCompletion);
     }


kelson42:
@jetownfeve21 It works :) Thx.

Issue #299, opened 2019-11-14
Translate Kiwix Desktop 2.0 with Translatewiki


kelson42:
Translatewiki should:
* load english master (bot)
* translate the strings in local languages (community)
* commit the translation in this repo (bot)

kelson42:
@translatewiki The documentation about Qt i18n is available here https://doc.qt.io/qt-5/linguist-translators.html

kelson42:
I reopen the ticket as we need to secure that these repo is properly read/written by @translatewiki. @Nikerabbit @Abijeet Would you be able to secure this?

kelson42:
@translatewiki. @Nikerabbit @Abijeet  Would you please be able to configure the Translatewiki bot to import/export strings? Do I need to open a ticket somewhere else?

Abijeet:
Hi @kelson42 - We've created a ticket on Phabricator to track the addition of this project to Translatewiki - https://phabricator.wikimedia.org/T251483

We should be able to finish this before the end of this week. Can you ensure that the translatewiki bot has read / write permissions to this repo?

kelson42:
@Abijeet Thank you. @translatewiki is part of the "Kiwix" Github organization and should have write access to all our repositories. I have as well sent an invitation to the "OpenZIM" organization as we will have to sync other repositories there as well.

Abijeet:
@kelson42 - The changes have been deployed. We're tracking the `master`, and pushing translations to the `translatewiki` branch. We've recently added functionality to create and submit a PR. Would you like to enable this for the Andorid and Desktop repositories?

Secondly, I would recommend adding some [message documentation](https://translatewiki.net/wiki/Translating:Localisation_for_developers#Message_documentation) to make it easier for translators. You can do that from here - https://translatewiki.net/wiki/Special:Translate?action=translate&group=kiwix-desktop&language=qqq&filter=

Abijeet:
Changes pushed here - https://github.com/kiwix/kiwix-desktop/commit/c224a38be1b099222bfeb067051248cfe857382b

kelson42:
@Abijeet Thank you very much, so far to me it looks good. I let the expert @mgautierfr @jetownfeve21 to review the PR I created based on the `translatewiki` branch.

mgautierfr:
> Would you like to enable this for the Andorid and Desktop repositories?

Yes please. It would be nice to have PR created atomically.

kelson42:
[...]

PR #351, opened 2020-01-14
Aria2 crash fix

### LINKED ISSUE #272: If aria2c can not be started, print the command line which has failed
Following #268, it looks like this is pretty difficult to understand what is going on. If the aria2c downloader can not be setup, which basically means that the aria2c binary has not been launched successfully, the full command should be printed somewhere (on the terminal for example) so we get a chance to understand why it failed.

jetownfeve21:
display aria2 launch cmd that doesn't crash

The program crashes because it tries to use mp_downloader->getAria2LaunchCmd() but the mp_downloader is null if the aria2 launch command fail.
Now the program displays an error message to warn users that all download fct will not working.

This PR https://github.com/kiwix/kiwix-lib/pull/306 is needed in order to display the aria2 launch cmd.

fix #272

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -11,27 +11,14 @@
 #include <QPrinter>
 #include <QPrintDialog>
 #include <thread>
-
-kiwix::Downloader* createDownloader() {
-    int attempt = 5;
-    while(attempt--) {
-        try {
-            return new kiwix::Downloader();
-        } catch (exception& e) {
-            qInfo() << "Cannot create downloader" << e.what();
-            std::this_thread::sleep_for(std::chrono::milliseconds(500));
-        }
-    }
-    return nullptr;
-}
+#include <QMessageBox>
 
 KiwixApp::KiwixApp(int& argc, char *argv[])
     : QApplication(argc, argv),
       m_settingsManager(),
       m_libraryDirectory(findLibraryDirectory()),
       m_library(),
-      mp_downloader(createDownloader()),
-      m_manager(&m_library, mp_downloader),
+      mp_downloader(nullptr),
       mp_server(new kiwix::KiwixServe(
         appendToDirectory(m_libraryDirectory.toStdString(),"library.xml"),
         m_settingsManager.getKiwixServerPort()))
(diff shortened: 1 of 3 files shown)


src/kiwixapp.cpp
```diff
@@ -11,31 +11,25 @@
...
+      mp_downloader(nullptr),
       mp_server(new kiwix::KiwixServe(
         appendToDirectory(m_libraryDirectory.toStdString(),"library.xml"),
         m_settingsManager.getKiwixServerPort()))
 {
+    try {
+        mp_downloader = new kiwix::Downloader();
+    } catch (exception& e) {
+        QMessageBox::critical(nullptr, tr("Cannot create downloader"),
+        tr("Impossible to launch downloader, Kiwix-desktop will start but all download functions will not working !<br><br>") + e.what());
```
mgautierfr: `tr` will try to load the localized text. So we should install the translators before.

src/kiwixapp.cpp
```diff
@@ -115,6 +105,9 @@ KiwixApp::~KiwixApp()
         mp_downloader->close();
         delete mp_downloader;
     }
+    if (mp_manager) {
+        delete mp_manager;
+    }
```
mgautierfr: mp_manager should always exists. No need to test if mp_manager is not null

src/kiwixapp.cpp
jetownfeve21: We will not use the tr() fct to translate, I want to remove all the "hard coded" text in another PR to fix https://github.com/kiwix/kiwix-desktop/issues/299 and use a translate file instead to work with translateWiki. This solution should also make it possible to better manage translations in HTML/JS files such as the contentmanager page.

mgautierfr: Ok, I don't know how translateWiki works. But it will probably use a function to get the text
to display to the user. And this function must be used after the configuration of the
translation system.

And for now, we are using `tr`. We must move this after the installation of the translators
anyway. At least to have the translation working while we are moving to translateWiki.

kelson42:
@jetownfeve21 Could you please finish this PR?

mgautierfr:
We are good !

Issue #170, opened 2019-06-07
Change profile/download path in Settings


gmikhail:
Now the program downloads the file on the system disk, regardless of its location. What in some cases may cause the disk to become full, which causes, complete freeze of kiwix-desktop and OS.

The following fixes are required (by priority):
- By default, the download should be in the current working folder (now everything is download into the system disk)
- Pre-calculate free space and warn about its shortage before download
- Ability to change the download path

kiwix-desktop_windows_x64_2.0-beta4 (2.0-beta4-11-ga005a5c)
Win 10, 128GB SSD

kelson42:
@gmikhail Thank you for your feedback. The plan is to offer a way to change the profile location (something similar to the old kiwix-xulrunner). And yes, the software shoudl check if there is enough disk place, I have created a ticket for that here #171

kelson42:
@jetownfeve21 Is that ticket fixed?

jetownfeve21:
No there is still the profile directory that can not be changed for now

kelson42:
@jetownfeve21 I want to but this ticket on stall for the moment as i believe:
* Moving the profile is touchy
* Most of the added value is already achieved with the download location feature

autolyk:
It seems Kiwix checks space only on system drive. I have enough space on the disk download directory's disk, but no enough on system disk. And I got "not enough storage available" error message.
(Windows 7, Kiwix Desktop x64 2.0.1-1)

Fubar1:
I am getting an error "not enough storage space" for an 89GB file even though the download directory has 2TB of space, C Drive has over 500GB, program drive has 1TB.

kelson42:
@Fubar1 Your comment has nothing to do with this ticket. But you report a real bug which has been meanwhile fixed, see #868.

kelson42:
Closing as the feature has been implemented

michaeljakob:
There is still no way to set download folder.

kelson42:
@michaeljakob which version do you use!

0o0will0o0:
> [@michaeljakob](https://github.com/michaeljakob) which version do you use!

Well, the current version on Windows only downloads on the main drive with no way to change the destination. Crazy it's been 5 years with no fix.

kelson42:
@0o0will0o0 Crazy you are not able to run/use  latest version 2.4.0!

Your comment is not helpful and pretty respectless.

0o0will0o0:
[...]

PR #355, opened 2020-01-22
[WIP] Download path can be changed in the settings

jetownfeve21:
need https://github.com/kiwix/kiwix-lib/pull/310
https://github.com/kiwix/kiwix-desktop/issues/170

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -217,12 +217,11 @@ QString ContentManager::downloadBook(const QString &id)
     return QString::fromStdString(download->getDid());
 }
 
-void ContentManager::eraseBookFilesFromComputer(const QString fileToRemove)
+void ContentManager::eraseBookFilesFromComputer(const QString dirPath, const QString filename)
 {
-    QString dirName = KiwixApp::instance()->getLibraryDirectory();
-    QDir dir(dirName, fileToRemove);
-    for(const QString& filename: dir.entryList()) {
-        dir.remove(filename);
+    QDir dir(dirPath, filename);
+    for(const QString& file: dir.entryList()) {
+        dir.remove(file);
     }
 }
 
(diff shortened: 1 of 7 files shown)


jetownfeve21:
@mgautierfr Can I have your review about this before starting to work on the profile path ?

kelson42:
@mgautierfr Please review this PR.

src/contentmanager.cpp
```diff
@@ -214,12 +217,13 @@ QString ContentManager::downloadBook(const QString &id)
     return QString::fromStdString(download->getDid());
 }
 
-void ContentManager::eraseBookFilesFromComputer(const QString fileToRemove)
+void ContentManager::eraseBookFilesFromComputer(const QString bookPath)
```
mgautierfr: I think this method is a bit too complex.
We only use this method in eraseBook and cancelBook. There, we create a Qstring from the book path and then we split it in the function to get the different parts.

We should change the signature of the method to directly give the different information we need.

src/settingsmanager.cpp
```diff
@@ -61,8 +65,35 @@ void SettingsManager::setZoomFactor(qreal zoomFactor)
...
+                                                    QString(),
+                                                    QFileDialog::ShowDirsOnly);
+    setDownloadPath(dir);
+}
+
 void SettingsManager::initSettings()
 {
     m_kiwixServerPort = m_settings.value("localKiwixServer/port", 8181).toInt();
     m_zoomFactor = m_settings.value("view/zoomFactor", 1).toDouble();
+    m_downloadPath = m_settings.value("download/path", QString::fromStdString(getDataDirectory())).toString();
```
mgautierfr: I would prefer "download/dir" (and `downloadDir` in all the method/variable name).

The `dir` is explicitly a directory. Naively, I would said that the download path is the full path of the zim file, not the directory where we are downloading it.

kelson42:
@jetownfeve21 
* Please remove "WIP" from the PR if your PR is ready
* Please put properly "Fixes #XYZ" to close the ticket automatically. If this is should not closing the ticket, please make a comment explaining what this PR does.

PR #361, opened 2020-01-24
Fixes problems with tab closing

### LINKED ISSUE #359: [REGRESSION] Impossible to close tabs if focus is on library tab
Steps:
* Open one content tab
* Switch to the library
* try to close the content tab

What happen:
* Nothing happen, the content tab button seems deactivated

What should happen:
* It should be closeable

jetownfeve21:
fix #359


kelson42:
@jetownfeve21 Work like it should
@mgautierfr I would be happy to merge it but it seems @jetownfeve21 wants confirmation from you that his code is OK

Issue #356, opened 2020-01-22
Opening a new tab should put the focus on the search bar


kelson42:
Creating a new tab with "+" or keyboard shortcut should put the focus on the search bar, so the user can directly type its search pattern without having to focus manually.

arghac14:
@kelson42 Can I work on this?

kelson42:
@arghac14 yes

PR #362, opened 2020-01-28
set the focus on the search bar when opening a new tab

jetownfeve21:
#356


src/mainwindow.cpp
```diff
@@ -34,6 +34,7 @@ MainWindow::MainWindow(QWidget *parent) :
             setWindowTitle(title + " - Kiwix");
         else
             setWindowTitle(tr("Library") + " - Kiwix");
+    connect(getTabBar(), &TabBar::newTabOpened, getTopWidget(), &TopWidget::onNewTabOpened);
```
kelson42: indentation is not correct

PR #366, opened 2020-01-30
New tag system

mgautierfr:


src/kiwixapp.cpp
```diff
@@ -196,6 +196,7 @@ void KiwixApp::setSideBar(KiwixApp::SideBarType type)
     switch(type) {
         case SEARCH_BAR:
             mp_mainWindow->findChild<TocSideBar*>("tocsidebar")->getFindLineEdit()->setFocus();
+            break;
```
jetownfeve21: Because the find in page search bar will no longer be in the sidebar (#327), I removed these lines in the PR https://github.com/kiwix/kiwix-desktop/pull/347/commits/31c0b53418f712de4500ab681809ea9f76ebd817, so I don't think that it's important to add this ```break```.

mgautierfr: I've removed the commit

mgautierfr:
> Most of the filters by category don't work properly on my side

There is a lot of change made on kiwix-lib/kiwix-tools on tags issue (and filter per category, especially https://github.com/kiwix/kiwix-lib/commit/7846b45befbd44edfd96f927fbced82ee35c7a86). Please test with a local kiwix-serve.
We will do a release of kiwix-lib/kiwix-tools soon.
If this change is "correct" with current kiwix-serve, we can merge this PR without waiting the release.

kelson42:
@mgautierfr @jetownfeve21 Please check that the library filters in Kiwix Desktop. We are just before 2.0 release. If necessary open the corresponding bug tickets.

mgautierfr:
@jetownfeve21 please test this again. Tag are displayed correctly on my side.

jetownfeve21:
@mgautierfr all category filters work for me except the stack exchange one

mgautierfr:
> @mgautierfr all category filters work for me except the stack exchange one

Fixed in last forced push ("stack_exchange" instead of "stackexchange")

mgautierfr:
Last push adapt this PR to the new translation system (to translate category names)

PR #374, opened 2020-02-18
new tab action opens an empty tab

### LINKED ISSUE #360: Opening a new tab should generate a new empty tab
Currently it takes the same content as the tab which currently has the focus. Except if we are on the library tab, then it opens correctly an empty tab.

It should always be an empty tab, at least as long as #140 is not implemented.

jetownfeve21:
fix #360

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -28,7 +28,6 @@ TabBar::TabBar(QWidget *parent) :
                 auto current = this->currentWidget();
                 auto widget = this->createNewTab(true);
                 QUITIFNULL(current);
-                widget->setUrl(current->url());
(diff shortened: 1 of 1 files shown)


src/tabbar.cpp
```diff
@@ -28,7 +28,6 @@ TabBar::TabBar(QWidget *parent) :
                 auto current = this->currentWidget();
                 auto widget = this->createNewTab(true);
                 QUITIFNULL(current);
```
mgautierfr: If we do nothing with `widget` and `current` we should not get them and store them in a variable.

The whole function can be reduced to `this->createNewTab(true);`

PR #375, opened 2020-02-18
Restore the title tab when searchbar loses focus

jetownfeve21:
reimplement QLineEdit::focusInEvent(QFocusEvent* event)
use setReadOnly method to display/hide blinking cursor

fix #330

PR #382, opened 2020-03-04
[Urgent]  Fix crash settings page

jetownfeve21:
This PR https://github.com/kiwix/kiwix-desktop/pull/375 makes the settings page crash.
It's because it tries to get the zimId of the current widget but obviously the settings page doesn't have one.

To prevent this from happening again the Tabbar::currentWidget method returns a nullptr if the current widget is the settings page (like the content manager page does)


mgautierfr:
It's merged.

However, I think that in the future, we should change the way we track the current id. I think a better solution would be to have the kiwixApp should use event to track which tab is currently displayed and set a "currentZim" property somewhere. (Or directly have a property in kiwixapp, and have webviews setting the property when they have the focus).

PR #386, opened 2020-03-12
Ctrl + Click on external link doesn't open a new empty tab

### LINKED ISSUE #326: Open in a new tab does not work fine with external links

jetownfeve21:
Connect the &QWebEnginePage::linkHovered signal to set the m_linkHovered
member.
The Ctrl + Click event is intercepted to directly open the m_linkHovered
variable if it's an external link

fix #326

### DIFF
--- a/src/webview.cpp
+++ b/src/webview.cpp
@@ -1,5 +1,6 @@
 #include "webview.h"
 
+#include <QDesktopServices>
 #include <QAction>
 #include <iostream>
 #include "kiwixapp.h"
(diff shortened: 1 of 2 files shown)


kelson42:
@jetownfeve21 Nice to see this PR!

kelson42:
@mgautierfr Could you please review this to merge before 2.0 release?

src/webview.cpp
```diff
@@ -75,6 +79,13 @@ bool WebView::eventFilter(QObject *src, QEvent *e)
         if (we->modifiers() == Qt::ControlModifier)
             return true;
     }
+    if (e->type() == QEvent::MouseButtonRelease) {
+        auto me = static_cast<QMouseEvent *>(e);
+        if (me->modifiers() == Qt::ControlModifier && !m_linkHovered.startsWith("zim://")) {
```
mgautierfr: There is two way to open a link in a new tab : with Ctrl-click and middle click.
You are catching the ControlModifier but you must also catch the middleButton.

PR #387, opened 2020-03-17
Prepare string files for Translatewiki

### LINKED ISSUE #299: Translate Kiwix Desktop 2.0 with Translatewiki

jetownfeve21:
fix #299


jetownfeve21:
@mgautierfr what do you think about this implementation ?
I still have to implement the translation in the html pages too

kelson42:
@jetownfeve21 What is format/schem of the XML to store the strings? Is that a custom format? We need to secure that this could be indgested by Translatewiki.

@Abijeet @translatewiki @Nikerabbit Would you be able to give here an advice? Maybe you have a XML base format you support out-of-the-box?

jetownfeve21:
@kelson42  I used the same format that it is used in Huggle which is another wikimedia project in C++/Qt https://github.com/huggle/huggle3-qt-lx

Nikerabbit:
Huggle is using our AndroidXml file format to read that string. I strongly advise against using that handler though, as it is among the most complicated formats that I have seen. We could consider making a simplified handler without the edge cases, but if you can use JSON file with simple key-value pairs, I would recommend that.

jetownfeve21:
@mgautierfr I added the implementation of the translation for the content manager page

jetownfeve21:
@Nikerabbit Alright I'll change the format

Nikerabbit:
https://github.com/wikimedia/mediawiki-extensions-Translate/blob/master/i18n/core/en.json here is an example of the JSON file format.

jetownfeve21:
Thank you

jetownfeve21:
So now it's a json file format

kelson42:
@jetownfeve21 so we can remove the `[WIP]`?

jetownfeve21:
There are still html pages that aren't ready for translation, I prefer to wait Matthieu's opinion before

resources/texts/_contentManager.html
```diff
@@ -10,30 +10,30 @@
 <div id="app">
   <div id="searchBar">
     <form>
-      <input id="searchInput" type="text" placeholder="Search files" oninput="setSearch(this.value)"/>
+      <input id="searchInput" type="text" :placeholder="translations['search-files']" oninput="setSearch(this.value)"/>
```
mgautierfr: It would be better to have a function (`gt` as on Qt side) to do the lookup for us.
Then we would have to write `gt('search-files')` only (a bit smaller)

src/fullscreennotification.cpp
```diff
@@ -7,7 +8,7 @@ FullScreenNotification::FullScreenNotification(QWidget *parent)
     : QLabel(parent)
     , m_previouslyVisible(false)
 {
-    setText(tr("You are now in full screen mode. Press ESC to quit!"));
+    setText(KiwixApp::instance()->getText("fullscreen-notification"));
```
mgautierfr: It would be nice to have a function or a macro `gt` to do it.
`gt("fullscreen-notification")` instead of `KiwixApp::instance()->getText("fullscreen-notification")`

src/language.h
```diff
@@ -0,0 +1,19 @@
+#ifndef LANGUAGE_H
+#define LANGUAGE_H
+
+#include <QString>
+#include <QMap>
+
+class Language
```
mgautierfr: This class is more about translation or text provider. Language seems a bit to generic.

src/kiwixapp.cpp
```diff
@@ -24,6 +24,7 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
         appendToDirectory(m_libraryDirectory.toStdString(),"library.xml"),
         m_settingsManager.getKiwixServerPort()))
 {
+    m_language.setTranslationsFromJsonFile(":/i18n/en.json");
```
mgautierfr: We should use the QLocale() to load the right json file.
And it would be better if we path the QLocale directly to the function and the function find itself what to load.

resources/translations.qrc
```diff
@@ -1,5 +1,6 @@
 <RCC>
     <qresource prefix="/">
         <file>i18n/kiwix-desktop_fr.qm</file>
+        <file>i18n/en.json</file>
```
mgautierfr: Do we still need `i18n/kiwix-desktop_fr.qm` ? All application translation should be in the json file no ?

src/language.cpp
```diff
@@ -0,0 +1,33 @@
...
+    auto jsonObj = doc.object();
+    for(auto &key : jsonObj.keys()) {
+        m_translations.insert(key, jsonObj.value(key).toString());
+    }
+    qInfo() << m_translations;
+}
+
+QString Language::getText(const QString &key)
+{
+    return m_translations.value(key);
```
mgautierfr: We should handle if the translation is not up to date (missing key).
As we will develop kiwix-desktop we will add new text. And translation will takes time to be updated. We must not fail here.

We can fallback to the en verision, or return the key as a text or ...

src/language.cpp
```diff
@@ -0,0 +1,33 @@
...
+    QFile file(filePath);
+    if (!file.open(QIODevice::ReadOnly)) {
+        return;
+    }
+    QJsonDocument doc = QJsonDocument::fromJson(file.readAll());
+    file.close();
+    if(doc.isNull()) {
+        qInfo() << "Invalid JSON...\n";
+        return;
+    }
```
mgautierfr: We should do something here. No translation means no text at all to display.
(If getText handle missing key, we can do nothing here)

mgautierfr:
I suppose the json file is specific to translateWiki.
I would like to mention that the Qt translation system is bit more complex than just a bank of text.
The translation file contains more information than just the text :
- Where the text is use in the program.
- What is the context, developer commentary (to help translators to know the context and tranlate correctly).
- If the original (english) text as been changed, the translation are marked as obsolete.

There is probably text in the ui files that we need to translate. It will be difficult to do this correctly :/

It seems that :
- wikitranslate support different file format : https://www.mediawiki.org/wiki/Help:Extension:Translate/File_format_support
- We can convert qt file format to another format : https://stackoverflow.com/questions/60277667/can-qt-lrelease-tool-process-xliff-files-with-the-xlf-ending

I think it would be better to just convert the qt format to another format and give it to wikitranslate (if wikitranslate cannot use directly qt format) instead of creating our whole translation system.
However, we should probably keep a small translation system for the html/js text (content manager view)

Nikerabbit:
> I suppose the json file is specific to translateWiki.
> I would like to mention that the Qt translation system is bit more complex than just a bank of text.
> The translation file contains more information than just the text :
> 
> * Where the text is use in the program.

This isn't often useful for translators.

> * What is the context, developer commentary (to help translators to know the context and tranlate correctly).

This is usually placed in a separate file (e.g. qqq.json) except in Gettext files.

> * If the original (english) text as been changed, the translation are marked as obsolete.

This is tracked on translatewiki.net. If the file format allows it, this status is also present in the translation files.

> 
> There is probably text in the ui files that we need to translate. It will be difficult to do this correctly :/
> 
> It seems that :
> 
> * wikitranslate support different file format : https://www.mediawiki.org/wiki/Help:Extension:Translate/File_format_support
> * We can convert qt file format to another format : https://stackoverflow.com/questions/60277667/can-qt-lrelease-tool-process-xliff-files-with-the-xlf-ending
> 
> I think it would be better to just convert the qt format to another format and give it to wikitranslate (if wikitranslate cannot use directly qt format) instead of creating our whole translation system.
> However, we should probably keep a small translation system for the html/js text (content manager view)

We do not currently support Qt .ts file format, but we can add support for it.

mgautierfr:
> This is usually placed in a separate file (e.g. qqq.json) except in Gettext files.

Yes but we don't have it. (And I don't want we implement this is our simple translation system.)

> We do not currently support Qt .ts file format, but we can add support for it.

It would be nice. I don't know if it is easy. Other project than us could benefit from this.
But it is not stricly necessary as we can convert from .ts (or .qm) file format to a format you support.

While you are here, what is the policy about wikiTranslate pushing on the source repository ?
I would like to avoid to "pollute" the git history with automatic commit every time someone change a translation.
Is there a way to regroup the changes ? Or have a script or action in wikiTranslate to update the translation when we want (before a release) ?

Nikerabbit:
> > We do not currently support Qt .ts file format, but we can add support for it.
> 
> It would be nice. I don't know if it is easy. Other project than us could benefit from this.
> But it is not stricly necessary as we can convert from .ts (or .qm) file format to a format you support.

Better to do less conversions to keep it simple.

> While you are here, what is the policy about wikiTranslate pushing on the source repository ?
> I would like to avoid to "pollute" the git history with automatic commit every time someone change a translation.
> Is there a way to regroup the changes ? Or have a script or action in wikiTranslate to update the translation when we want (before a release) ?

We push translations to git two times per week.

resources/texts/_contentManager.html
jetownfeve21: I prefer to use a map which is filled at the initialisation because using a function for each text to translate that will call  a c++ method (each call is asynchrone) is much more complicated than only one call at the beginning I think

mgautierfr: I have the idea that the function do the lookup in the map for us.
It is just syntactic sugar. Using a map is the good thing to do.

jetownfeve21:
@mgautierfr I added the translation for the settings page and more security to the loading of the translations : 
- the local language translation file is loaded by default, if it doesn't exist the english one is used
- if the language is not english then every empty values are filled thanks to the english file

src/kiwixapp.h
```diff
@@ -63,6 +64,7 @@ class KiwixApp : public QApplication
     KiwixApp(int& argc, char *argv[]);
     virtual ~KiwixApp();
     static KiwixApp* instance();
+    static QString gt(const QString &key);
```
mgautierfr: Make this a plain function instead of a method.
Then we can use `gt("key")` instead of `KiwixApp::get("key")`

src/translation.cpp
```diff
@@ -0,0 +1,51 @@
...
+    if (locale.bcp47Name() == "en") {
+        m_translations = defaultText;
+        return;
+    }
+    m_translations = JsonFileToQMap(":/i18n/" + locale.bcp47Name() + ".json");
+    for (auto &key : m_translations.keys()) {
+        if (m_translations.value(key).isEmpty()) {
+            m_translations.insert(key, defaultText.value(key));
+        }
+    }
```
mgautierfr: THere is very few chance the there is a key in the translation with an empty value. (but indeed we need to test that).
What we really want to check is if a key is missing. We should loop over the key of `defaultText` (and test the value in `m_translations`.

src/translation.cpp
```diff
@@ -0,0 +1,51 @@
...
+    auto jsonObj = doc.object();
+    for (auto &key : jsonObj.keys()) {
+        translations.insert(key, jsonObj.value(key).toString());
+    }
+    return translations;
+}
+
+QString Translation::getText(const QString &key)
+{
+    return m_translations.value(key);
```
mgautierfr: We should also add a test here.
The english json may not be up to date. Or a typo in the key.

src/translation.cpp
```diff
@@ -0,0 +1,51 @@
...
+    auto jsonObj = doc.object();
+    for (auto &key : jsonObj.keys()) {
+        translations.insert(key, jsonObj.value(key).toString());
+    }
+    return translations;
+}
+
+QString Translation::getText(const QString &key)
+{
+    return (m_translations.contains(key)) ? m_translations.value(key) : "missing translation";
```
mgautierfr: Maybe return key instead of "missing translation". At least user would see something roughly useful.

resources/js/_contentManager.js
```diff
@@ -18,6 +18,10 @@ function getIndexById(id) {
     return index;
 }
 
+function setTranslations(translations) {
+    app.translations = createDict(TRANSLATION_KEYS, translations);
```
mgautierfr: Is createDict function missing ?

resources/js/tools.js
```diff
@@ -0,0 +1,7 @@
+function createDict(keys, values) {
```
jetownfeve21: > Is createDict function missing ?

Is it better in another file like that ?

mgautierfr: Yes, we will probably have more code in common to contentManager and settings

mgautierfr:
I think we are good!!
If you're good, I let you remove the `[WIP]` and merge.

kelson42:
@mgautierfr So finally we have only one file per language or two?

mgautierfr:
We have moved most of our string to the json file.
We still have a Qt translation file to handle texts in the ui files. (I haven't check if we have texts there. But we want to keep it in case we add texts later anyway)

kelson42:
OK, this is not how I want it because we have two translations files. I see the JSON files in the PR, but where are the Qt string files? Which strings are still left? How do we secure they will be translated as well?

jetownfeve21:
@mgautierfr @kelson42 I added translation for ui files. I don't know if the translation of the ui generated by Qt (the context menu in web page for example) are handled by Qt translation system ?

mgautierfr:
>  but where are the Qt string files?

https://github.com/kiwix/kiwix-desktop/tree/master/resources/i18n

> Which strings are still left?

All `<string>Foo</string>` in those files https://github.com/kiwix/kiwix-desktop/tree/master/ui

> How do we secure they will be translated as well?

Upload the Qt translation files to wikitranslate as well

> I don't know if the translation of the ui generated by Qt (the context menu in web page for example) are handled by Qt translation system ?

They are generated with lupdate and lrelease : https://github.com/kiwix/kiwix-desktop/blob/master/kiwix-desktop.pro#L170-L171

src/about.cpp
```diff
@@ -10,8 +11,19 @@ About::About(QWidget *parent) :
 {
     ui->setupUi(this);
     auto htmlText = ui->aboutText->toHtml();
```
mgautierfr, line 13: Ho!!!

This make me realize we already have a system to load different resources depending on the lang.
The QTResource system can load different resources depending of the lang (https://doc.qt.io/qt-5/resources.html)
This is already what we do for this file. The about text is translated to fr here (https://github.com/kiwix/kiwix-desktop/tree/master/resources/texts/fr) and depending of the lang, qt will load the english version or the french version if we are in french locale (because of this https://github.com/kiwix/kiwix-desktop/blob/master/resources/kiwix.qrc#L76-L78).

We could use the same system to load the translation in the json file.
And keep this whole htmlText in one file instead of integrate it in the json file.

jetownfeve21: > We could use the same system to load the translation in the json file.
> And keep this whole htmlText in one file instead of integrate it in the json file.

Not sure to understand

mgautierfr: Qt already has a system to load a resource file specific to a lang. This is already use for the `about.html` resource. We have a "default" one in english (in resources/texts/about.html) and one specific to french in resources/texts/fr/about.html.
We could have the same things for the json file containing the translation. We could have a json file (texts.json) in resources/i18n and a texts.json per lang (resources/i18n/<lang>/texts.json). This way we simply would have to ask to qt to load the resource and it will load the right resource for us.

But saying that, it wouldn't protect us from a not up to date translation (with missing key).
Let's move with this PR. We have to think a bit more to see how (and if) we integrate the resource system.

jetownfeve21: OK I see, I'll open a ticket about that

Issue #206, opened 2019-07-15
Incorrect handling of hyperlinks for articles with slash(es) in the URL


Jaifroid:
In relation to https://github.com/kiwix/kiwix-js/pull/498 (and see in particular comment https://github.com/kiwix/kiwix-js/pull/498#issuecomment-511320797), it seems that kiwix-desktop_windows_x64_2.0-beta5 is not dealing correctly with articles that are in a subdirectory. This currently affects Stackexchange ZIMs, whereby clicking on a hyperlink (e.g. to a tag/category) from an article is currently broken in the desktop application ("Your file was not found").

Please note that this has interactions with https://github.com/openzim/mwoffliner/issues/726 . For example, the current landing page of `wikivoyage_en_all_novid_2019-06.zim` has this URL:

`Wikivoyage%3AOffline_reader_Expedition/Home_page`

Note the forward slash. Hyperlinks to other pages on this page should have an extra `../` in front of them, but they do not currently (and they currently work, nevertheless, in Kiwix Desktop). That issue has been fixed in mwoffliner https://github.com/openzim/mwoffliner/issues/726 , so that when a new ZIM is produced using the same landing page, Kiwix Desktop will probably be unable to follow the link correctly (as is currently the case with subdirectories in Stackexchange ZIMs). Therefore, it will need fixing in Kiwix Desktop (and other readers based on libzim, I think). _EDIT: Although wikivoyage_en_all_novid_2019-07.zim works in Kiwix Desktop beta, it is because of https://github.com/openzim/mwoffliner/issues/904, and not because this issue has actually been fixed._

@kelson42, we could do with some advice on how to handle this situation in Kiwix JS. If we fix the Stackexchange issue, we break Kiwix JS working with the current (buggy) Wikivoyage. If we don't fix, we will probably have a broken Wikivoyage on the next release of Wikivoyage, if it uses the same landing page URL. So catch 22.

Pinging @mossroy so he is aware of this issue.

Jaifroid:
To reproduce this issue with Stackexchange ZIMs:

* Open a Stackexchange ZIM. I used `mathoverflow.net_eng_all_2018-08.zim`
* Go to any question page
* Click on a tag on that page. You will get "Your file was not found".

[...]

PR #390, opened 2020-03-25
Correctly handle redirection in zim file.

### LINKED ISSUE #384: Links broken in Sotoki ZIM files
Steps:
* Download https://download.kiwix.org/zim/stack_exchange/3dprinting.stackexchange.com_en_all_2020-01.zim
* Click on the first question
* Click on the first question in the "Related" box (top right)

I get:
* Blanck page

I want:
* The proper article

Remark: Things work fine with kiwix-serve

mgautierfr:
Sotoki zim file use redirection to article not in the same "directory".
So it breaks relative urls.

Instead of providing the content of the target entry, we must redirect
to the target entry. So, relative url are not broken.

Fix #384

### DIFF
--- a/src/urlschemehandler.cpp
+++ b/src/urlschemehandler.cpp
@@ -42,10 +42,16 @@ UrlSchemeHandler::handleContentRequest(QWebEngineUrlRequestJob *request)
             return;
         }
     }
-    try {
-        entry = entry.getFinalEntry();
-    } catch (kiwix::NoEntry&) {
-        request->fail(QWebEngineUrlRequestJob::UrlNotFound);
+    if (entry.isRedirect()) {
+        try {
+            entry = entry.getFinalEntry();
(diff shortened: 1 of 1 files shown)


src/urlschemehandler.cpp
```diff
@@ -42,12 +42,19 @@ UrlSchemeHandler::handleContentRequest(QWebEngineUrlRequestJob *request)
...
         }
     }
+    bool isRedirect = entry.isRedirect();
     try {
         entry = entry.getFinalEntry();
     } catch (kiwix::NoEntry&) {
         request->fail(QWebEngineUrlRequestJob::UrlNotFound);
         return;
     }
+    if (isRedirect) {
```
jetownfeve21: We only use this variable here. Maybe we could use `entry.isRedirect()` directly.

jetownfeve21:
fix #206 too

PR #400, opened 2020-04-07
Don't use `gt` in the contructor.

mgautierfr:
The `gt` function get the instance of the kiwixApp.

So we cannot call this function in the constructor of kiwixApp itself.
Move all the (complex) initialization step in another method call after
the kiwixApp is created.

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -37,30 +37,6 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
     m_appTranslator.load(QLocale(), "kiwix-desktop", "_", ":/i18n/");
     installTranslator(&m_appTranslator);
     
-    try {
-        mp_downloader = new kiwix::Downloader();
-    } catch (exception& e) {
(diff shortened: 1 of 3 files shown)


mgautierfr:
@jetownfeve21 I've just found this problem trying to adapt the new tag sytem to the translation system.

src/kiwixapp.cpp
```diff
@@ -88,6 +64,33 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
...
+    setStyle(QStyleFactory::create("Windows"));
+    QFile styleFile(":/css/style.css");
+    styleFile.open(QIODevice::ReadOnly);
+    auto byteContent = styleFile.readAll();
+    styleFile.close();
+    QString style(byteContent);
+    setStyleSheet(style);
+
 
     createAction();
```
jetownfeve21, line 95: The `gt` function is also used in the `createAction` method

mgautierfr: `createAction` is called in the `init` method.

jetownfeve21: Ah yes my bad

PR #404, opened 2020-04-12
Add content type filter

### LINKED ISSUE #280: Implement filter by content-type in Library
See also #279

jetownfeve21:
fix #280

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -345,6 +345,12 @@ void ContentManager::setCurrentCategoryFilter(QString category)
     emit(filterParamsChanged());
 }
 
+void ContentManager::setCurrentContentTypeFilter(QList<ContentTypeFilter*>& contentTypeFilters)
+{
+    m_contentTypeFilters = contentTypeFilters;
+    emit(filterParamsChanged());
+}
+
 void ContentManager::updateLibrary() {
     if (m_local) {
         emit(pendingRequest(false));
@@ -378,7 +384,6 @@ QStringList ContentManager::getBookIds()
     std::vector<std::string> tags;
     if (m_categoryFilter != "all" && m_categoryFilter != "other") {
         tags.push_back("_category:"+m_categoryFilter.toStdString());
-        filter.acceptTags(tags);
     }
     if (m_categoryFilter == "other") {
         for (auto& category: S_CATEGORIES) {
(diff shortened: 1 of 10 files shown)


jetownfeve21:
There is some ui problems for me but I think it's better to improve it in another PR because this concerns the entire contentManagerSide bar
For example the visual difference between the buttons that display the filters and the filters themselves aren't clear in my opinion

kelson42:
@jetownfeve21 You remember we had remove/hide the select box for that a few months ago? You should reuse it I guess (maybe this is already the case).

jetownfeve21:
Yes the button was here but did nothing

src/contentmanager.cpp
```diff
@@ -388,6 +393,39 @@ QStringList ContentManager::getBookIds()
         }
         filter.rejectTags(tags);
     }
+
+    if (m_contentTypeFilter != "all") {
+        auto filters = m_contentTypeFilter.split(";");
+        for (auto &filter : filters) {
+            auto f = filter.split(" : ");
+            if (f[0] == gt("videos") && f[1] == gt("yes")) {
```
mgautierfr, line 401: Please do not do a test of the translated text.
This is somehow a golden rule. You don't know what will be the translation. Maybe in some language to English words are translated in the same text. (I doubt it could happen in this context but ... golden rule).

What we want here is probably a map, mapping a contentType ("videos", "pictures" or better, enum values) to a boolean.

jetownfeve21:
@mgautierfr Is it better with the implementation of QCheckBoxes like this ?

mgautierfr:
Yes, far better. I think we should use it also for the language and the category (but this is another improvement for another PR)

Few things :
- Remove the commented code.
- `m_contentTypeFilter` isn't use anymore.
- If I deactivate all filter by clicking on them, "All" should be made "bold" again.
- There is too many space between the filters.
- It would be better to move the filter buttons in a "frame/box" and hide/show this box when "ContentType" is clicked. Instead of having filter in the same level and hide/show them individually.

But the main idea is far better (and simpler) with checkboxes.

kelson42:
@mgautierfr We need a review from you here

mgautierfr:
Code seems good.
But there is a wrong alignment :
Capture d’écran du 2020-05-27 15-28-40

jetownfeve21:
@mgautierfr 

> But there is a wrong alignment :

fixed
image

PR #408, opened 2020-04-13
Searchbar ux improvement

### LINKED ISSUE #405: Improvements of the fulltext search
* Show all suggestions without scrollbar
* Should not load an article except if this is a perfect match
* Just clicking on "Enter" should make the fulltext search

jetownfeve21:
fix #405

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -103,5 +103,6 @@
     "wikispecies":"Wikispecies",
     "wikiversity":"Wikiversity",
     "wikivoyage":"Wikivoyage",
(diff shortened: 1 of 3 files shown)


mgautierfr:
This remove the ability to search for a word even if the word is a perfect match of the first title.
We want to allow the user to do a full text search for "Home" even if a "Home" page exists.

kelson42:
@jetownfeve21 On my side I'm not really satisfied that the fulltext search item moved from the bottom to the top level. There is not easy way to keep it on at the bottom? Not sure this is a blocker for that PR though.

PR #424, opened 2020-04-22
Single application kiwix-desktop

### LINKED ISSUE #147: Two Kiwix-Desktop instances can be run in parallel
I don't think this is really something we want... or we should guaranty that everything will works fine. Is that the case? A Singleton Windows seems more appropriate.

jetownfeve21:
fix #147

### DIFF
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -111,6 +111,7 @@ FORMS += \
     src/readinglistbar.ui \
     ui/localkiwixserver.ui
 
+include(subprojects/QtSingleApplication/src/qtsingleapplication.pri)
 CODECFORSRC = UTF-8
 
 isEmpty(PREFIX) {
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -15,12 +15,14 @@
 #include <QMessageBox>
 
 KiwixApp::KiwixApp(int& argc, char *argv[])
-    : QApplication(argc, argv),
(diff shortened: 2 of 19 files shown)


jetownfeve21:
@mgautierfr not sure if the qt solution is installation is correct like that

jetownfeve21:
https://code.qt.io/cgit/qt-solutions/qt-solutions.git/tree/qtsingleapplication/INSTALL.TXT
In the INSTALL.TXT, they recommend to use it as a library but we can compile the sources directly too

kelson42:
@mgautierfr @jetownfeve21 Could we please move on this PR?

mgautierfr:
> In the INSTALL.TXT, they recommend to use it as a library but we can compile the sources directly too

They said it is preferable for licensing issue, but we don't have this problem. The singleappliaction base app is bsd, we can use it in a gpl project.

On the opposite side, we need to have binaries for linux and windows (and potentially with different versions). And there is no dll in the PR. It would be preferable to compile the code all the time and use it directly (without intermediate dll/so).

jetownfeve21:
> On the opposite side, we need to have binaries for linux and windows (and potentially with different versions). And there is no dll in the PR. It would be preferable to compile the code all the time and use it directly (without intermediate dll/so).

That's true, it compiles the code all the the time now

mgautierfr:
Have you tested it on windows ? (I haven't)

mgautierfr:
Then we are good.

PR #440, opened 2020-05-04
Fluid suggestions list UX

### LINKED ISSUE #265: Handle suggestions list properly to have a fluid UX
To avoid to slow down the whole system. See https://github.com/kiwix/kiwix-desktop/issues/43#issuecomment-529819752 for more details about the idea.

All of this needs to be revamped/tested with ZIM files with really a lot of articles like `wikipedia_en_all_mini_2020-04.zim`.

### LINKED ISSUE #428: Suggestion/Search broken in new tab (v2.0.1)
No suggestions/search available if new tab is open.

Opening new tab should empty the search box.

jetownfeve21:
- add a delay before searching suggestions
  A timer of 100ms is started each time the text is edited, if the timer
  time-outs the suggestion's search is executed.

- execute the search in another thread
  The suggestion's search is handled by the worker "SuggestionListWorker" in dedicated thread.
  The semaphore ensures that there is only one worker at a time.
  The "m_actualSearch" is updated by the main thread with the last search requested. Its access is 
  protected by a mutex.
  The "checkAbort" method is called many times to cancel the search if a new one has been 
  launched meanwhile :
    if the original text (given in parameter) is different from "m_actualSearch", it means that a new 
    search has been launched meanwhile.
  The worker sends the suggestions list and the corresponding urls list to the main thread.
  The main thread displays them.

fix #265
and fix #428 too

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -1,10 +1,10 @@
 #include "searchbar.h"
 
 #include <QCompleter>
-#include <QTimer>
 #include <QFocusEvent>
 
 #include "kiwixapp.h"
+#include "suggestionlistworker.h"
 
 SearchButton::SearchButton(QWidget *parent) :
     QPushButton(parent),
@@ -62,6 +62,8 @@ SearchBar::SearchBar(QWidget *parent) :
     m_completer(&m_completionModel, this),
     m_button(this)
 {
+    mp_typingTimer = new QTimer(this);
+    mp_typingTimer->setSingleShot(true);
     setPlaceholderText(gt("search"));
     m_completer.setCompletionMode(QCompleter::UnfilteredPopupCompletion);
     m_completer.setCaseSensitivity(Qt::CaseInsensitive);
@@ -75,13 +77,15 @@ SearchBar::SearchBar(QWidget *parent) :
     QString style(byteContent);
     m_completer.popup()->setStyleSheet(style);
 
-    connect(this, &QLineEdit::textEdited, this, &SearchBar::updateCompletion);
+    qRegisterMetaType<QVector<QUrl>>("QVector<QUrl>");
+    connect(mp_typingTimer, &QTimer::timeout, this, &SearchBar::updateCompletion);
     connect(KiwixApp::instance(), &KiwixApp::currentTitleChanged,
             this, &SearchBar::on_currentTitleChanged);
     connect(this, &QLineEdit::textEdited, this,
             [=](const QString &text) {
                 m_searchbarInput = text;
                 m_returnPressed = false;
+                mp_typingTimer->start(100);
     });
     connect(this, &QLineEdit::textChanged, this,
             [=](const QString &text) {
@@ -92,6 +96,18 @@ SearchBar::SearchBar(QWidget *parent) :
     connect(this, &QLineEdit::returnPressed, this, [=]() {
         m_returnPressed = true;
     });
+}
+
+void SearchBar::hideSuggestions()
+{
+    m_completer.popup()->hide();
+}
+
+void SearchBar::clearSuggestions()
+{
+    QStringList empty;
+    m_completionModel.setStringList(empty);
+    m_urlList.clear();
 }
 
 void SearchBar::on_currentTitleChanged(const QString& title)
@@ -117,7 +133,7 @@ void SearchBar::focusInEvent( QFocusEvent* event)
     if (event->reason() == Qt::ActiveWindowFocusReason ||
         event->reason() == Qt::MouseFocusReason) {
         connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
-        this, &SearchBar::openCompletion);
+        this, QOverload<const QModelIndex &>::of(&SearchBar::openCompletion));
     }
     QLineEdit::focusInEvent(event);
     m_button.set_searchMode(true);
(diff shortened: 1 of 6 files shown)


src/searchbar.cpp
```diff
@@ -75,13 +77,37 @@ SearchBar::SearchBar(QWidget *parent) :
...
+    SuggestionListWorker *worker = new SuggestionListWorker;
+    worker->moveToThread(&m_suggestionThread);
+    connect(&m_suggestionThread, &QThread::finished, worker, &QObject::deleteLater);
+    connect(this, &SearchBar::searchSuggestion, worker, &SuggestionListWorker::searchSuggestions);
+    connect(worker, &SuggestionListWorker::suggestionTitle, this, [=](const QString& title) {
+        m_suggestionsList.append(title);
+    });
+    connect(worker, &SuggestionListWorker::suggestionUrl, this, [=](const QUrl& url) {
+        m_urlList.append(url);
+    });
```
mgautierfr, line 89: It is better to have only one signal here to ensure that both `m_suggestionList` and `m_urlList` are sync.
There is no reason that title and url are "sent" separatly.

src/suggestionlistworker.cpp
```diff
@@ -0,0 +1,46 @@
...
+void SuggestionListWorker::searchSuggestions(const QString& text)
+{
+    auto currentWidget = KiwixApp::instance()->getTabWidget()->currentWebView();
+    auto qurl = currentWidget->url();
+    qInfo() << "Search bar url is " << qurl;
+    auto currentZimId = qurl.host().split(".")[0];
+    auto reader = KiwixApp::instance()->getLibrary()->getReader(currentZimId);
+    QUrl url;
+    url.setScheme("zim");
+    if (reader) {
```
mgautierfr, line 18: Try with `QThread::sleep(3);` here.

jetownfeve21:
@mgautierfr I'm not sure the way I stop the worker when a new search is launched is fine, I don't find another way to do it

jetownfeve21:
fix https://github.com/kiwix/kiwix-desktop/issues/428 also

kelson42:
@mgautierfr I think @jetownfeve21 needs a bit of guidance here.

kelson42:
@jetownfeve21 Please comment/resolve the discussions which should.

jetownfeve21:
I'm just waiting for @mgautierfr 's review, specially about the way I cancel previous search when a new one is running

src/suggestionlistworker.cpp
```diff
@@ -0,0 +1,79 @@
...
+  m_semaphore(1)
+{
+}
+
+bool SuggestionListWorker::checkAbort(const QString& text)
+{
+    QMutexLocker locker(&m_mutex);
+    if (text != m_actualSearch) {
+        emit(searchAborted());
+        m_semaphore.release();
```
mgautierfr, line 17: Please move the release of the resource (semaphore) in the same function/place that acquire it.

src/suggestionlistworker.h
```diff
@@ -0,0 +1,34 @@
...
+    void setActualSearch(const QString& text);
+    bool isRunning() { return (m_semaphore.available() == 0); }
+
+signals:
+    void searchFinished(const QStringList& suggestions, const QVector<QUrl>& urlList);
+    void searchAborted();
+
+private:
+    QSemaphore m_semaphore;
+    QMutex m_mutex;
```
mgautierfr, line 30: I'm not sure about what the semaphore and the mutex are protecting.. and why two type of thread synchronization ?
Can you add some comments ? (And potentially rename them to a more explicit name)

mgautierfr:
After reading the QThread documentation (https://doc.qt.io/qt-5/qthread.html), it seems that the signal is queued and it is handle by the suggestionThread only when the previous signal is ended (or aborted). So it seems that we don't need a semaphore because only one "code" is running in the same time (but I may be wrong).
However, it is probable that most of the time to search for suggestion is taken only by the method `searchSuggestionsSmart`. If this method takes X seconds to complete, the thread will always spend X seconds in this method and there is no way to abort the thread while we are in. So, in case of long (and multiple) searches, we will spend time in the suggestionThread to search for all suggestions and block new ones. At least we will not block the UI but the suggestions will not come quicker.

I would move to the option of reimplementing run (as proposed in the documentation) instead of using the signal mechanism.
For each suggestionsearch, we would spawn a new thread. We initialize with the data (text to search, reader, empty vector for the result) and with a token (in simple incrementing integer) and launch the search code in the new thread.
If there is a previous search, we quit it (or terminate it, if it's safe to do it).
The search code do the search and fill the result's vectors.
The main thread is connected to the `finish` signal. When it is emited, the main thread check that the token is the good one and if yes, take the results from the vector. There is no synchronization to do here as we never read/write the value in the same time.
We may end to have several search in the same time (but it is thread safe to create several searcher) but the last search will be not delayed by previous one (and we will use only the last one).

mgautierfr:
To be exhaustive here, there are three optimizations in this PR:
- Run the suggestion search only 100ms after the last key input.
- Run the suggestion search in a different thread than the ui.
- Run the suggestion search, cancelling previous running search.

If we are on a slow computer (you can simulate this with a `QThread::sleep(3);` just after `searchSuggestionsSmart`) and the user were searching for `home`:
- Before this PR, we would do 4 searches (one for each key stroke), taking 12 seconds and blocking the ui for 3 seconds each time.
- With this PR, if the user enter `home` quickly, we do only one search, taking 3 seconds and not blocking the ui. This is a huge improvement.
- If the user enter `ho` quickly, pause, and finish with `me`, we will do 2 searches, taking 6 seconds. We will not block the ui but we will have result only 6 seconds after. All my explanation about the thread cancellation is about this case.

The technical thing about the semaphore/mutex has to be fixed but we may accept the PR without the cancellation and do it in another PR.

kelson42:
@jetownfeve21 Any news here?

mgautierfr:
Everything seems good. I've just rebase on `origin/master`

kelson42:
Great, works really good for me. Good job!

PR #442, opened 2020-05-06
Change download directory setting UX

### LINKED ISSUE #437: Changing download directory does not create it if necessary
On Kiwix Desktop 2.0.1 with Linux Mint 19.3 whenever I try to change the download directory (F12 -> Download directory) a alert dialog with the message `Invalid download path` appears.
image

jetownfeve21:
- remove "apply" button
- kiwixPort and zoomFactor are saved onchange
- download directory has a confirm dialog

fix #437

### DIFF
--- a/resources/css/_settingsManager.css
+++ b/resources/css/_settingsManager.css
@@ -3,7 +3,7 @@ html, body {
     margin: auto;
     height: 100%;
     position: relative;
-    width: 75%;
+    width: 80%;
     overflow: hidden;
 }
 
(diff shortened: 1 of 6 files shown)


jetownfeve21:
I cannot reproduce the case on my computer, @mgautierfr do you have this crash ?

mgautierfr:
@kelson42 I cannot reproduce your bug.

If you agree with the change, you can merge this PR.

kelson42:
@mgautierfr I have achieved to reproduced a few days ago, let me try again.

kelson42:
This is the result of the last testing:
image

I believe it would be better if:
* The path on the left
* Considering the box is read-only, no border

I was not able to reproduce the crash but had an other problem:
* Create new directory /tmp/4242
* Chosing it
* Dialog displays properly: "The new download directory path will be: /tmp/4242"
* So far so good
* Now trying to choose "/tmp", I get no error but the directory is not updated from "/tmp/4242" to "/tmp".

mgautierfr:
> The path on the left
Considering the box is read-only, no border

Ok

> Now trying to choose "/tmp", I get no error but the directory is not updated from "/tmp/4242" to "/tmp".

Be sure that you don't just go to the parent directory in the browser. If you do just that, the directory "4242" is still selected (even if you are not "in it").

kelson42:
> > Now trying to choose "/tmp", I get no error but the directory is not updated from "/tmp/4242" to "/tmp".
> 
> Be sure that you don't just go to the parent directory in the browser. If you do just that, the directory "4242" is still selected (even if you are not "in it").

Indeed, so we have only the styling problem.

mgautierfr:
Should be good now.

PR #445, opened 2020-05-11
Sort language list

### LINKED ISSUE #407: Language listing is incorrect
It looks like the language sorting is done by alphabetical order of each language's self description or ISO code (e.g. in the example below Welsh is listed under C because it is `cymraeg/cy` and German is `deutsch/de`:

image

This is not something an average user not privvy to these languages would/could know, and therefore to them they are as good as not listed.

The correct listing order should be that of the interface language (English in the case at hand, with Welsh among W languages and German along with other Gs).

jetownfeve21:
fix #407


kelson42:
@jetownfeve21 From a user perspective, seems to work. Thank you!

PR #456, opened 2020-05-26
add "open link in web browser" action in context menu

### LINKED ISSUE #326: Open in a new tab does not work fine with external links

jetownfeve21:
reimplement the contextMenuEvent : if the link hovered is an external link
it removes from the standard context menu the "openLinkinNewTab" action and
and add the "open link in web browser"

fix #326

### DIFF
--- a/src/webview.cpp
+++ b/src/webview.cpp
@@ -1,6 +1,7 @@
 #include "webview.h"
 
 #include <QDesktopServices>
+#include <QMenu>
 #include <QAction>
 #include <iostream>
 #include "kiwixapp.h"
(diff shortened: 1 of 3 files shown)


src/webview.cpp
```diff
@@ -71,6 +72,24 @@ void WebView::wheelEvent(QWheelEvent *event) {
...
+    if (!m_linkHovered.isEmpty() && !m_linkHovered.startsWith("zim://")) {
+        pageAction(QWebEnginePage::OpenLinkInNewTab)->setVisible(false);
+        auto openLinkInWebBrowserAction = new QAction(gt("open-link-in-web-browser"));
+        menu->insertAction(pageAction(QWebEnginePage::DownloadLinkToDisk) ,openLinkInWebBrowserAction);
+        connect(openLinkInWebBrowserAction, &QAction::triggered, this, [=](bool checked) {
+            Q_UNUSED(checked);
+            QDesktopServices::openUrl(m_linkHovered);
+        });
+    } else {
+        pageAction(QWebEnginePage::OpenLinkInNewTab)->setVisible(true);
```
mgautierfr, line 88: This will also set the `OpenLinkInNewTab` visible if the link is empty.

src/webview.cpp
```diff
@@ -71,6 +72,24 @@ void WebView::wheelEvent(QWheelEvent *event) {
     }
 }
 
+void WebView::contextMenuEvent(QContextMenuEvent *event)
+{
+    auto menu = this->page()->createStandardContextMenu();
+    pageAction(QWebEnginePage::OpenLinkInNewWindow)->setVisible(false);
```
mgautierfr, line 78: This should be set only once at application initialization.

src/webview.cpp
```diff
@@ -71,6 +72,24 @@ void WebView::wheelEvent(QWheelEvent *event) {
...
+    if (!m_linkHovered.isEmpty() && !m_linkHovered.startsWith("zim://")) {
+        pageAction(QWebEnginePage::OpenLinkInNewTab)->setVisible(false);
+        auto openLinkInWebBrowserAction = new QAction(gt("open-link-in-web-browser"));
```
mgautierfr, line 81: Why not create the action with all our action in kiwixapp ?

src/webview.cpp
jetownfeve21: It will not work this way. I use this method to get the default contextmenu that Qt will use (it is different depending on the situation) and then add the action

kelson42:
@mgautierfr Would you be please be able to complete this PR and merge?

kelson42:
I have rebased the branch

mgautierfr:
I changed back to the previous @jetownfeve21 version where the `openLinkInWebBrowserAction` is created on the fly when we create the context menu.

Is it useless to create it globally as we need to add it to the menu every time and connect/disconnect a lambda for each menu.
I've just add a `deleteLater` when the menu is deleted to avoid memory leaks

Issue #358, opened 2020-01-24
Replace tab's title ellpsis by a fade-out


kelson42:
We currently use an ellipsis (`...`) if the title of the tab is to big to fit to the tab. But we should have a fade out instead. Like in the mockup, see:
![image](https://user-images.githubusercontent.com/1029718/73072233-ff828280-3eb4-11ea-85af-3cc977a7ebc2.png)

yashasvi-goel:
Can I work on this issue?

kelson42:
@yashasvi-goel Yes :)

yashasvi-goel:
I have written the CSS for completing this task. I'm unable to replace the CSS in place of setElideMode().

kelson42:
> I have written the CSS for completing this task. I'm unable to replace the CSS in place of setElideMode().

@mgautierfr @jetownfeve21 Any idea?
@yashasvi-goel Please make a WIP PR.

kelson42:
fade-out seems to need to reimplement our own analog of QTabBar so it can worth 10+ hours. I doubt if it really that important

Alternatively, we can do a research to see maybe that change will come in new versions of Qt; or file a bug otherwise, hope someone else will implement it

Also, I can search if any other open source application have it. Firefox and Chromium doesn't use Qt, but probably there Konquerror in KDE, or some other web-browser or some other Qt-based app have those tabs

I looked into the implementation of QTabBar, and didn't see other way as to copy and tune that code

kelson42:
@asashnov  I don't understand why you would need to subclass `QTabBar`. What is the problem exactly with the current PR? You apply a style to the `QTabBar::tab` which is only the text part of tab, so if the ext is short, the `QTabBar::tab` has a short width.. and so the fadout applies in the middle of the tab?

asashnov:
@kelson42 Yes, It would not be a big problem if fadeout applies to the end of the tab.  But tab text is aligned to center now, and it is not customizable with `text-align: left` because it is not supported by QTabBar (only by QPushButton and QProgressBar).
https://doc.qt.io/qt-5/stylesheet-reference.html#list-of-properties

kelson42:
@asashnov Can you measure the width of the text and apply conditionaly the fadeout style?

asashnov:
@kelson42 two things to consider here:
* on Mac and Linux the text in tab is centered (see screenshot)
* we cannot apply fadeout style to the individual tab, only to all of them.

[...]

PR #457, opened 2020-05-27
Change tab title ellipsis (...)

jetownfeve21:
I don't find a way to apply a fade out effect when the title is too long to fit in the tab, I just succeed to remove the 3 dots ellipsis but even this solution isn't great :
It uses the `fontMetrics().elidedText(QString text, Qt::ElideRight, int width)` method to get the elided string, if the string has the ellipsis '...', it removes it. But this method has an `int width` parameter that represent the space in pixel for the text, I don't find a proper way to have it for a tab, so it's hard coded now, which is bad.

Without the fade out effect, the result looks more like a bug :
image

What we have currently:
image
 
#358

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -113,15 +113,27 @@ void TabBar::openUrl(const QUrl& url, bool newTab)
     webView->setUrl(url);
 }
 
+QString TabBar::textWithoutEllipsis(QString text)
+{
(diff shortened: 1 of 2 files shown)


kelson42:
@jetownfeve21 You should not apply a fade-out effect when the title is too long IMO. You should apply a CSS which always apply at the right of the tab and if the title is too long it will have a user visible impact.

kelson42:
Closing this PR as so far we don't know how to handle this.

PR #463, opened 2020-06-18
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates


mgautierfr:
The `about-source-code` should contains a link using `{{GITHUB_URL}}` with the last version (master) of kiwix-desktop. (see https://github.com/kiwix/kiwix-desktop/commit/913b5abff66b42a34a6fdc5ec55431183c8bb284).
(Same for the `about-version` and `about-reporting-issue`)

It seems that the translation of `source-code` has been keep and used for `about-source-code`.
It should be better to not have translation at all if the translation is not ok. At least we will display the English text.

@nikerabbit, do you know if this is something transitory or if there is something wrong on someone side ?

Abijeet:
@mgautierfr  - This was a mistake on our side. When these changes were being processed by Translatewiki.net, I decided to rename `source-code` to `about-source-code` (same with the other two messages) to maintain the translation history. Looking at the changes now, I realize that the previous message and the current message are incompatible.

I apologize, and will keep this in mind in the future.

Abijeet:
@mgautierfr - We've manually deleted the translations, can you please check if everything looks OK?

resources/i18n/nl.json
```diff
@@ -0,0 +1,49 @@
...
+	"size": "Grootte",
+	"date": "Datum",
+	"open": "Openen",
+	"delete": "Verwijderen",
+	"download": "Downloaden",
+	"resume": "Hervatten",
+	"pause": "Pauzeren",
+	"cancel": "Annuleren",
+	"apply": "Toepassen",
+	"about-version": "Versie : {{VERSIE}}",
```
mgautierfr, line 37: The `{{VERSION}}` token must not be translated.

resources/i18n/pl.json
```diff
@@ -105,5 +114,9 @@
 	"details": "Cały artykuł",
 	"yes": "tak",
 	"no": "nie",
-	"no-filter": "brak filtra"
+	"no-filter": "brak filtra",
+	"open-link-in-web-browser": "Otwórz link w przeglądarce",
+	"download-dir-dialog-title": "Czy na pewno chcesz zmienić katalog pobierania?",
+	"download-dir-dialog-msg": "Nową ścieżką katalogu pobieranych plików będzie: {{DIRECTORY}}",
```
mgautierfr, line 120: It would be better to keep the `\n` after the `:`.

resources/i18n/pt-br.json
```diff
@@ -66,17 +67,16 @@
...
-	"report-issue-3/3": "Mencione a versão e a data de compilação no problema.",
-	"libraries": "Bibliotecas",
+	"about-kiwix-desktop-title": "Kiwix Desktop",
+	"about-kiwix-desktop-description": "O Kiwix permite que você tenha toda a Wikipédia em mãos, onde quer que você vá! Em um barco, no meio do nada ou na cadeia, o Kiwix fornece acesso a todo o conhecimento humano. Você não precisa da Internet, tudo está armazenado no seu computador.",
+	"about-learn-more-about-kiwix": "Saiba mais sobre o Kiwix",
+	"about-release-title": "Lançamento",
+	"about-source-code": "Este software é lançado sob os termos da GNU General Public License version 3. Veja o código fonte em <a href='{{GITHUB_URL}}'>Github</a>.",
+	"about-version": "Versão : {{VERSION}}",
+	"about-reporting-problem-title": "Reportar um problema",
+	"about-report-issue": "Para abrir um problema, visite <a href='{{TRACKER_URL}}'>Kiwix-desktop issue tracker</a> e abra um problema descrevendo o seu problema.",
```
mgautierfr, line 77: Shouldn't we translate `Kiwix-desktop issue tracker` ?

resources/i18n/fr.json
```diff
@@ -68,17 +69,15 @@
...
-	"report-issue-2/3": "et ouvrez un ticket en décrivant votre problème.",
-	"report-issue-3/3": "Veuillez mentionner la version et la date de compilation dans le rapport.",
-	"libraries": "Bibliothèques",
+	"about-kiwix-desktop-title": "Bureau Kiwix",
+	"about-kiwix-desktop-description": "Kiwix vous permet d'avoir tout Wikipedia à portée de main où que vous alliez! Sur un bateau, au milieu de nulle part ou en prison, Kiwix vous donne accès à l'ensemble du savoir humain. Vous n'avez pas besoin d'internet, tout est conservé dans votre ordinateur.",
+	"about-learn-more-about-kiwix": "En apprendre plus sur Kiwix",
+	"about-release-title": "Release",
+	"about-version": "Version : {{VERSION}}",
+	"about-reporting-problem-title": "Rapporter un problème",
+	"about-report-issue": "Pour déclarer une faute, allez sur <a href='{{TRACKER_URL}}'>Suivi des problèmes du bureau Kiwix</a> et ouvrez un ticket en décrivant votre problème.",
```
mgautierfr, line 78: "Faute" seems a bit wrong here. I would translate it by "bogue" (or simply keep keep the previous translation : "Pour ouvrir un rappor de bogue" instead of "Pour déclarer une faute".

resources/i18n/nl.json
mgautierfr: I've fix this with a specific commit. But it should be fixed "upstream".

Abijeet: Imported the message into translatewiki.net

I've also added a validator for variables (strings in the format `{{VAR}}`) to ensure that they remain in translations as is.

validator

PR #483, opened 2020-07-06
Repro build

### LINKED ISSUE #476: Don't encode current date in build
Encoding the current date in the build makes it non-reproducible - see https://reproducible-builds.org/

They recommend getting the date from something consistent, e.g. the timestamp of the git commit: `git log -1 --pretty=%ct` (see https://reproducible-builds.org/docs/source-date-epoch/).

However, as pointed out in https://github.com/kiwix/kiwix-desktop/issues/470#issuecomment-652401126 always shelling out to git (for this and the pre-existing GIT_VERSION) is problematic because packages for Debian/Fedora are building from a tarball, not git.

My suggestion is to do something like:
```bash
BUILD_DATE="$(date --utc --date="@${SOURCE_DATE_EPOCH:-$(git log -1 --pretty=%ct)}")"
BUILD_VERSION="${BUILD_VERSION:-$(git describe --dirty=* --tags --always)}"
```

Then it would be the packagers responsibility to set $SOURCE_DATE_EPOCH and $BUILD_VERSION, and for normal builds, it falls back to using the info from git.

cc @xvitaly

mgautierfr:
`GIT_VERSION` and `BUILD_DATE` are only displayed to the user.
It was useful pre-release to identify the exact version of the build in
bug report.

Now that we are releasing kiwix-desktop, we don't need the
`BUILD_DATE` and we must use the exact version instead of the git version.

Fix #476

Need PR kiwix/kiwix-lib#376 to compile.


xvitaly:
Please remove also `/static` parameter from https://github.com/kiwix/kiwix-desktop/blob/master/kiwix-desktop.pro#L127-L129

PR #484, opened 2020-07-08
Better about text for better translation.

### LINKED ISSUE #399: Use Qt resource system to load the right translation file
Like @mgautierfr said in this [PR](https://github.com/kiwix/kiwix-desktop/pull/387) : 
> Qt already has a system to load a resource file specific to a lang. This is already use for the about.html resource. We have a "default" one in english (in resources/texts/about.html) and one specific to french in resources/texts/fr/about.html.
We could have the same things for the json file containing the translation. We could have a json file (texts.json) in resources/i18n and a texts.json per lang (resources/i18n//texts.json). This way we simply would have to ask to qt to load the resource and it will load the right resource for us.
> But saying that, it wouldn't protect us from a not up to date translation (with missing key).
Let's move with this PR. We have to think a bit more to see how (and if) we integrate the resource system.

mgautierfr:
Update a bit the "about.html" text to not split sentences in several parts
and allow translator to translate the whole sentence at once.

Fix #399


mgautierfr:
@nikerabbit you opinion on this would be appreciable.

Nikerabbit:
Looks good to me.

PR #485, opened 2020-07-08
Free fonts

### LINKED ISSUE #474: Fonts appear to be non-free
Both of the bundled fonts, Segoe UI and Ubuntu appear to be under a non-free license, incompatible with Kiwix's GPL license.

Segoe UI is a font licensed by Microsoft and as far as I can tell, not available under any free license. [Wikipedia](https://en.wikipedia.org/wiki/Segoe#Related_fonts) mentions that Selawik is a similar font by Microsoft, that is released under the free Open Font License. See https://github.com/microsoft/Selawik.

Ubuntu font is licensed under the Ubuntu Font License, which Debian and Fedora both consider to be non-free. See <https://en.wikipedia.org/wiki/Ubuntu_(typeface)#Ubuntu_Font_Licence>. I didn't immediately find any listing of similar free fonts.

kelson42:
Fixes #474... but honestly I see no visual difference!


kelson42:
@legoktm Not sure what happened here, but the problems you have reported should be fixed now.

mgautierfr:
What is the license of the Selawik font ?
Where does it come from ?

We should probably add license file for the font.

kelson42:
> We should probably add license file for the font.

@mgautierfr Yes. What looks better: in the about? As a separate file in the font directory?

legoktm:
The license/source of the font is https://github.com/microsoft/Selawik

> @mgautierfr Yes. What looks better: in the about? As a separate file in the font directory?

Can we do both? A text file in the fonts/Selawik directory and then a line on the about page?

mgautierfr:
Both is good.
The about page is for the users of the application.
The license file is more a legal issue for developers/packagers/lawyers.

The license of the font is SIL Open Font License. (https://github.com/microsoft/Selawik/blob/master/LICENSE.txt)
This license mandate us to at least add the license file in our sources (from what I understand, IANAL).

legoktm:
> The license of the font is SIL Open Font License. (https://github.com/microsoft/Selawik/blob/master/LICENSE.txt)
> This license mandate us to at least add the license file in our sources (from what I understand, IANAL).

https://tldrlegal.com/license/open-font-license-(ofl)-explained says we need to include the full text of the license.

From the license itself:
```
2) Original or Modified Versions of the Font Software may be bundled,
redistributed and/or sold with any software, provided that each copy
contains the above copyright notice and this license. These can be
included either as stand-alone text files, human-readable headers or
in the appropriate machine-readable metadata fields within text or
binary files as long as those fields can be easily viewed by the user.
```

So including a copy of the license in the source will ensure we stay compliant with the license for tarballs and including it in about.html ("human-readable headers") covers binary distributions.

kelson42:
@mgautierfr @legoktm Done.

kelson42:
@mgautierfr Ok, I believe now this is good and the CI pass. Sorry for the many back and forth.

legoktm:
Yay! I don't know if you all have other upcoming changes planned, but if we could do a patch release for this and the reproducible builds fixes, I'm ready to upload it to Debian.

kelson42:
@legoktm A new release will happen in the next days.

PR #506, opened 2020-08-05
debian: Remove aria2 build dependency

### LINKED ISSUE #505: aria2 deb package dependence seems wrong
Kiwix and Libkiwix aria2c deps seems wrong. At least on Kiwix-Desktop it should be a runtime dependence, and not a source dependence.

legoktm:
The dependency is inherited from libkiwix, once
https://github.com/kiwix/kiwix-lib/pull/402 is merged.

Fixes #505.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -6,8 +6,7 @@ Build-Depends: debhelper-compat (= 13),
  pkg-config,
  qtbase5-dev,
  qtwebengine5-dev,
- libkiwix-dev (>= 9.3.0),
- aria2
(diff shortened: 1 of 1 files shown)


mgautierfr:
Aria2 is technically wrapped by kiwix-lib but is used by kiwix-desktop. Other tools using kiwix-lib doesn't use aria2.

It is more a matter of debian policy, but you may not want to force the dependency at kiwix-lib level as you mandate all others tools to have aria2 installed even if they don't use it.

Issue #517, opened 2020-08-30
Allow to delete from local library only


kelson42:
Currently we can only delete from the fs and from the library.

Open after talk on #515

kelson42:
At the core of my thoughs about this problem we have the idea that if the content has been added manually (so download via a third party application), then it might be bold to just delete the file. But otherwise, this seems the best behaviour (at least per default).

I see these possible solutions:
* By deleting offer a choice between deleting from the library and deleting both from the library + fs.
* Delete the library+fs in case the file has been download via download manager, otherwise (external loading) delete only from the library

We should not forget to consider #479

kelson42:
@juuz0 yes, but onky if downloaded via third party app. otherwise does not make sense IMO. All good and logical for you?

kelson42:
I was sure I had close that one. Now we can delete or keep in trash for both:
* If external ZIM, then trash will allow to restore in original location
* If internal ZIM, then it will be restored and as the internal download directory is monitored file will reappeared automgically in the local library of Kiwix.

Nothing more is needed IMO.

PR #531, opened 2020-10-21
Do not delete zim file from fs when we delete book from the library.

### LINKED ISSUE #479: Deleting a ZIM file should move it to the trash can
When I delete a ZIM file from the library, it is also deleted from the hard disk at the same time! And the basket remains empty!
**Ubuntu 18.04 Unity** + **Kiwix 2.0.1-1** (*.appimage):
Kiwix-deleting-ZIM

The situation is exactly the same in the operating system **Ubuntu 18.04 MATE** + **Kiwix 2.0.1-1** (*.appimage).

mgautierfr:
Fix #479

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -78,6 +78,7 @@
     "kiwix-server-running-message":"The Kiwix Server is running and can be accessed in the local network at :",
     "kiwix-server-description":"Starting a server allows other computers in the local network to access your Kiwix library with a standard web browser.",
     "fullscreen-notification":"You are now in full screen mode. Press ESC to quit!",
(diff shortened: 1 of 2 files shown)


kelson42:
@mgautierfr what does this PR exactly?

mgautierfr:
> what does this PR exactly? 

With this PR, when a user remove a book in kiwix-desktop, the book is removed from the library but the zim file is not deleted from file system. Instead, we display a message saying the file(s) has not been deleted and telling the path where to find the zim files.

> IMO the way it works on iOS is the proper way.

I don't know how iOS is working :)

kelson42:
@mgautierfr I think you mean this PR should fix https://github.com/kiwix/kiwix-desktop/issues/517... an I have put there what I expect for a solution

kelson42:
Superseeded by #531

Issue #320, opened 2019-12-02
UI not respecting user/OS color theme


Boruch-Baum:
The linux appimage for kiwix 2.0-rc4 does not respect the user's choice of theme (eg. gtk/qt)

kelson42:
I confirm that on Ubuntu 20.04. Just changing the the default color theme has no impact on Kiwix, it should.

asashnov:
Seems we need to include some Qt style plugin into our AppImage:

https://wiki.archlinux.org/index.php/Uniform_look_for_Qt_and_GTK_applications#QGtkStyle

kelson42:
@asashnov I don't believe this is a problem (only) with appimage. This is a general problem (with self-compiled, deb packages, flatak), right?

wengxt:
https://github.com/kiwix/kiwix-desktop/blob/cb5cfdceaf7840c456de67814e7395a7dd6c75a5/src/kiwixapp.cpp#L70

This is the part of the code that causing problem on Linux.
1. It forces the "Windows" qt widget style on Linux, while normally this is configured by desktop
2. The next few lines set an custom stylesheet, this overrides the widget background color, but it does not set the text color.
https://github.com/kiwix/kiwix-desktop/blob/cb5cfdceaf7840c456de67814e7395a7dd6c75a5/resources/css/style.css#L2
here's my suggestion:
1. only set widget style for windows system.
2. As for color:
   - Only set style related to color on windows
   - Or set the widget text color in stylesheet to totally override system color (but I guess people still going to complain like this issue)
   - Or have an option: "Use system color", if disabled, apply the color related css.

Here's a screenshot under KDE's Breeze dark after I tried to remove the setStyle and all color related settings in css.
<img width="522" alt="图片" src="https://user-images.githubusercontent.com/259684/106376095-b4f8d900-6346-11eb-82c6-5def500a6bec.png">

wengxt:
Another option is only set stylesheet against kiwix's main window (still if you want to use the same css, the text color need to be set otherwise it still use the system text color which may conflict with the background), instead of applying to QApplication.

kelson42:
@wengxt Thank you for your in-depth analysis.

The ticket does not really details how Kiwix is expected to work:
* First it should have a night/day theme depending how the OS is configured. Not sure this option exists on both windows/linux. Not sure either if it fundamentaly differs from have a custom theme.
* We have a UI theme for Kiwix, as far as possible it should be used.
[...]

PR #538, opened 2020-11-21
Fixes #358 - fade-out instead elide text (trim with...) in tabs

### LINKED ISSUE #358: Replace tab's title ellpsis by a fade-out

asashnov:
Fixes #358 - fade-out instead elide text (trim with...) in tabs

This is how it looks like with the change:
Screenshot from 2020-11-21 21-38-03

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -7,6 +7,7 @@
 #include <QToolButton>
 #include <QToolTip>
 #include <QCursor>
+#include <QPainter>
 
 #define QUITIFNULL(VIEW) if (nullptr==(VIEW)) { return; }
 #define QUITIFNOTCURRENT(VIEW) if((VIEW)!=currentWidget()) {return;}
@@ -18,12 +19,13 @@ TabBar::TabBar(QWidget *parent) :
 {
     QTabBar::setDrawBase(false);
     setTabsClosable(true);
-    setElideMode(Qt::ElideRight);
+    setElideMode(Qt::ElideNone);
     setDocumentMode(true);
     setFocusPolicy(Qt::NoFocus);
     setIconSize(QSize(30, 30));
     connect(this, &QTabBar::currentChanged, this, &TabBar::onCurrentChanged);
     auto app = KiwixApp::instance();
+
     connect(app->getAction(KiwixApp::NewTabAction), &QAction::triggered,
             this, [=]() {
                 this->createNewTab(true);
@@ -125,7 +127,30 @@ void TabBar::setTitleOf(const QString& title, ZimView* tab)
         auto url = QUrl(title);
         setTabText(mp_stackedWidget->indexOf(tab), url.path());
     } else {
-        setTabText(mp_stackedWidget->indexOf(tab), title);
+        int idx = mp_stackedWidget->indexOf(tab);
+        setTabToolTip(idx, title);
+
+        // This logic is taken from the implementation:
+        // <QTDIR>/5.12.6/Src/qtbase/src/widgets/widgets/qtabbar.cpp
+        // void QTabBar::initStyleOption(QStyleOptionTab *option, int tabIndex) const
+        QStyleOptionTab tab;
+        initStyleOption(&tab, idx);
+        QRect textRect = style()->subElementRect(QStyle::SE_TabBarTabText, &tab, this);
+
+        // but instead of eliding text as QTabBar::initStyleOption() does
+        // we cut it and store the flag if it was cut
+        QString cut = fontMetrics().elidedText(title, Qt::ElideRight, textRect.width());
+        // strip ... from the end (this three dots are one char)
+        if (cut.size() < title.size()) {
+            cut = cut.mid(0, cut.size() - 1);
+            // set flag that the text was too long, was cut and this tab
+            // need 'fade out' effect while drawing
+            setTabData(idx, QVariant::fromValue(true));
+        }
+        else {
+            setTabData(idx, QVariant::fromValue(false));
+        }
+        setTabText(idx, cut);
     }
 }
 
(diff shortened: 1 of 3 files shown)


kelson42 (changes requested):
@asashnov Thank you for this promising PR but:
* It applies to any title, even if very smal ("Bird" for example) and in fact there should be not fading because there is a lot of space.
* I wonder if it would not be better practive to have the CSS in the stylesheet file, instead of having it mixed with the C++ code in `setStyleSheet()`.

kelson42:
@asashnov Is that ready to review again?

asashnov:
@kelson42 No, this PR is not ready yet.

I moved the style to .css file, but didn't found a way to use separate styles in QTabBar for each tab. Seems it should involve to create our own implementation.

Also, I'll think about implementing ::paint() method to render kind of fade-out effect on top of standard rendering.

mgautierfr:
> Also, I'll think about implementing ::paint() method to render kind of fade-out effect on top of standard rendering.

You've arrived to the same conclusion than me. The css rule you apply is nice but it applies to all texts, even if they are not too long.
We haven't found a way to do it without reimplementing the `paint` method of the widget.
This is obviously doable, but we need to understand the low level parts of Qt.

kelson42:
And this is not possible to have the width of the widget and do the math to apply the CSS only if necessary (width big enough)?

asashnov:
@kelson42  I found a way to get QRect for text in the tab.
But stylesheet cannot be applied only to specified tabs (see my comment in the issue).

Instead, we can fillRect() in `paintEvent()` and use a gradient, the tab background color, and the alpha channel to imitate the effect we had with this CSS.

Screenshot from 2020-12-09 01-00-50

In this screenshot the background color for the active tab is wrong, and no alpha channel gradient, but the position of QRect is right.
Also, flags if we need this trick or not is also set correctly.

asashnov:
Unfortunately, there is NO an easy way to get the background color of selected tab like:

```c
  QColor c0;
  c0 = tab.palette.color(QPalette::Active, QPalette::Button);
  c0 = QApplication::style().standardPalette().color(QPalette::Active, QPalette::Button);
  c0 = palette().color(QPalette::Active, QPalette::Button);
```
(I tried many variants also)

Since we install our own custom style from 'style.css'
`QApplication::style()` returns `QStyleSheetStyle` object.

This class is a private part of Qt (no public header for it).

Moreover, in its implementation (see Qt/5.12.6/Src/qtbase/src/widgets/styles/qstylesheetstyle.cpp)
it declares local for this .cpp file class `QRenderRule`
which actually takes case of getting the exact CSS rule for selected tab.

Also, I saw in the sources there is really no way to specify tab by tab number.

For more details see:

```c
void QStyleSheetStyle::drawControl(ControlElement ce, const QStyleOption *opt, QPainter *p,
            const QWidget *w) const
{
...
case CE_TabBarTabShape:
  ...
   QRenderRule subRule = renderRule(w, opt, PseudoElement_TabBarTab);


`QStyle` and its derived classes have no interface for getting colors back.
They can only draw a control or its part on some `QPaintDevice`.

So we have no easy way here to get back our 'white' color we set our style.css file:

```css
QTabBar::tab:selected {
  background-color: white;
  border-bottom: 2px solid white;
}
```

So I decided to just hardcode it in the code and write comments that it should be in sync in both places.

An alternative solution may involve 'picking' this active tab background color from the widget
after it has been painted by `QTabBar::paintEvent()`.

kelson42:
@mgautierfr Please review

kelson42:
@asashnov I have rebased on git `master` and test it. Work perfectly I believe. Might need a bit of rework at the time we will support the "dark mode" and user modes... but this is an other topic. Thank you very much.

kelson42 (approved):
On the user side looks good. Waiting @mgautierfr review for the code.

asashnov:
Rebased, squashed, long commentary removed (it still here in this 'Conversation').

src/tabbar.cpp
```diff
@@ -254,3 +279,65 @@ void TabBar::mousePressEvent(QMouseEvent *event)
...
+
+        QRect textRect = style()->subElementRect(QStyle::SE_TabBarTabText, &tab, this);
+
+        QRect tail = textRect;
+        int offset = textRect.width() * 0.8;
+        if (tab.direction == Qt::RightToLeft) {
+            tail.setWidth(textRect.width() - offset);
+        } else {
+            tail.setLeft(tail.x()+offset);
+            tail.setWidth(textRect.width() - offset);
```
mgautierfr, line 311: I would move the two `tail.setWidth(textRect.width() - offset);` out of the if/else.

src/tabbar.cpp
```diff
@@ -254,3 +279,65 @@ void TabBar::mousePressEvent(QMouseEvent *event)
...
+            continue;
+
+        QStyleOptionTab tab;
+        initStyleOption(&tab, i);
+
+        QRect textRect = style()->subElementRect(QStyle::SE_TabBarTabText, &tab, this);
+
+        QRect tail = textRect;
+        int offset = textRect.width() * 0.8;
+        if (tab.direction == Qt::RightToLeft) {
```
mgautierfr, line 307: It is nice to handle the LeftToRight text but I haven't succeed to test it.
I've tried to run kiwix-desktop in Arabic and use arabic content but the fade-out is always on the right.
(But the title is correctly cut on the left)

mgautierfr (changes requested):
We are definitively on the good way but it seems there is a small issue with RightToLeft text.

asashnov:
Qt::LayoutDirection - Specifies the direction of Qt's layouts and text handling (https://doc.qt.io/qt-5/qt.html#LayoutDirection-enum)

When we open arabic wiki, and see arabic in tab's title, we change just 'text layout'.

But my code checked 'widget layout' of the application- that was a problem.

Fixed: In the current implementation the text layout direction is given, not application widgets layout.

src/tabbar.cpp
asashnov: Rewritten without `offset` variable, not needed: `QRect::moveRight()` is really what I needed.

src/tabbar.cpp
asashnov: `tab.direction` was 'widget layout', not `text layout`. Fixed in the current revision (8759c81e).

asashnov:
> We are definitively on the good way but it seems there is a small issue with RightToLeft text.

Fixed in 7252b1b3 (Feb 15, 2021)

mgautierfr (approved):

kelson42 (approved):

mgautierfr:
We are good :tada:

Just mentioning that the the need to keep the selected_bg_color in sync with the css will be a bit complex with https://github.com/kiwix/kiwix-desktop/issues/320.
We cannot just have a const value. But this is a problem for #320 :)

PR #557, opened 2020-12-10
Fix printsupport in .pro file

### LINKED ISSUE #556: Missing "printsupport" in .pro
When packaging this for Guix, compilation failed with

```
src/kiwixapp.cpp:12:10: fatal error: QPrinter: No such file or directory
```

Adding QT+= printsupport to the .pro file fixed it, as suggested in
https://stackoverflow.com/questions/19145763/qt-cannot-open-include-file-qprinter.

Ambrevar:
This may be required to avoid errors like:

    src/kiwixapp.cpp:12:10: fatal error: QPrinter: No such file or directory

Fixes #556.

### DIFF
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -6,6 +6,7 @@
 
 QT       += core gui network
 QT       += webenginewidgets
+QT       += printsupport
 
 CONFIG += link_pkgconfig
 


kelson42:
@Ambrevar Thank you!

PR #568, opened 2020-12-30
Settings tab will now be activated if already opened.

### LINKED ISSUE #545: Usability: Settings tab should be activated if already opened
Found in current 'master' (830e19bb)

## Steps to reproduce:

1. Run the program
1. Open any file from library (it was 11 Gb EN Wikipedia in my case)
1. click + in tabbar to create an empty tab
1. From ... menu choose `Settings` to open Settings tab
1. Switch back to the opened file (Wikipedia in my case)
1. Again go to ... menu and choose Settings

## Actual result

Nothing happens, the current tab is still Wikipedia tab.

## Expected result

Once we have `Settings` tab already opened, it should be activated (switched in), so the user can see it.

Saket-07:
Fix #545. The settings tab will now be activated as the current index will now change to the index of settings tab if it is already opened and a user opens settings from the three dot drop down menu.

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -50,6 +50,7 @@ TabBar::TabBar(QWidget *parent) :
     connect(app->getAction(KiwixApp::SettingAction), &QAction::triggered,
             this, [=]() {
                 if (KiwixApp::instance()->getSettingsManager()->isSettingsViewdisplayed()) {
+                    setCurrentIndex(m_settingsIndex);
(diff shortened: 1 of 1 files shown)


kelson42:
@Saket-07 Thak you very much for you PR. It will be reviewd during next week. In the meantime, feel free to work on an other ticket.

mgautierfr:
Thanks for your PR !

Issue #541, opened 2020-11-25
Crash then typing in top search box while Settings are opened


asashnov:
kiwix-desktop 'master' (830e19bb)

Found while research #454

## Steps to reproduce

1. Right-top menu, Settings
2. While Settings tab is open, type any symbols into search box on top.

## Actual result

Segmentation fault.

## Expected result

Search request should have no effect while Settings tab is opened;
or this Search box should be disabled.

## Additional info

```
1   QWebEnginePage::url() const                                                                                                                                                   0x7ffff77a7724 
2   QWebEngineView::url() const                                                                                                                                                   0x7ffff77b3b77 
3   SearchBar::updateCompletion                                                                                                                            searchbar.cpp      158 0x5555555a0fb1 
4   QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (SearchBar:: *)()>::call(void (SearchBar:: *)(), SearchBar *, void * *) qobjectdefs_impl.h 136 0x5555555a3eb0 
5   QtPrivate::FunctionPointer<void (SearchBar:: *)()>::call<QtPrivate::List<>, void>(void (SearchBar:: *)(), SearchBar *, void * *)                       qobjectdefs_impl.h 169 0x5555555a3b7c 
6   QtPrivate::QSlotObject<void (SearchBar:: *)(), QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase *, QObject *, void * *, bool *)          qobject_impl.h     120 0x5555555a3575 
7   QMetaObject::activate(QObject *, int, int, void * *)                                                                                                                          0x7fffeeb15876 
8   QTimer::timeout(QTimer::QPrivateSignal)                                                                                                                                       0x7fffeeb21d17 
9   QTimer::timerEvent(QTimerEvent *)                                                                                                                                             0x7fffeeb21fe8 
10  QObject::event(QEvent *)                                                                                                                                                      0x7fffeeb1645b 
11  QApplicationPrivate::notify_helper(QObject *, QEvent *)                                                                                                                       0x7ffff046d70c 
12  QApplication::notify(QObject *, QEvent *)                                                                                                                                     0x7ffff0474b37 
13  QCoreApplication::notifyInternal2(QObject *, QEvent *)                                                                                                                        0x7fffeeaeaf68 
14  QTimerInfoList::activateTimers()                                                                                                                                              0x7fffeeb3cd4e 
[...]

PR #569, opened 2021-01-01
Don't update completion if Settings Page opened

HanYoung-uwu:
Fix https://github.com/kiwix/kiwix-desktop/issues/541
Can't find a way that is both elegant and doesn't require rewritten half of the logic. Check if Settings Page is open first. So far Settings Page is the only widget that can't be converted to WebView. This PR doesn't fix the root issue.

### DIFF
--- a/src/suggestionlistworker.cpp
+++ b/src/suggestionlistworker.cpp
@@ -14,6 +14,8 @@ void SuggestionListWorker::run()
     QVector<QUrl> urlList;
 
     auto currentWidget = KiwixApp::instance()->getTabWidget()->currentWebView();
+    if(!currentWidget)
+        return;
     auto qurl = currentWidget->url();
     auto currentZimId = qurl.host().split(".")[0];
     auto reader = KiwixApp::instance()->getLibrary()->getReader(currentZimId);
--- a/src/tabbar.h
+++ b/src/tabbar.h
@@ -25,7 +25,8 @@ public:
     ZimView* createNewTab(bool setCurrent);
     ZimView* widget(int index) { return (index != 0) ? static_cast<ZimView*>(mp_stackedWidget->widget(index)) : nullptr; }
     WebView* currentWebView() { auto current = mp_stackedWidget->currentWidget();
-                               if (mp_stackedWidget->currentIndex() == 0) return nullptr;
+                               if (mp_stackedWidget->currentIndex() == 0 ||
+                                   mp_stackedWidget->currentIndex() == m_settingsIndex) return nullptr;
                                return static_cast<ZimView*>(current)->getWebView();
                              }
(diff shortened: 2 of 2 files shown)


src/searchbar.cpp
```diff
@@ -154,6 +154,11 @@ void SearchBar::updateCompletion()
 {
     mp_typingTimer->stop();
     clearSuggestions();
+    // workaround for #541
+    // return if settings open
+    if (KiwixApp::instance()->getSettingsManager()->isSettingsViewdisplayed()) {
+        return;
+    }
     auto currentWidget = KiwixApp::instance()->getTabWidget()->currentWebView();
```
mgautierfr, line 162: This would be better to fix `currentWebView` implementation.

As you can see https://github.com/kiwix/kiwix-desktop/blob/master/src/tabbar.h#L27-L35, 
currentWidget return a nullptr if the current webview is the setting view (before casting it), but currentWebView don't.

And while you are here, there are few place where we use `currentWebView` method without checking the returned pointer. We should also fix this.

mgautierfr (changes requested):

src/searchbar.cpp
HanYoung-uwu: Fixed as you suggested. However, I only found one place where lacking nullptr safeguard.(there are some place that don't matter if the pointer is nullptr)

mgautierfr:
> there are some places that don't matter if the pointer is nullptr

It seems there is at least the fullscreen part which use the pointer returned by getCurrentWebView.
We should protect it.
But is may not be straightforward. The fullscreen functionality works by moving the webview (not the window) fullscreen. If there is no current web view (because we are in setting or catalog view), we have to define what to do (fullscreen something else, deactivate fullscreen, ...).
This may not be technically complex but we have to think a bit before.

I will open a issue for that. You may want to fix this (starting by discutting on the issue). But you don't have to. The bug exists independently of this PR.


---

As we are doing a PR, not yet merged into master, I prefer to simply remove the first commit (https://github.com/kiwix/kiwix-desktop/pull/569/commits/2380775d08cbfb9bb67ba593adb4028a411cac23) instead of revert it. Can you modify the branch ? I will merge after.

HanYoung-uwu:
> As we are doing a PR, not yet merged into master, I prefer to simply remove the first commit
I thought we can squash commits (a gitlab feature that I'm used to) :P

mgautierfr (approved):

mgautierfr:
Thanks a lot for your PR.

> I thought we can squash commits (a gitlab feature that I'm used to) :P

I prefer to not squash commits of a PR, it allows to keep a reference to the PR itself in the git history, and "regroup" the different commits of a PR together (without merging them).

kelson42:
@HanYoung-uwu Thank you for your PR!

Issue #498, opened 2020-07-23
Language filter does not apply on local library


Ubuntu-2020:
**Kiwix 2.0.4** (*.appimg).
Sorting by language in the library doesn't work.
I'm sorry that there is no picture. I have two files in my **Library**.
One ZIM file from the **Russian Wikipedia**. The second ZIM file is from the **English Wikipedia**.
In the side column, I chose **German**, then **English** language, and nothing happened.
I think that when choosing **German**, the two ZIM files mentioned above should have disappeared from the **library**.
And when choosing **English**, the ZIM file from the English Wikipedia should have appeared.
But nothing happened.

kelson42:
I confirm

HanYoung-uwu:
I'd see this as a feature not implemented rather than bug. ZIM file format doesn't store language information so there isn't a reliable way to determine the language of a local file. The current language filter set the URL query language parameter, the language information is stored remotely. However, it's still possible to parse the file name and try to determine the language, then implement the feature to filter the local files by language.

mgautierfr:
> ZIM file format doesn't store language information so there isn't a reliable way to determine the language of a local file

Yes, it is. You can access it from the reader (https://github.com/kiwix/kiwix-lib/blob/master/include/reader.h#L184) and from the book (https://github.com/kiwix/kiwix-lib/blob/master/include/book.h#L56)
The local library also handle it and you can filter it using the language (as for the remote library) (https://github.com/kiwix/kiwix-lib/blob/master/include/library.h#L99)

---

In fact, the remote library runs the exact same code that the local library, the "remote languages" are determined using the languages stored in the zim files.

HanYoung-uwu:
> > ZIM file format doesn't store language information so there isn't a reliable way to determine the language of a local file
> 
> Yes, it is. You can access it from the reader (https://github.com/kiwix/kiwix-lib/blob/master/include/reader.h#L184) and from the book (https://github.com/kiwix/kiwix-lib/blob/master/include/book.h#L56)
> The local library also handle it and you can filter it using the language (as for the remote library) (https://github.com/kiwix/kiwix-lib/blob/master/include/library.h#L99)
> 
> In fact, the remote library runs the exact same code that the local library, the "remote languages" are determined using the languages stored in the zim files.

Sorry, I must have missed it. I'll try to implement it.

PR #574, opened 2021-01-08
apply language filter to local library

HanYoung-uwu:
related: #498

Apply language filter to local library.
In the process, I found that not all language supported by Kiwix is included in static_content.cpp. For instance, Abkhaz language(abk in Kiwix) isn't included. One solution would be generating static_content.cpp on compile time with kiwix header file. Another thing is the default language isn't set to **All** so not all local books are displayed as default. Probably better to set default language to **All**.


kelson42:
@HanYoung-uwu I missed this PR. thank you. Asking for review.

HanYoung-uwu:
oops

Issue #571, opened 2021-01-04
Fix fullscreen if there is no current webview (zim)


mgautierfr:
The fullscreen functionality works by putting the current (zim) webview fullscreen.
But there is no safeguard if there is no current zim displayed and we end by passing a nullptr to the FullScreen class.

We have a technical bug (using a nullptr) and a functional point to fix (what to do if there is no current webview)

The simpler seems to deactivate the fullscreen if there is no current zim.
But then we have to deactivate the action, not just add a safeguard.
And we probably need to test corner use case (closing the webview while we are fullscreen, ...)

Or we may want to keep the fullscreen functionality if there is no current zim, but we must decide what is the correct behavior.

asashnov:
From the initial report:

> The fullscreen functionality works by putting the current (zim) webview fullscreen.
> ...
> we may want to keep the fullscreen functionality if there is no current zim, but we must decide what is the correct behavior

What I see now in 'master' (cb5cfdce) is not putting the current webview to fullscreen. Instead, full screen mode just illuminates the window header and border, but the entire window content is show: 'Search',  'Main menu', 'Full screen' buttons, tabbar. And it works as expected, without crashed in this mode.

So, if we are happy with the current 'master' (cb5cfdce) there is nothing to fix, all tabs, including Settings, are shown perfectly in this window mode.

Chrome and Firefox works differently in fullscreen mode:
* Firefox show tabbar and address bar when the mouse hit upper screen side.
* Chome just shows a round (X) button to exit from fullscreen mode, not showing tabbar, but it's still possible to switch between tabs with Ctrl+TAB hotkey.

kelson42:
It should work like for Firefox. Like @asashnov said, first, this is the the windows content which is put fullscreen, not the webview. I think this is OK, and therefore there is no point to deactivate the option if we there is no WebView visible (we display the library). But what needs to be done, is that the top bar with tab header need to disappear and only appear when we display a tab and we have the mouse pointer at the top. That way we could have a "real" article fullscreen.

asashnov:
@HanYoung-uwu  Do you have any changes to review now?

The only PR associated with this issue (#580) is closed.
And no changes in https://github.com/HanYoung-uwu/kiwix-desktop/tree/master

HanYoung-uwu:
> @HanYoung-uwu Do you have any changes to review now?
> 
> The only PR associated with this issue (#580) is closed.
[...]

PR #580, opened 2021-01-21
Fix fullscreen if there is no current webview (zim)

HanYoung-uwu:
Exit full screen if last tab closed or settings menu is bringing to the front. Disable full screen action if no tab opened or on settings menu.

#571

### DIFF
--- a/resources/org.kiwix.desktop.appdata.xml
+++ b/resources/org.kiwix.desktop.appdata.xml
@@ -26,13 +26,13 @@
   <update_contact>mgautier_AT_kymeria.fr</update_contact>
   <screenshots>
     <screenshot type="default">
-      <image>http://download.kiwix.org/screenshots/kiwix_desktop_library.png</image>
+      <image>https://download.kiwix.org/screenshots/kiwix_desktop_library.png</image>
     </screenshot>
     <screenshot>
-      <image>http://download.kiwix.org/screenshots/kiwix_desktop_remote_library.png</image>
+      <image>https://download.kiwix.org/screenshots/kiwix_desktop_remote_library.png</image>
     </screenshot>
     <screenshot>
-      <image>http://download.kiwix.org/screenshots/kiwix_desktop_ted.png</image>
+      <image>https://download.kiwix.org/screenshots/kiwix_desktop_ted.png</image>
     </screenshot>
   </screenshots>
 </component>
--- a/resources/texts/_contentManager.html
+++ b/resources/texts/_contentManager.html
@@ -40,7 +40,7 @@
       <details v-for="book in displayedBooks(books, displayedBooksNb)" class="book">
         <summary v-bind:id="book.id" class="tablerow book-summary">
           <span class="tablecell cell0">
-            <img v-if="book.faviconUrl" v-bind:src="'http://' + book.faviconUrl" />
+            <img v-if="book.faviconUrl" v-bind:src="'https://' + book.faviconUrl" />
             <img v-else-if="book.faviconMimeType" v-bind:src="'zim://' + book.id + '.favicon.meta'" />
           </span>
           <span class="tablecell cell1">
--- a/resources/texts/about.html
+++ b/resources/texts/about.html
@@ -4,7 +4,7 @@
 <img src="qrc:/icons/kiwix/scalable/kiwix-desktop.svg" height="240" width="240" style="float:right"/>
 <h1>{{KIWIX_DESKTOP_TITLE}}</h1>
 <p>{{KIWIX_DESKTOP_DESCRIPTION}}</p>
-<p><a href="http://kiwix.org">{{LEARN_MORE_ABOUT_KIWIX}}</a></p>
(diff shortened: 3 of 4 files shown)


src/kiwixapp.h
```diff
@@ -83,6 +83,7 @@ class KiwixApp : public QtSingleApplication
     QString getText(const QString &key) { return m_translation.getText(key); };
 
     bool isCurrentArticleBookmarked();
+    void setFullScreenDisable(bool disable) {mpa_actions[ToggleFullscreenAction]->setDisabled(disable);}
```
asashnov, line 86: I don't think we need to introduce this new method here. It's better (more readable, even longer) already used approach:
```c
KiwixApp::instance()->getAction(KiwixApp::ToggleFullscreenAction)
    ->setDisabled(true);
```

asashnov (changes requested):

asashnov (changes requested):
Adding a new (shortcut) method will likely decrease the code readability:
https://github.com/kiwix/kiwix-desktop/pull/580#discussion_r563719152

kelson42:
@HanYoung-uwu Are you ready for an other review?

src/tabbar.cpp
```diff
@@ -192,6 +192,10 @@ void TabBar::closeTab(int index)
     removeTab(index);
     webview->close();
     delete webview;
+
+    // exit full screen on last tab closed
+    if(this->count() <= 2 &&KiwixApp::instance()->getAction(KiwixApp::ToggleFullscreenAction)->isChecked())
```
kelson42, line 197: Please keep a space after the if keyword.

asashnov: and after &&

kelson42: done

kelson42:
@asashnov Is this PR good to merge in your opinion?

asashnov:
@HanYoung-uwu the code looks good, but I discovered strange (unwanted to me as a user) behaviour.

Steps to reproduce:
* run the program. Only library tab is shown;
* activate 'Main menu (...)' -> Settings

Actual result:
The settings tab is shown, but the program switched to fullscreen also.

Expected result:
The 'Settings' tab should be shown as usual, without the change of window size (current 'master' cb5cfdce is OK in that sense).

asashnov (changes requested):
Code seems good, but Main menu -> Settings forces the app go fullscreen (see https://github.com/kiwix/kiwix-desktop/pull/580#issuecomment-770775880)

asashnov (changes requested):
Commit message should contain Fixes #571  so GitHub can really associate this PR with the issue (the link in PR description works but is not making the association, it's just a mention of some issue to GitHub now).

And by the way, what we are going to achieve with this fix? As I understand, there was a crash, but it doesn't crash now:
https://github.com/kiwix/kiwix-desktop/issues/571#issuecomment-771032667

HanYoung-uwu:
> Commit message should contain Fixes #571 so GitHub can really associate this PR with the issue (the link in PR description works but is not making the association, it's just a mention of some issue to GitHub now).
> 
> And by the way, what we are going to achieve with this fix? As I understand, there was a crash, but it doesn't crash now:
> [#571 (comment)](https://github.com/kiwix/kiwix-desktop/issues/571#issuecomment-771032667)

I'm also confused by this, IIRC, we pass nullptr to fullscreen. But as that doesn't crash the app, this MR doens't really fixed anything. I wonder if I close this MR instead.

kelson42:
@HanYoung-uwu Why have you closed the PR?

HanYoung-uwu:
> @HanYoung-uwu Why have you closed the PR?

Bad git practice. I always forgot to open a new branch for MR. Then I force sync the master with upstream.

kelson42:
@HanYoung-uwu You should not open a new PR. You should modify your history in place in this current MR.

HanYoung-uwu:
> @HanYoung-uwu You should not open a new PR. You should modify your history in place in this current MR.

I'm not open a new PR, I mean I forgot I have this PR try to merge my master into this master. And I force reset my master to this master to sync. In the process unintentionally closed the PR.

kelson42:
@HanYoung-uwu First I would recommend you to not work on master. You should work in a branch. That would ease the whole operation. What the plan now?

HanYoung-uwu:
> @HanYoung-uwu First I would recommend you to not work on master. You should work in a branch. That would ease the whole operation. What the plan now?

> I always forgot to open a new branch for MR

This PR needs more discussion I guess. I'm not very sure what is the expected result. The Outcome of my edits is if in settings or blank view kiwix-desktop will auto exit from fullscreen mode and disable it.

kelson42:
@HanYoung-uwu I believe to have clarified the ticket. Look at Firefox to see how the top bar disappearance shoudl work.

HanYoung-uwu:
So the current Full Screen function is implemented via `showFullScreen` in mainwindow class. I understand that @kelson42 wants to full screen the webview not the whole application window. I propose to connect the fullscreen action to tabbar instead. That way, we can full screen the webview, thus essentially making the web content full screen. Reviews appreciated.

PR #581, opened 2021-01-21
apply language filter to local library

HanYoung-uwu:
related: #498

Apply language filter to local library.
In the process, I found that not all language supported by Kiwix is included in static_content.cpp. For instance, Abkhaz language(abk in Kiwix) isn't included. One solution would be generating static_content.cpp on compile time with kiwix header file. Another thing is the default language isn't set to All so not all local books are displayed as default. Probably better to set default language to All.

I forgot about this PR and force pushed to my fork, thus unintentionally closed the [previous](https://github.com/kiwix/kiwix-desktop/pull/574) PR. My apologies.

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -406,6 +406,8 @@ QStringList ContentManager::getBookIds()
 
     filter.acceptTags(tags);
     filter.query(m_searchQuery.toStdString());
+    if (m_currentLanguage != "*")
+        filter.lang(m_currentLanguage.toStdString());
 
     if (m_local) {
         filter.local(true);


src/contentmanager.cpp
```diff
@@ -406,6 +406,8 @@ QStringList ContentManager::getBookIds()
 
     filter.acceptTags(tags);
     filter.query(m_searchQuery.toStdString());
+    if(m_currentLanguage != QLatin1Char('*'))
+        filter.lang(m_currentLanguage.toStdString());
```
kelson42, line 410: I wonder why `QLatin1Char()` is necessary? please put a space after the `if`.

kelson42 (changes requested):

src/contentmanager.cpp
HanYoung-uwu: We're compare QString to QChar, QLatin1Char ensure the '*' is constructed at compile time, no runtime cost. Otherwise QChar will be contructed from '*' (char) at runtime.

mgautierfr: I don't see any clue that `QLatin1Char` is build as compile time. (https://doc.qt.io/qt-5/qlatin1char.html)
What is probably done, is that we are creating a QLatin1Char, converting it to a QChar (using the implicit constructor), converting it to a QString(using the implicit constructor) then comparing the QString using `operator!=`.

We should probably use `QStringLiteral` to construct a QString at compile time or even better `QLatin1String` as `operator!=` is overloaded for `QLatin1String` (https://doc.qt.io/qt-5/qstring.html#QStringLiteral)

HanYoung-uwu: Sorry for being misleading, you're right. Since the QString uses char * internally, and provides const char * overload for !=, I switched to just plain c char array.

mgautierfr: To be totally thorough, the `!=` operator convert the char* to a QString internally.
But we are not in a critical path, the conversion will have no impact. It is simpler to keep the plain `char*`.

mgautierfr (approved):

src/contentmanager.cpp
HanYoung-uwu: Qt doc says QString accepts const char * to avoid construct QString [there](https://doc.qt.io/qt-5/qlatin1string.html#details)
But it [constructs QString nonetheless](https://doc.qt.io/qt-5/qstring.html#operator-not-eq-1)

:man_facepalming:

PR #589, opened 2021-02-04
Https library link

### LINKED ISSUE #587: library catalog downloaded over HTTP
Originally filed at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=981466 - can these URLs be switched to use HTTPS? Or is there a reason they still use HTTP?

----
I noticed that the data feed is not downloaded using https, so network
attackers could modify the data feed to change my choice of downloads
to something I didn't want to download.

Also most of the datasets point at http instead of https URLs even
though the servers do support https. It would be good if kiwix had a
list of download servers that support https and then always use https
to contact those download servers.

```
   $ kiwix-desktop 
   QSocketNotifier: Can only be used with threads started with QThread
   Compiled with Qt Version  5.15.1
   Runtime Qt Version  5.15.2
   add widget
   
   (kiwix-desktop:1410327): GLib-GObject-WARNING **: 23:11:12.766: The property GtkSettings:gtk-fallback-icon-theme is deprecated and shouldn't be used anymore. It will be removed in a future version.
   Downloading "http://library.kiwix.org:80/catalog/search?lang=eng&count=0"
   session saved
```

kelson42:
Fixes #587

### DIFF
--- a/resources/org.kiwix.desktop.appdata.xml
+++ b/resources/org.kiwix.desktop.appdata.xml
@@ -26,13 +26,13 @@
   <update_contact>mgautier_AT_kymeria.fr</update_contact>
   <screenshots>
     <screenshot type="default">
-      <image>http://download.kiwix.org/screenshots/kiwix_desktop_library.png</image>
+      <image>https://download.kiwix.org/screenshots/kiwix_desktop_library.png</image>
     </screenshot>
     <screenshot>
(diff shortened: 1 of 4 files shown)


kelson42:
@mgautierfr Merging, too trivial, you can let me know if you see a problem.

PR #606, opened 2021-02-25
Fixes #604 - Prepend / to an article path to make QUrl parser happy

### LINKED ISSUE #604: Search / Suggestions  are totally broken with latest (dev) libzim 7.0.0
I always get a blank page if:
* click on any suggestion
* click on a suggestion which is a redirect
* make a fulltext search

asashnov:
Fixes #604 - Prepend / to an article path to make QUrl parser happy

QUrl::setPath() should be called with an absolute path, starting with /

See my findings in https://github.com/kiwix/kiwix-desktop/issues/604#issuecomment-786147789 and https://github.com/kiwix/kiwix-desktop/issues/604#issuecomment-786158804

But don't be hurry to merge it, it may be not in line with the current architecture. Probably this starting / should be added at 'kiwix-lib' layer, not in 'kiwix-desktop', see my comment  
https://github.com/kiwix/kiwix-desktop/issues/604#issuecomment-786181829

### DIFF
--- a/src/suggestionlistworker.cpp
+++ b/src/suggestionlistworker.cpp
@@ -26,7 +26,8 @@ void SuggestionListWorker::run()
         kiwix::SuggestionsList_t suggestions;
         reader->searchSuggestionsSmart(m_text.toStdString(), 15, suggestions);
         for (auto& suggestion: suggestions) {
-            url.setPath(QString::fromStdString(suggestion[1]));
+            QString path = QString("/") + QString::fromStdString(suggestion[1]);
+            url.setPath(path);
             suggestionList.append(QString::fromStdString(suggestion[0]));
             urlList.append(url);
         }


kelson42:
@mgautierfr I wonder a bit why we have this problem here and not in kiwix-serve.

kelson42:
@asashnov This PR probably will build a buggy version of Kiwix-Desktop if we use older version of libkiwix. Do we have a way somewhere to require version 10+ of libkiwix?

src/suggestionlistworker.cpp
```diff
@@ -26,7 +26,9 @@ void SuggestionListWorker::run()
         kiwix::SuggestionsList_t suggestions;
         reader->searchSuggestionsSmart(m_text.toStdString(), 15, suggestions);
         for (auto& suggestion: suggestions) {
-            url.setPath(QString::fromStdString(suggestion[1]));
+            QString path(QString::fromStdString(suggestion[1]));
+            path.push_front('/');
+            url.setPath(path);
```
kelson42, line 31: @asashnov Why exactly we need a leading `/`? I believe we should probably later deal with that, at the moment the URL has to be open. Do you know where is that?

mgautierfr:
> @mgautierfr I wonder a bit why we have this problem here and not in kiwix-serve.

Because we don't use the suggestion url/path in kiwix-serve. We do a search with the title, search which will be transform to a redirect if an article with the exact same title is found.

> Why exactly we need a leading /? I believe we should probably later deal with that, at the moment the URL has to be open. Do you know where is that?

The standard for uri (https://tools.ietf.org/html/rfc2396) says that the part after a authority must be a absolute path and that an absolute path must start by a '/'.

---

Libzim do not store url as key. Entries are reference by a key we call `path`. libzim handle this path as a array of byte and don't care about `/` in it. So the notion of absolute/relative path doesn't exist in libzim.
This is why the new api change `url` to `path`. And libzim do not add a leading `/` as there is no need/sense for it.

On kiwix-desktop side, we are definitively in url context. The path is a absolute path, interpreted in the context of a uri and so must start with a `/`.

kiwix-lib is somehow in the middle. I would say that `reader::searchSuggestion` return a list of entry corresponding to the suggestion. No a list of url. So we must not add a `/` in kiwix-lib
(At least not here. In the server part, if the suggestion return urls, we would have to it).
(But this is a API break. Next release of kiwix-lib must bump the major version).

kelson42:
@mgautierfr So? Libkiwix seems OK, what about this PR? Does the appending of the leading slash is done properly and at the right place?

mgautierfr:
We already add a leading "/" here :
- https://github.com/kiwix/kiwix-desktop/blob/master/src/kiwixapp.cpp#L251
- https://github.com/kiwix/kiwix-desktop/blob/master/src/urlschemehandler.cpp#L37-L38

Each time a different way. (It would be nice to homogenize that, maybe by having a helper method ?)

We also set the QUrl path here : https://github.com/kiwix/kiwix-desktop/blob/master/src/readinglistbar.cpp#L58
But we don't add a leading "/". It would be nice to check it is working.

kelson42:
Yes, even if this bug looks silly, I find it quite symptomatic of an architecture problem. We should secure this as much as possible.

asashnov:
Activating a link from 'Reading list' works.

I changed my code to be in the same style with those other two places with adding /: Use QString("/") instead of my .push_front('/')

kelson42 (approved):

kelson42:
I don’t think this is how the problem shoukd ve fixed, but for the moment this is probably good enough.

PR #616, opened 2021-03-19
Reorder include to avoid header conflict beetween Xapiand and Qt

kelson42:


mgautierfr:
You don't give me the chance to review that.
A good practice to to always have file `foo.cpp` include associated `foo.h` first.
This way we can be sure that all needed include are correctly include in the header file.

kelson42:
@mgautierfr OK and we should fix things following your recommendation https://github.com/kiwix/kiwix-desktop/pull/616#issuecomment-802741973

PR #621, opened 2021-04-03
History shortcuts

### LINKED ISSUE #618: Implement history navigation shortcuts
* ALT+left key to go back in the history (if possible)
* ALT+right key to go forward in the history (if possible)

kelson42:
Fixes  #618

### DIFF
--- a/src/kiwixapp.h
+++ b/src/kiwixapp.h
@@ -45,6 +45,8 @@ public:
         ZoomInAction,
         ZoomOutAction,
         ZoomResetAction,
+        HistoryBackAction,
+        HistoryForwardAction,
         HelpAction,
         FeedbackAction,
         ReportBugAction,
--- a/src/topwidget.cpp
+++ b/src/topwidget.cpp
@@ -10,25 +10,21 @@
 TopWidget::TopWidget(QWidget *parent) :
     QToolBar(parent)
 {
-    mp_historyBackAction = new QAction(this);
-    mp_historyBackAction->setIcon(QIcon(":/icons/back.svg"));
-    mp_historyBackAction->setText(gt("back"));
-    mp_historyBackAction->setToolTip(gt("back"));
-    mp_historyBackAction->setEnabled(false);
-    connect(mp_historyBackAction, &QAction::triggered, [](){
+    auto app = KiwixApp::instance();
+
+    QAction *back = app->getAction(KiwixApp::HistoryBackAction);
+    connect(back, &QAction::triggered, [](){
         KiwixApp::instance()->getTabWidget()->triggerWebPageAction(QWebEnginePage::Back);
     });
-    addAction(mp_historyBackAction);
-    widgetForAction(mp_historyBackAction)->setObjectName("backButton");
-    mp_historyForwardAction = new QAction(this);
-    mp_historyForwardAction->setIcon(QIcon(":/icons/forward.svg"));
-    mp_historyForwardAction->setText(gt("forward"));
-    mp_historyForwardAction->setToolTip(gt("forward"));
-    mp_historyForwardAction->setEnabled(false);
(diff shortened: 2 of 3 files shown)


src/kiwixapp.h
```diff
@@ -45,6 +45,8 @@ class KiwixApp : public QtSingleApplication
         ZoomInAction,
         ZoomOutAction,
         ZoomResetAction,
+        HistoryBackAction,
+        HistoryForwardAction,
```
mgautierfr, line 49: You declare a `HistoryForwardAction` but never define it.

You never use `HistoryBackAction` and `HistoryForwardAction`.

src/topwidget.cpp
```diff
@@ -13,6 +13,7 @@ TopWidget::TopWidget(QWidget *parent) :
     mp_historyBackAction->setIcon(QIcon(":/icons/back.svg"));
     mp_historyBackAction->setText(gt("back"));
     mp_historyBackAction->setToolTip(gt("back"));
+    mp_historyBackAction->setShortcut(QKeySequence(Qt::ALT + Qt::Key_Left));
```
mgautierfr, line 16: Instead of changing this "local" action, we should create, configure and use the global action.

mgautierfr (changes requested):
The idea behind the global action (in kiwixapp) is to define some ... global actions. They are configured once on then we can simply get the already configured action and add it to a widget (toolbar, menu, ...).

For exemple, this is the case for `ToggleReadingListAction`. We configure it in kiwixapp and add it to the `TopWidget` using `addAction(KiwixApp::instance()->getAction(KiwixApp::ToggleReadingListAction));`

I don't remember why we don't use global action for history back and forward, but if we go this way we must go fully.

src/topwidget.cpp
kelson42: OK, but is that a good idea to put an icon on a global action?

mgautierfr: Yes, this is the icon associated with the action (the same way than text and tooltip)
If the "use case" of the action is made without icon, it will not be used.

kelson42:
@asashnov Could you please complete this PR?

kelson42:
@asashnov If this ready for a new review? If yes, please re-request one.

src/topwidget.cpp
```diff
@@ -10,27 +10,21 @@
...
-    widgetForAction(mp_historyBackAction)->setObjectName("backButton");
-    mp_historyForwardAction = new QAction(this);
-    mp_historyForwardAction->setIcon(QIcon(":/icons/forward.svg"));
-    mp_historyForwardAction->setText(gt("forward"));
-    mp_historyForwardAction->setToolTip(gt("forward"));
-    mp_historyForwardAction->setShortcut(QKeySequence(Qt::ALT + Qt::Key_Right));
-    mp_historyForwardAction->setEnabled(false);
-    connect(mp_historyForwardAction, &QAction::triggered, [](){
+    addAction(back);
+    widgetForAction(back)->setObjectName("backButton");
```
mgautierfr, line 20: I add to do some git bisect to find why we add this (and not for the forwardButton). (https://github.com/kiwix/kiwix-desktop/commit/e9420e10c72884e77092a10694c2fbe43be70e43)

Can you add a small comment `// For css style` at the end of the line please ?

mgautierfr (changes requested):
Global code is good.

Can you merge the two last commits ?
The first one is globally wrong and I prefer to have somehow "autonomous" commit in the git history.
But please a line in the commit message telling that original commit author is @kelson42 and you've fixed it (to keep authorship)

src/topwidget.cpp
kelson42: done

PR #628, opened 2021-04-22
Fix pagination by setting pageLength properly

### LINKED ISSUE #617: [regression] Pagination problem with fulltext search.
Two questions please:

Is there a Kiwix To windows Shortcut key to return to the previous page?
When I do a full text search it shows me only the first 25 results. Do you know such a problem?
Thank you!

maneeshpm:
Fixes #617 

The `pageLength` parameter of `kiwix::SearchRenderer` is not set properly which causes it to take an arbitrarily large value, breaking pagination.

The solution is to use the same convention as kiwix-serve. Try to read the `pageLength` from `query` if found, else use a default value of 25. Use this `pageLength` to find `end`. Use `renderer.setPageLength()` method to set the `pageLength` properly.

The changes included in this PR are:
- Find `pageLength` from the query, fall back to a default value of 25
- Use this `pageLength` to find `end`
- Set the `pageLength` in the renderer

### DIFF
--- a/src/urlschemehandler.cpp
+++ b/src/urlschemehandler.cpp
@@ -101,10 +101,12 @@ UrlSchemeHandler::handleSearchRequest(QWebEngineUrlRequestJob* request)
     int temp = query.queryItemValue("start").toInt(&ok);
     if (ok)
       start = temp;
-    int end = 25;
-    temp = query.queryItemValue("end").toInt(&ok);
+    int pageLength = 25;
+    temp = query.queryItemValue("pageLength").toInt(&ok);
     if (ok)
-      end = temp;
+      pageLength = temp;
+
+    auto end = start + pageLength;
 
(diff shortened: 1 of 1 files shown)


maneeshpm:
@kelson42 @mgautierfr While working on this I found another issue. The `getEstimatedResultCount()` method breaks on both kiwix-desktop and kiwix-serve when changing the pages and takes a value of 200. This probably needs a fix in kiwix-lib, I will open a ticket there once am certain of the issue.

kelson42:
@maneeshpm Can you please open a bug ticket and track it down?

mgautierfr:
That is a consequence of https://github.com/kiwix/kiwix-lib/pull/399.
I haven't realize that we reuse the htmlRendered here and so we add to update the "server" part of kiwix-desktop.

mgautierfr (approved):

Issue #602, opened 2021-02-19
Add command-line --version flag


legoktm:
Debian CI has the option to run a command to verify the package runs properly after being installed, usually by running --help or --version. This would catch runtime errors like a missing dependency not being installed or the wrong library version being present. You can see what I've done for [kiwix-tools so far](https://salsa.debian.org/legoktm/kiwix-tools/-/blob/master/debian/tests/control).

So could kiwix-desktop also have a CLI `--version` option that just printed the version and quit?

Currently I get something like:
```
root@d290802461b3:/# QT_QPA_PLATFORM=minimal kiwix-desktop --version          
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
This plugin does not support createPlatformOpenGLContext!
[6276:6276:0219/223902.567966:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
root@d290802461b3:/# QT_QPA_PLATFORM=minimal kiwix-desktop --version --no-sandbox
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
This plugin does not support createPlatformOpenGLContext!
Compiled with Qt Version  5.15.1
Runtime Qt Version  5.15.2
This plugin does not support application fonts
This plugin does not support application fonts
This plugin does not support application fonts
This plugin does not support application fonts
This plugin does not support application fonts
kiwix-desktop: Unknown options: version, no-sandbox.
```

kelson42:
@HanYoung-uwu Maybe something you would be interested to have a look? The version is defined in the `.pro` file.

kelson42:
Adding a minimal help for `-h` and `--help` would be welcomed as well.

PR #631, opened 2021-05-02
add version and help commandline options

HanYoung-uwu:
As requested by #602

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -7,7 +7,6 @@
 #if QT_VERSION >= QT_VERSION_CHECK(5, 12, 0)
   #include <QWebEngineUrlScheme>
 #endif
-
 int main(int argc, char *argv[])
 {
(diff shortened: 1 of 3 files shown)


kelson42:
@HanYoung-uwu Thank you very much

src/main.cpp
```diff
@@ -7,7 +7,8 @@
 #if QT_VERSION >= QT_VERSION_CHECK(5, 12, 0)
   #include <QWebEngineUrlScheme>
 #endif
-
+#define _STR(...) # __VA_ARGS__
+#define STR(X) _STR(X)
```
mgautierfr, line 11: We don't need this, it is already in `src/kiwixapp.h`

mgautierfr (changes requested):
Just a small change. Everything else is good.

HanYoung-uwu:
force pushed

PR #634, opened 2021-05-03
FileDialog style and other efficiency fixes

HanYoung-uwu:
Setting windows style on non-windows platform makes FileDialog looks hideous. 
Default style
After set the Windows style

Note that "Windows" style is not what you usually want even if you like windows style. "QWindowsVistaStyle" is what people usually choose. 

Other changes include:
* Avoid extra copy on `KiwixApp::openZimFile`
* Remove `close()` on QFile because it close file in destructor, call it here you pay one more check on destructing (and extra line of code)

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -67,7 +67,6 @@ void KiwixApp::init()
     setApplicationName("Kiwix");
     setDesktopFileName("kiwix.desktop");
 
-    setStyle(QStyleFactory::create("Windows"));
     QFile styleFile(":/css/style.css");
     styleFile.open(QIODevice::ReadOnly);
     auto byteContent = styleFile.readAll();
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -70,7 +70,6 @@ void KiwixApp::init()
     QFile styleFile(":/css/style.css");
     styleFile.open(QIODevice::ReadOnly);
     auto byteContent = styleFile.readAll();
-    styleFile.close();
     QString style(byteContent);
     setStyleSheet(style);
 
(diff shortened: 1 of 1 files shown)


kelson42:
I see no visible change on Ubuntu, I guess this impacts only Windows users.

HanYoung-uwu:
> I see no visible change on Ubuntu, I guess this impacts only Windows users.

Well, this does impact me who is on Arch/KDE.

HanYoung-uwu:
> I see no visible change on Ubuntu, I guess this impacts only Windows users.

Is your filedialog like the second picture above?

kelson42:
@HanYoung-uwu This is how it looks like for me 
image

HanYoung-uwu:
> @HanYoung-uwu This is how it looks like for me
> image

So in your case, QFileDialog opens native filedialog using XDG, your native filedialog is gnome so Qt style doesn't affect you. But for system which native filedialog is written in Qt, ugly windows theme is used.

src/kiwixapp.cpp
```diff
@@ -159,14 +157,15 @@ void KiwixApp::openZimFile(const QString &zimfile)
...
-        _zimfile = QDir::toNativeSeparators(_zimfile);
+                    getMainWindow(),
+                    gt("open-zim"),
+                    QString(),
+                    "ZimFile (*.zim*)");
+
         if (_zimfile.isEmpty()) {
             return;
         }
+        _zimfile = QDir::toNativeSeparators(_zimfile);
```
mgautierfr: This change should be made in previous commit.

mgautierfr (changes requested):
Two small changes on the git commits.
Noting to say on the code itself.

---

> Remove close() on QFile because it close file in destructor, call it here you pay one more check on destructing (and extra line of code)

This comment could be put in the commit message itself
("Don't close the QFile explicitly as it will be close anyway in the destructor.")

HanYoung-uwu:
force pushed. Please re-review

mgautierfr (approved):

mgautierfr:
We are good. Thanks !

PR #643, opened 2021-05-13
port settings to Qt Widget

### LINKED ISSUE #640: Improve Settings panel Look & feel
Like for #412, we should improve the look & feel of the settings panel and it should not rely anymore on vue.js.

It should look like this (only consider the style)
image

and it looks like this:
image

HanYoung-uwu:
Fixes #640


kelson42:
@HanYoung-uwu Thank you very much for this PR. This is how it looks lie for me:
image

What shoudl be improved at least is:
* The font for the local server port, is somehow different (smaller?)
* The background color should be set
* Button should be like in the Kiwix Server button
* We should have a look to the two select boxes tosomehow have a "flat" style (but not sure what all has to be done there).

src/settingsmanager.cpp
```diff
@@ -13,12 +13,18 @@ SettingsManager::SettingsManager(QObject *parent)
...
+    if (!view || !m_settingsViewDisplayed)
+        view = new SettingsView();
+
+    view->init(m_kiwixServerPort, m_zoomFactor * 100, m_downloadDir);
+    connect(view, &QObject::destroyed, this, [=]() { m_settingsViewDisplayed = false;});
+    connect(view, &SettingsView::serverPortChanged, this, &SettingsManager::setKiwixServerPort);
+    connect(view, &SettingsView::downloadDirChanged, this, &SettingsManager::setDownloadDir);
+    connect(view, &SettingsView::zoomFactorChanged, this, &SettingsManager::setZoomFactor);
     m_settingsViewDisplayed = true;
     return view;
```
mgautierfr, line 29: This method is globally wrong.
- It is surprising to have a static variable here. It would be more "natural" to use a member of `SettingManager`
- If we have not null `view` and false `m_settingViewDisplayed`, it means that the view has been destroyed. So `view` is a dandling pointer. This is a "wrong" situation.
- If we call `getView` when the view is displayed, we will init and connect twice.
- `m_settingViewDisplayed` is never used, we could remove it (it was already the case before, but let's simplify this code in this PR)
- In fact, we don't need to keep a reference on `view` here as Qt will handle the lifetime of the widget for us once we will add it to the parent widget. (And we could rename this method to `createView`)

src/settingsmanager.cpp
```diff
@@ -62,6 +68,9 @@ void SettingsManager::setKiwixServerPort(int port)
 
 void SettingsManager::setZoomFactor(qreal zoomFactor)
 {
+    if (zoomFactor > 1)
+        zoomFactor /= 100;
+
```
mgautierfr, line 73: This is surprising. If the user enter a value of `1.5`, we will set 1.5%

mgautierfr (changes requested):
I'm not against the change from `settingmanagerview` to `settingsview` but it would have been easier for the review to rename the widget in one commit and change its implementation in another one.

(And I agree with @kelson42, the style should be adapted)

src/settingsmanager.cpp
HanYoung-uwu: That's because in the UI part, QSpinBox takes value from 0 to 100, so I need to divide it to the intended value. The `if` here is in case other parts use this function with correct value (0-1)

kelson42:
Now that we have merged the Kiwix Serve revamped dialog, I think we can finally attempt to finish/polish that one. I just have rebased the PR.

PR #646, opened 2021-05-20
Bind backward & forward mouse buttons to web history

### LINKED ISSUE #647: Mouse Thumb buttons support to navigate in the history
Based on a comment of #618, it would be good to implement mouse thumb history back/forward actions.

Goutte:
Spawned from #618 

Fixes #647 

It works with my own mouse. (Logitech MX518)

### DIFF
--- a/src/webview.cpp
+++ b/src/webview.cpp
@@ -112,6 +112,16 @@ bool WebView::eventFilter(QObject *src, QEvent *e)
             QDesktopServices::openUrl(m_linkHovered);
             return true;
         }
+        if (me->button() == Qt::BackButton)
+        {
+            back();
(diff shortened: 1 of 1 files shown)


kelson42:
@Goutte Thank you very much for the patch. Would be good if someone could test the patch. I have a logitech myself and will try to give a try, but would be better to test with a mouse from an other vendor.

mgautierfr:
Works also with my mouse (but it is also a logitech).
I think we can merge this patch as it is now. The change is pretty straightforward, at worst it will not work with other mouse, which is already the case now.

mgautierfr (approved):

Issue #540, opened 2020-11-23
Codebase: emiting a signal forcedly from outside is a bad practice


asashnov:
kiwix-desktop 'master' (830e19bb)

Found during working on #454

From src/zimview.cpp:
```c
emit mp_tabBar->currentTitleChanged(str);
emit mp_tabBar->currentZimIdChanged(zimId);
emit mp_tabBar->webActionEnabledChanged(QWebEnginePage::Back, mp_webView->isWebActionEnabled(QWebEnginePage::Back));
emit mp_tabBar->webActionEnabledChanged(QWebEnginePage::Forward, mp_webView->isWebActionEnabled(QWebEnginePage::Forward));
emit mp_library->bookmarksChanged();
```

According to Qt documentation, this is a bad practice.
By signals, only the object itself should notify that its internal state has changed:

https://doc.qt.io/qt-5/signalsandslots.html

> Signals
> Signals are emitted by an object when its internal state has changed in some way that might be interesting to the object's client or owner. Signals are public access functions and can be emitted from anywhere, but we recommend to only emit them from the class that defines the signal and its subclasses.

kelson42:
@asashnov Fell free to propose an alternative if this helps to fix #454 or trivial to do.

kelson42:
What could be an alternative approach?

asashnov:
Good approach is suggested by Qt documentation:

> Signals are emitted by an object when its internal state has changed

> we recommend to only emit them from the class that defines the signal

> It is even possible to connect a signal directly to another signal

> All classes that inherit from QObject or one of its subclasses (e.g., QWidget) can contain signals and slots. Signals are emitted by objects when they change their state in a way that may be interesting to other objects. This is all the object does to communicate. It does not know or care whether anything is receiving the signals it emits. This is true information encapsulation, and ensures that the object can be used as a software component.

kelson42:
@asashnov What is left here to do? I also want to emphasis that I suspsect that this massive regression https://github.com/kiwix/kiwix-desktop/issues/775 might be somehow related to our revamping work around the sidebar. Would you be able please to have a look and leave a comment?

kelson42:
@asashnov I rechallenge you on this: what is left to do?

asashnov:
kelson42@ Part3 would be illiminating ```emit mp_library->bookmarksChanged();```

kelson42:
@shaopengkin Can we move forward on this issue while PR currently in review waits for the reviewer?

[...]

PR #656, opened 2021-06-16
Partial fix of #540: do not emit others object signals.

asashnov:
This pull request fixes the codebase: signals should be emitted only by the object itself.

This is the "part 1" PR for #540:

* stop emitting TabBar::currentTitleChanged() from class ZimView
* stop emitting TabBar::currentZimIdChanged() from class ZimView - removed, not used anywhere

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -338,6 +338,18 @@ void TabBar::fullScreenRequested(QWebEngineFullScreenRequest request)
     }
 }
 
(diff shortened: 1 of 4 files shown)


mgautierfr:
The code seems good to me.
But a bug in libkiwix prevent the packages to be build (kiwix/libkiwix#564).
We could merge this PR right now, but as there is no urgence of this PR, I would prefer to wait we fix the issue in libkiwix and be sure we can compile correctly here.

Issue #669, opened 2021-07-31
URL tooltip is not displayed naymore on the bottom left corner


kelson42:
See 
![image](https://user-images.githubusercontent.com/1029718/127735666-dcc52ac1-33fd-4b26-ae00-bc34f4e8b6a9.png)

It is displayed a few pixels to high.... this looks like a regression as it was working

kelson42:
@asashnov Actually `QToolTip::showText(QPoint(0, 0), link)` does not even print exatly in the top left corner either... There is still a top and left margin... so I guess something fundamental is wrong here.

![image](https://user-images.githubusercontent.com/1029718/127737292-24791f48-07c7-40dc-af7e-80fa069ef698.png)

PR #670, opened 2021-07-31
Fixing qtool tip coordinates describe in issue #669

grivetti:
I fixed the issue #669 by changing the value of height in zimview.cpp at line 83

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -2,8 +2,9 @@
 
 #include "kiwixapp.h"
 #include "static_content.h"
+#include <kiwix/tools/networkTools.h>
+#include <kiwix/tools/otherTools.h>
 #include <kiwix/manager.h>
-#include <kiwix/tools.h>
 
 #include <QDebug>
 #include <QUrlQuery>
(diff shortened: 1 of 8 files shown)


kelson42:
@grivetti Thx, but please rebase on latest git master head. Then I would review.

kelson42:
@grivetti I had a look in detail and it looks like only https://github.com/kiwix/kiwix-desktop/pull/670/commits/7b2784595df3c4dca1cdd0878b0f60782f5f833c#diff-c4d37bdb4319da5738318ea9345602ceb48e654b597e4634b64ff3c0690d76e2R83 is really necessary (and I have to add 3 pixels to make it perfect... not clear why). But otherwise this is good.

This is the appropriate diff IMO:
 ```
$ git diff master
diff --git c/src/zimview.cpp w/src/zimview.cpp
index 93e79c2..d13748e 100644
--- c/src/zimview.cpp
+++ w/src/zimview.cpp
@@ -79,9 +79,7 @@ ZimView::ZimView(TabBar *tabBar, QWidget *parent)
                      * we can be sure the current tab is a web page
                      * (mp_tabBar->currentWebView() is not nullptr)
                      */
-                    auto height = mp_tabBar->currentWebView()->height() + 1;
-
-                    auto pos = mp_tabBar->mapToGlobal(QPoint(-3, height));
+                    auto pos = mp_tabBar->mapToGlobal(QPoint(-3, mp_tabBar->currentWebView()->height() - mp_tabBar->PdmHeightMM + 1));
                     QToolTip::showText(pos, link);
                 }
             });
```

I don't really understand why you need to do more... maybe you have an old version of kiwix-desktop code base, or maybe libkiwix?

kelson42 (changes requested):
Thx for the fix. Please apply changes or I can superseed this PR with a new one including your solution.

grivetti:
Thank you for the reply, during the debugging, I forgot the -3 value in width, now I made the pull request with the changes with the solution

kelson42:
@grivetti I have superseeded this PR by #672 because your PR still has a lot of code which is unrelated to the bug fix and IMO not even necessary at all. Thx again for the patch.

PR #675, opened 2021-08-06
Fixes #592 Explicitly specify black font color by default

### LINKED ISSUE #592: UI Text color should be specified
We have a custom/explicit background color, but the color of the text is the "default" them one. As a consequence, if someone as an "incompatible" UI theme then Kiwix Desktop will be unusable. We need to have an explicit color for the text color.

See more explanations at https://github.com/kiwix/kiwix-desktop/issues/320#issuecomment-775329595

asashnov:
Fixes #592 Explicitly specify black font color by default
(like it is in light color themes).

### DIFF
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -1,5 +1,6 @@
 QWidget {
     background-color: #EAECF0;
+    color: #000000;
 }
 
 


kelson42:
@asashnov Thx for the fix. I have assigned @mgautierfr as a reviewer, if you are not ready with the code, please put the PR in draft.

kelson42:
@mgautierfr Please review this PR

mgautierfr (approved):

PR #678, opened 2021-08-12
Proper human friendly URLs for kiwix-serve

### LINKED ISSUE #676: HTTP daemon does not have a proper human friendly name if not started with kiwix-serve
If started with Kiwix Desktop, URL look like:
http://192.168.0.101:8181/search?content=5149203d-1c60-a927-0095-278e34e6897a&pattern=estrutura+biologica

I suspect the problem comes from the `humanfriendly()` function which probably falls back.

This function should be fixed and in place of using the ZIM filename (which probably is not available here), it should build the human friendly name from the ZIM metadata (name, publisher, date, flavour).

MananJethwani:
this resolves #676

### DIFF
--- a/src/kiwixapp.h
+++ b/src/kiwixapp.h
@@ -16,6 +16,7 @@
 #include <QApplication>
 #include <QErrorMessage>
 #include <QTranslator>
+#include <kiwix/name_mapper.h>
 
 
 class KiwixApp : public QtSingleApplication
@@ -115,6 +116,7 @@ private:
     TabBar* mp_tabWidget;
     SideBarType m_currentSideType;
     QErrorMessage* mp_errorDialog;
+    kiwix::HumanReadableNameMapper m_nameMapper;
     kiwix::Server m_server;
     Translation m_translation;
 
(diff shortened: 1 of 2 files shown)


kelson42 (approved):
From a user perspective, seems to work fine

src/kiwixapp.h
```diff
@@ -16,6 +16,8 @@
 #include <QApplication>
 #include <QErrorMessage>
 #include <QTranslator>
+#include <kiwix/name_mapper.h>
+#include <kiwix/tools.h>
```
mgautierfr, line 20: Why do you need to include tools ?

kelson42: Seems useless indeed. Compile and run properly after my rebase and removal.

kelson42:
@MananJethwani @mgautierfr Can we merge this?

MananJethwani:
@kelson42 I believe we can merge this waiting for @mgautierfr.

mgautierfr (approved):

PR #680, opened 2021-08-23
Back/Forward buttons now have menus with history items

### LINKED ISSUE #3: History should be visible
Long click on "back" and "forward" buttons should show the history

asashnov:
Issue #3: Back/Forward buttons now have menus with history items

To see this change in action:
* open any zim file. See Back/Forward buttons: they are disabled, no submenu
* click on any link in opened document. Back button became active, long press on it will show the title of previously visited page (but don't select any item this time)
* click on any link again.  Long press on Back button. This time submenu will have two  items
* click on Back button. Forward button should be activated, long press on Forward button will show the title of document which had be just seen
* click on Back button again. Back button will be disabled. Forward button menu will contain two items.

### DIFF
--- a/src/topwidget.cpp
+++ b/src/topwidget.cpp
@@ -3,9 +3,11 @@
 #include "kconstants.h"
 #include "kiwixapp.h"
 #include "mainmenu.h"
+#include "tabbar.h"
 
 #include <QMouseEvent>
 #include <QAction>
+#include <QToolButton>
 
 TopWidget::TopWidget(QWidget *parent) :
     QToolBar(parent)
@@ -51,6 +53,10 @@ TopWidget::TopWidget(QWidget *parent) :
     addAction(menuAction);
     setContextMenuPolicy( Qt::PreventContextMenu );
 
+    // This signal emited more often than the history really updated
+    // but for now we have no better signal for it.
+    connect(KiwixApp::instance(), &KiwixApp::currentTitleChanged,
+            this, &TopWidget::updateBackForwardButtons);
 
 #if !SYSTEMTITLEBAR
     addAction(QIcon(":/icons/minimize.svg"), "minimize", parent, SLOT(showMinimized()));
@@ -101,4 +107,34 @@ void TopWidget::mouseMoveEvent(QMouseEvent *event) {
     auto delta = event->globalPos() - m_cursorPos;
     parentWidget()->move(delta);
     event->accept();
+}
+
+QToolButton* TopWidget::getBackButton() const
+{
+    auto app = KiwixApp::instance();
+    QAction *back = app->getAction(KiwixApp::HistoryBackAction);
+    return qobject_cast<QToolButton*>(widgetForAction(back));
+}
+
+QToolButton* TopWidget::getForwardButton() const
+{
+    auto app = KiwixApp::instance();
+    QAction *forward = app->getAction(KiwixApp::HistoryForwardAction);
+    return qobject_cast<QToolButton*>(widgetForAction(forward));
+}
+
+void TopWidget::updateBackForwardButtons()
+{
+    WebView *webview = KiwixApp::instance()->getTabWidget()->currentWebView();
+
+    if (webview) {
+        back_menu.reset(webview->getHistoryBackMenu());
+        forward_menu.reset(webview->getHistoryForwardMenu());
+    } else {
+        back_menu.reset();
+        forward_menu.reset();
+    }
+
+    getBackButton()->setMenu(back_menu.data());
+    getForwardButton()->setMenu(forward_menu.data());
 }
--- a/src/topwidget.h
+++ b/src/topwidget.h
@@ -4,6 +4,7 @@
 #include <QToolBar>
 #include <QLineEdit>
 #include <QWebEnginePage>
+#include <QScopedPointer>
 
 #include "searchbar.h"
 
@@ -15,19 +16,24 @@ public:
     virtual ~TopWidget();
 
     SearchBar &getSearchBar() { return m_searchEntry; };
+
+public slots:
+    void handleWebActionEnabledChanged(QWebEnginePage::WebAction action, bool enabled);
+    void updateBackForwardButtons();
+
 protected:
     void mousePressEvent(QMouseEvent *event);
     void mouseMoveEvent(QMouseEvent *event);
 
 private:
     SearchBar m_searchEntry;
-    QAction* mp_historyBackAction;
-    QAction* mp_historyForwardAction;
     QPoint m_cursorPos;
     ulong m_timestamp;
+    QScopedPointer<QMenu, QScopedPointerDeleteLater> back_menu;
+    QScopedPointer<QMenu, QScopedPointerDeleteLater> forward_menu;
 
-public slots:
-    void handleWebActionEnabledChanged(QWebEnginePage::WebAction action, bool enabled);
+    QToolButton* getBackButton() const;
+    QToolButton* getForwardButton() const;
 };
 
 #endif // TOPWIDGET_H
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -50,10 +50,6 @@ SearchBar > QPushButton {
     width: 32px;
 }
 
-/* ----------------------------------------
-   Menu
-*/
-
 TopWidget QToolButton:pressed,
 TopWidget QToolButton::hover {
     border: 1px solid #3366CC;
@@ -70,17 +66,25 @@ TopWidget QToolButton::menu-arrow {
     image: none;
 }
 
-TopWidget QToolButton#leftHistoryButton {
-    margin-left: 4px;
+TopWidget QToolButton::menu-indicator {
+    width: 0px; /* it can be hidden at all by setting 0px */
+    height: 0px;
+    subcontrol-origin: padding;
 }
 
-TopWidget QToolButton#rightHistoryButton {
-    margin-right: 4px;
+TopWidget QToolButton#backButton {
+    margin-left: 6px;  /* see also: void WebViewBackMenu::showEvent(QShowEvent *) { geo.setX(geo.x() + 6); } */
 }
 
 TopWidget QToolButton#fullScreenButton {
     margin-right: 6px;
 }
+
+
+
+/* ----------------------------------------
+   Menu
+*/
 
 QMenu {
     border: none;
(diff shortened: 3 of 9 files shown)


kelson42:
@asashnov Thx, how do you test this exactly?

kelson42 (changes requested):
Nice. It works fine to me. Here are my polishing requests:
* Put the lists in reverted order like in a browser (the most recent on the top)
* Contextual dialog should be displayed exactly below the bottom left (with a vertical space of a few pixels beetween the bottom of the button and the top of the contextual menu)
* List Item left margin is inappropriate (too big).

kelson42:
@asashnov Any news on this PR, we need to move forward!

src/topwidget.cpp
```diff
@@ -94,3 +98,40 @@ void TopWidget::mouseMoveEvent(QMouseEvent *event) {
...
+    QAction *forward = app->getAction(KiwixApp::HistoryForwardAction);
+    return qobject_cast<QToolButton*>(widgetForAction(forward));
+}
+
+void TopWidget::updateBackForwardButtons()
+{
+    if (back_menu) {
+        back_menu->deleteLater();
+        back_menu = Q_NULLPTR;
+    }
```
mgautierfr, line 121: Do we really need to delete it ?
Who is the owner of the menu ? https://doc.qt.io/qt-5/qtoolbar.html#widgetForAction is not explicit about if the returned pointer is owned by the action or not.

src/topwidget.h
```diff
@@ -15,19 +15,24 @@ class TopWidget : public QToolBar
...
     void mouseMoveEvent(QMouseEvent *event);
 
 private:
     SearchBar m_searchEntry;
-    QAction* mp_historyBackAction;
-    QAction* mp_historyForwardAction;
     QPoint m_cursorPos;
     ulong m_timestamp;
+    QMenu *back_menu = Q_NULLPTR;
+    QMenu *forward_menu = Q_NULLPTR;
```
mgautierfr, line 32: We should probably use a QScopedPointer (https://doc.qt.io/qt-5/qscopedpointer.html) here.
It would avoid us the `if (back_menu) { [delete] };` part (if we need it) and also automatically "delete" the pointer when we destruct the topWidget (and you forgot this part)

mgautierfr (changes requested):

kelson42:
@asashnov If this is ready for a new review, please request one.

kelson42:
I have rebased the branch on master

src/topwidget.cpp
asashnov: From doc on `void QToolButton::setMenu(QMenu *menu)`:
> Ownership of the menu is not transferred to the tool button.

QToolBar creates and owns QToolButton ("Back"), but this "Back" button doesn't own the history menu.

src/topwidget.h
asashnov: Yes, good idea to illiminate this error-prone approach.
I use deleteLater() just for a case if page URL changed during history menu somehow is shown.  Hard to reproduce, but I think it's possible situation.

src/topwidget.cpp
```diff
@@ -94,3 +98,33 @@ void TopWidget::mouseMoveEvent(QMouseEvent *event) {
...
+
+    if (webview) {
+        back_menu.reset(webview->getHistoryBackMenu());
+        forward_menu.reset(webview->getHistoryForwardMenu());
+    } else {
+        back_menu.reset();
+        forward_menu.reset();
+    }
+
+    getBackButton()->setMenu(back_menu.get());
```
mgautierfr, line 128: The get method has been introduce in qt 5.11 and bionic use qt 5.9 (https://doc.qt.io/qt-5/qscopedpointer.html#get)
This is why the CI doesn't compile on bionic.

mgautierfr:
Capture d’écran du 2021-09-08 11-23-35

The two back/forward buttons now have a small arrow at bottom because they have menu attached to them.
It breaks a bit the homogenety of the interface. Is it possible to remove them ?
(In the same time, users may not know that they can long click on the button to have the menu if we remove them).

asashnov:
@kelson42 sorry for the delay, I was on my 1-week trip to beautiful Saint-Petersburg while summer is not over.

All the points are fixed except the last one:

> Switching from a long click on one button (backward for example) to a long click to click on the other button (forward) fails. The menu is not displayed.

In my experiments, Firefox and Chrome behave the same:
When long press Back, and move outside of the button or history menu, menu is not hidden on left mouse button release. So need to click somewhere else first. Only Safari on Mac is different: it hides history menu automatically.

I think we may leave it "as is" because it's the expected user experience on many platforms/programs.

src/topwidget.cpp
asashnov: Fixed, used data()

asashnov:
@mgautierfr   The size of menu indicator decreased, now looks better. Can be hidden by setting its 'width: 0px;' in css.

kelson42:
@asashnov Thx for the fix and explanation, but I believe the following points have not been treated:

> * Contextual dialog should be displayed exactly below the bottom left (with a vertical space of a few pixels beetween the bottom of the button and the top of the contextual menu)
> 
> * List Item left margin is inappropriate (too big).

kelson42 (changes requested):
@asashnov Any news on this front?

mgautierfr:
The small arrow is now cut 
Capture d’écran du 2021-09-21 11-39-13

And it seems the size of the button change a bit depending on what your click on. It seems that selected buttons are a bit bigger.
Maybe not related to this change however.

kelson42:
@asashnov What is the status here? The PR is pretty old now and only a few details need to be fixed!

asashnov:
Menu position, and paddings were fixed.

kelson42 (approved):
Great, from a user perspective LGTM. Thx!

src/topwidget.cpp
```diff
@@ -43,6 +45,8 @@ TopWidget::TopWidget(QWidget *parent) :
     addAction(menuAction);
     setContextMenuPolicy( Qt::PreventContextMenu );
 
+    connect(KiwixApp::instance(), &KiwixApp::currentTitleChanged,
+            this, &TopWidget::updateBackForwardButtons);
```
mgautierfr, line 49: I wonder if we really want to connect to the `currentTitleChanged` ?
The title will change if we change page (and we need to update buttons) but it will also change if the html tell it so (in instance, if I remember correctly, the title is changed twice when we change page. One first time with the "path" of the page and a second time with the title after the html is parsed).

In the same time, `updateBackForwardButtons` is triggered by this event but it doesn't rely on it. It ask for the real history to update the button. So it may be not so important.

mgautierfr (commented):
I have the same problem than my previous comment, small arrows are cut and the buttons change size.

My comment on the code is a simple comment, I will not block for this.

kelson42:
@asashnov I have rebased this branch but what is the exact status? I worry that on @mgautierfr' computer the button arrow is partly hidden. If you don't find a solution to this problem, maybe just not having a small arrow could be the best thing to do?

kelson42:
@asashnov This PR has not moved since more than a month, we really need to do something. Either you have in mind a small hack to secure the newly added arrows display better on Fedora, or then I would propose to just get rid of them. Whatever the solution, we need to do something and merge this PR soon.

asashnov:
Tried to fix Back / Forward buttons menu indicator sizes but having troubles in trying to test on Fedora.
In Qt sources, it seems both weigh and hight may matter, so I defined the second. And, for a case, the position of menu indicator.

Anyway, my suggestion would be to merge the current functionality as it works as expected, and open a new issue on this "menu indicators cut on Fedora".

kelson42:
@asashnov Why not just removing this small arrows?! They are not requested. If we are not sure they are displayed properly, then why keeping them?

asashnov:
I managed to build the program on Fedora 34 and run the app image. Arrows displayed correctly with my latest changes in CSS:

Screenshot from 2021-11-20 21-38-58

But it can be hidden by setting 'menu-indicator' height and width to 0px.

kelson42:
@mgautierfr Doesnthis latest version works better?

mgautierfr:
It doesn't.
What Fedora version are you using ? Which Qt version ?
It seems you are usin a AppImage build. Does it works also with the "native" build ?

I'm on fedora 34 with Qt 5.15.2-22
I've tested both with the "native build" and a (locally build) AppImage.

kelson42:
@asashnov I start to be annoyed by the situation. It makes 2 months that this PR is blocked and consumes time for both @mgautierfr and you. This for a part of the feature which is not required. I don't want to continue that path as we have limited resources and other priorities. Therefore, please remove these arrows so we can merge a feature which AFAIK works perfectly for everybody. If you really want to sort-out what goes wrong, open a new PR with this buggy arrows, but please don't consume much more time on them.

asashnov:
> It doesn't. What Fedora version are you using ? Which Qt version ? It seems you are usin a AppImage build. Does it works also with the "native" build ?
> I'm on fedora 34 with Qt 5.15.2-22 I've tested both with the "native build" and a (locally build) AppImage.

I tested on Fedora 34x64 VirtualBox image from osboxes.org, with latest updates.
The system was clean: extract .vdi from archive, install latest updates, and use kiwix-build exactly by the commands from my PR: https://github.com/kiwix/kiwix-build/pull/506
No other customization or settings done, I didn't do anything in that system- only deployed it for the test.

Qt from Fedora repo (not from Qt official site):
qt5-qtbase-common-5.15.2-27.fc34.noarch
qt5-qtbase-5.15.2-27.fc34.x86_64
qt5-qtbase-gui-5.15.2-27.fc34.x86_64
qt5-qtwayland-5.15.2-14.fc34.x86_64

I don't know how to run "native" build (this is also a subject to add in README.md of kiwix-build), so I built AppImage.

asashnov:
Done, the menu indicators are hidden now. Commits are rebased, squashed, and ready to be merged now as one commit.

kelson42:
@asashnov Thank you!
@mgautierfr Are we good to merge?

kelson42 (approved):

mgautierfr:
> I don't know how to run "native" build (this is also a subject to add in README.md of kiwix-build), so I built AppImage.

Just run kiwix-desktop binary in `BUILD_native_dyn/INSTALL/bin`  (you may have to set your env variable LD_LIBRARY_PATH before to allow the loader to find the .so)
```bash
$ export LD_LIBRARY_PATH = [...]/BUILD_native_dyn/INSTALL/lib64
$ [...]/BUILD_native_dyn/bin/kiwix-desktop
```

PR #682, opened 2021-08-26
Add tooltips to show keyboard shortcuts 

### LINKED ISSUE #665: Additionally show the keyboard shortcuts for buttons in their tooltip
Most applications show a description of the action of a button, along with the assigned keyboard shortcut (if any)  in a tooltip, when hovering. As a first-time user, I was expecting this, especially since the F11-Fullscreen shortcut "did not work". (See separate issue https://github.com/kiwix/kiwix-desktop/issues/304)

Here's how it looks like currently:
hovering-tooltip
_The tooltip on hover is displayed without keyboard shortcut information_

Hovever, in the menu, however, the shortcut (CRTL+T in this case), is displayed properly.

Expected behaviour: The tooltip should display the keyboard shortcut along the text.
hovering-tooltip-with-shortcut
_The proposed tooltip with keyboard shortcut information_

Note: I experience this on a fresh, updated Ubuntu 20.04, using Version 2.0.5

juuz0:
Fixes #665 

Tool tips for the following have been added -
New Tab (Ctrl + T)
Close Tab (Ctrl + F4)
Reading List (Ctrl + Shift + 2)
Open File (Ctrl + O)
Toggle fullscreen (F11)

image

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -281,57 +281,52 @@ bool KiwixApp::isCurrentArticleBookmarked()
     return false;
 }
 
-#define CREATE_ACTION_ICON(ID, ICON, TEXT) \
-    mpa_actions[ID] = new QAction(QIcon(":/icons/" ICON ".svg"), TEXT)
 #define CREATE_ACTION(ID, TEXT) \
     mpa_actions[ID] = new QAction(TEXT)
-#define SET_SHORTCUT(ID, SHORTCUT) mpa_actions[ID]->setShortcut(SHORTCUT)
+#define SET_SHORTCUT(ID, TEXT, SHORTCUT) \
+    mpa_actions[ID]->setShortcut(SHORTCUT); \
+    mpa_actions[ID]->setToolTip(TEXT + " (" + QKeySequence(SHORTCUT).toString() + ")" )
+#define CREATE_ACTION_SHORTCUT(ID, TEXT, SHORTCUT) \
+    CREATE_ACTION(ID, TEXT); \
+    SET_SHORTCUT(ID, TEXT, SHORTCUT)
+#define CREATE_ACTION_ICON_SHORTCUT(ID, ICON, TEXT, SHORTCUT) \
+    mpa_actions[ID] = new QAction(QIcon(":/icons/" ICON ".svg"), TEXT); \
+    SET_SHORTCUT(ID, TEXT, SHORTCUT)
 #define HIDE_ACTION(ID) mpa_actions[ID]->setVisible(false)
 #define DISABLE_ACTION(ID) mpa_actions[ID]->setDisabled(true)
 
 void KiwixApp::createAction()
 {
-    CREATE_ACTION_ICON(KiwixServeAction, "share", gt("local-kiwix-server"));
-    SET_SHORTCUT(KiwixServeAction, QKeySequence(Qt::CTRL+Qt::Key_I));
+    CREATE_ACTION_ICON_SHORTCUT(KiwixServeAction, "share", gt("local-kiwix-server"), QKeySequence(Qt::CTRL+Qt::Key_I));
 
-    CREATE_ACTION_ICON(RandomArticleAction, "random", gt("random-article"));
-    SET_SHORTCUT(RandomArticleAction, QKeySequence(Qt::CTRL+Qt::Key_R));
+    CREATE_ACTION_ICON_SHORTCUT(RandomArticleAction, "random", gt("random-article"), QKeySequence(Qt::CTRL+Qt::Key_R));
     connect(mpa_actions[RandomArticleAction], &QAction::triggered,
             this, [=]() { this->openRandomUrl(false); });
 
-    CREATE_ACTION(OpenHomePageAction, gt("home-page"));
-    SET_SHORTCUT(OpenHomePageAction, QKeySequence(Qt::ALT + Qt::Key_Home));
+    CREATE_ACTION_SHORTCUT(OpenHomePageAction, gt("home-page"), QKeySequence(Qt::ALT + Qt::Key_Home));
 
-    CREATE_ACTION_ICON(HistoryBackAction, "back", gt("back"));
-    SET_SHORTCUT(HistoryBackAction, QKeySequence(Qt::ALT + Qt::Key_Left));
+    CREATE_ACTION_ICON_SHORTCUT(HistoryBackAction, "back", gt("back"), QKeySequence(Qt::ALT + Qt::Key_Left));
     DISABLE_ACTION(HistoryBackAction);
 
-    CREATE_ACTION_ICON(HistoryForwardAction, "forward", gt("forward"));
-    SET_SHORTCUT(HistoryForwardAction, QKeySequence(Qt::ALT + Qt::Key_Right));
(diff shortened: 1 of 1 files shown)


kelson42:
@juuz0 Please don't merge master in your branch. Instead make a rebase.

juuz0:
Would this work?

kelson42:
@juuz0 Yes, we are a bit picky about how the feature branches revision log should look like… all this to keep a clean revision log once the PR will be merged.

src/kiwixapp.cpp
```diff
@@ -316,10 +316,12 @@ void KiwixApp::createAction()
 
     CREATE_ACTION(NewTabAction, gt("new-tab"));
     SET_SHORTCUT(NewTabAction, QKeySequence::AddTab);
+    mpa_actions[NewTabAction]->setToolTip(gt("new-tab").append(" (Ctrl+T)"));
```
mgautierfr, line 319: The sequence for AddTab may not be `Ctrl+T` depending of the platform.
We must not hardcode this.
The best way is to get the string from https://doc.qt.io/qt-5/qkeysequence.html#toString
This way we will always display the right shortcut to the user.

I also wonder if the property `shortcutVisibleInContextMenu` of the action itself (https://doc.qt.io/qt-5/qaction.html#shortcutVisibleInContextMenu-prop) could help us also.

mgautierfr (changes requested):
As said in the code comment, we must not hardcode the shortcut displayed to the user.

juuz0:
On it, should I push the changes in this branch (issue665) only?

kelson42:
@juuz0 yes please, where else?

kelson42:
@juuz0 thank you for the fixes, is this ready now for a new review?

juuz0:
@kelson42 Yes!

mgautierfr (changes requested):
The code is functionally good but difficult to read and redundant with other configuration of the action (we are writing twice the sequence and the label of the action).

It would be better to addapt the existing macros and have a new macro `CREATE_ACTION_SHORTCUT`/`CREATE_ACTION_ICON_SHORTCUT` which create the action, set the shortcut and the tooltip for us in "one" shoot.

kelson42:
@juuz0 Would you be able to follow @mgautierfr advices and make the necessary improvements?

juuz0:
@kelson42 @mgautierfr  I think this is what was asked :)

src/kiwixapp.cpp
```diff
@@ -382,38 +368,31 @@ void KiwixApp::createAction()
...
-    SET_SHORTCUT(ZoomResetAction, QKeySequence(Qt::CTRL+Qt::Key_0));
+    CREATE_ACTION_SHORTCUT(ZoomResetAction, gt("zoom-reset"), QKeySequence(Qt::CTRL+Qt::Key_0));
 
-    CREATE_ACTION(HelpAction, gt("help"));
-    SET_SHORTCUT(HelpAction, QKeySequence::HelpContents);
+    CREATE_ACTION_SHORTCUT(HelpAction, gt("help"), QKeySequence::HelpContents);
     HIDE_ACTION(HelpAction);
 
-    CREATE_ACTION(FeedbackAction, gt("feedback"));
+    CREATE_ACTION_SHORTCUT(FeedbackAction, gt("feedback"), 0);
```
mgautierfr, line 380: Could we keep a `CREATE_ACTION` for action without shortcut ?

src/kiwixapp.cpp
```diff
@@ -281,84 +281,72 @@ bool KiwixApp::isCurrentArticleBookmarked()
...
-#define CREATE_ACTION_ICON(ID, ICON, TEXT) \
-    mpa_actions[ID] = new QAction(QIcon(":/icons/" ICON ".svg"), TEXT)
-#define CREATE_ACTION(ID, TEXT) \
-    mpa_actions[ID] = new QAction(TEXT)
-#define SET_SHORTCUT(ID, SHORTCUT) mpa_actions[ID]->setShortcut(SHORTCUT)
+#define CREATE_ACTION_ICON_SHORTCUT(ID, ICON, TEXT, SHORTCUT) \
+    mpa_actions[ID] = new QAction(QIcon(":/icons/" ICON ".svg"), TEXT); \
+    mpa_actions[ID]->setShortcut(SHORTCUT); \
+    mpa_actions[ID]->setToolTip(TEXT + " (" +  QKeySequence(SHORTCUT).toString() + ")" ); \
+    mpa_actions[ID]->setIconVisibleInMenu(false)
```
mgautierfr, line 288: We probably don't want to set all icons invisible in menu.
We were setting the icon invisible only for closeTabAction.
Now we are setting this for all actions.

mgautierfr (changes requested):
We are almost good. This is far better.
There is still two small changes.

For the macros, we can keep the following ones :
```cpp
#define CREATE_ACTION(ID, TEXT) \
    mpa_actions[ID] = new QAction(TEXT)
#define SET_SHORTCUT(ID, TEXT, SHORTCUT) \
    mpa_actions[ID]->setShortcut(SHORTCUT); \
    mpa_actions[ID]->setToolTip(TEXT + " (" + KeySequence(SHORTCUT).toString() + ")" )
#define CREATE_ACTION_SHORTCUT(ID, TEXT, SHORTCUT) \
    CREATE_ACTION(ID, TEXT); \
	SET_SHORTCUT(ID, TEXT, SHORTCUT)
#define CREATE_ACTION_ICON_SHORTCUT(ID, ICON, TEXT, SHORTCUT) \
    mpa_actions[ID] = new QAction(QIcon(":/icons/" ICON ".svg"), TEXT); \
    SET_SHORTCUT(ID, TEXT, SHORTCUT)
```

src/kiwixapp.cpp
juuz0: Yeah, I kind of assumed all should be invisible in menu, will fix this thanks!

src/kiwixapp.cpp
```diff
@@ -281,57 +281,52 @@ bool KiwixApp::isCurrentArticleBookmarked()
...
-    mpa_actions[ID] = new QAction(QIcon(":/icons/" ICON ".svg"), TEXT)
 #define CREATE_ACTION(ID, TEXT) \
     mpa_actions[ID] = new QAction(TEXT)
-#define SET_SHORTCUT(ID, SHORTCUT) mpa_actions[ID]->setShortcut(SHORTCUT)
+#define SET_SHORTCUT(ID, TEXT, SHORTCUT) \
+    mpa_actions[ID]->setShortcut(SHORTCUT); \
+    mpa_actions[ID]->setToolTip(TEXT + " (" + QKeySequence(SHORTCUT).toString() + ")" )
+#define CREATE_ACTION_SHORTCUT(ID, TEXT, SHORTCUT) \
+    CREATE_ACTION(ID, TEXT); \
+	SET_SHORTCUT(ID, TEXT, SHORTCUT)
```
mgautierfr, line 291: Tab instead of spaces.

mgautierfr (changes requested):
On very small tiny change and we are good.

You can squash all commits into one for this PR. The change is pretty self contained.

juuz0:
I am not sure on how that tab problem happened but hopefully this is fine now.

kelson42:
@juuz0 Thx for your work, merging now.

PR #684, opened 2021-08-30
Added attribute to show tooltips even when the window is unfocused.

### LINKED ISSUE #666: Tooltips inconsistency: not shown, when application is not focused
Most applications (e.g. Firefox, Thunderbird, LibreOffice on Ubuntu OS at least) show their tooltips when hovering the mouse over a button, even when they are not have focus (are not receiving actual keyboard input at the moment).

Kiwix, in this case, is highlighting the action, if available (lighting up in blue style), but the respective tooltip is not shown. Clicking the button however, immediately transfers the focus to the app and excecutes the action, as expected.

Here's how it looks like, when the mouse hovers over an actionable item, when the app does not have focus:
notfocusednotooltip

The expected behaviour would be that the tooltip is shown, as if the application had focus.

Note: 
- This might be an issue releated to some default behaviour of Qt.
- I experience this on a fresh, updated Ubuntu 20.04, using Version 2.0.5

juuz0:
Fixes #666 

The attribute of QWidget : Qt::WA_AlwaysShowToolTips  is added.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -17,6 +17,7 @@ MainWindow::MainWindow(QWidget *parent) :
     mp_about(new About(this)),
     mp_localKiwixServer(new LocalKiwixServer(this))
 {
+    QWidget::setAttribute(Qt::WA_AlwaysShowToolTips);
     mp_ui->setupUi(this);
     mp_ui->tabBar->setExpanding(false);
     mp_ui->tabBar->setStackedWidget(mp_ui->mainView);


kelson42:
@juuz0 nice job :)

juuz0:
Thanks :) 
Why is this failing though?

kelson42:
@juuz0 This is a new CI configuration attempt, which is still incomplete. Nothin wrong with your code I guess.

mgautierfr (approved):

mgautierfr:
Can you rebase on last master version before we merge please ?

kelson42:
Considering that there is only one commit, just squashing + merging.

PR #689, opened 2021-09-05
Added condition to zoom in/out only for the current tab

### LINKED ISSUE #654: Zoom in/out/reset shoold work forthe current tab only
Currently (57162a79) Zoom in/out/reset action applied to all the opened tabs.

## Actual result

Pressing Ctrl++  Ctrl+-  (or zoom in/out with mouse wheel) is applied to every opened tab at the moment.

## Expected result

Pressing Ctrl++  Ctrl+-  (or zoom in/out with mouse wheel) should be applied to the current (active) tab only, as it is in all major web browsers.

## Additional details

In code that happens because Zoom in/out/reset is connected to every tab object created, regardless of the current tab:

https://github.com/kiwix/kiwix-desktop/blob/57162a79ed21b46a65622093aa12cf5d99738cff/src/zimview.cpp#L21

Suggested solution:

Handle Zoom in/out/reset like we handle 'Print' command (only the currently active tab is printed):

https://github.com/kiwix/kiwix-desktop/blob/57162a79ed21b46a65622093aa12cf5d99738cff/src/kiwixapp.cpp#L305

https://github.com/kiwix/kiwix-desktop/blob/57162a79ed21b46a65622093aa12cf5d99738cff/src/kiwixapp.cpp#L184

juuz0:
Fixes #654

### DIFF
--- a/src/zimview.cpp
+++ b/src/zimview.cpp
@@ -42,6 +42,8 @@ ZimView::ZimView(TabBar *tabBar, QWidget *parent)
             });
     connect(app->getAction(KiwixApp::ZoomResetAction), &QAction::triggered,
             this, [=]() {
+                if (mp_tabBar->currentZimView() != this)
+                    return;
                 auto settingsManager = KiwixApp::instance()->getSettingsManager();
                 mp_webView->setZoomFactor(settingsManager->getZoomFactor());
                 auto key = mp_webView->zimId() + "/zoomFactor";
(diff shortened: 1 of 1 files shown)


kelson42:
Rebased on master

mgautierfr (approved):

mgautierfr:
Thanks for this change.

PR #712, opened 2021-11-01
Detect and add new zim files from Download Dir into local library

### LINKED ISSUE #644: On Startup detect new zim files added to download directory and ask to add them to local files library (all at once)
This should be a bandwidth saver (both server and client side) for people who download once and load on multiple devices. Right now you would have to add files one at a time

juuz0:
Fixes #644

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -34,6 +34,7 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
     }
     qInfo() << "Compiled with Qt Version " << QT_VERSION_STR;
     qInfo() << "Runtime Qt Version " << qVersion();
+    m_library.syncNewBooksInLibrary(kiwix::getDataDirectory());
     m_qtTranslator.load(QLocale(), "qt", "_",
                         QLibraryInfo::location(QLibraryInfo::TranslationsPath));
     installTranslator(&m_qtTranslator);
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -14,6 +14,7 @@
 #include <QPrintDialog>
 #include <thread>
 #include <QMessageBox>
+#include <QtConcurrent/QtConcurrentRun>
 
 KiwixApp::KiwixApp(int& argc, char *argv[])
     : QtSingleApplication("kiwix-desktop", argc, argv),
@@ -34,7 +35,10 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
     }
     qInfo() << "Compiled with Qt Version " << QT_VERSION_STR;
     qInfo() << "Runtime Qt Version " << qVersion();
-    m_library.syncNewBooksInLibrary(kiwix::getDataDirectory());
+
+    QFuture<void> future = QtConcurrent::run([=](){
+        m_library.syncNewBooksInLibrary(kiwix::getDataDirectory());
+    });
     m_qtTranslator.load(QLocale(), "qt", "_",
                         QLibraryInfo::location(QLibraryInfo::TranslationsPath));
     installTranslator(&m_qtTranslator);
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -51,6 +51,32 @@ QString Library::openBookFromPath(const QString &zimPath)
     return QString::fromStdString(id);
 }
 
+QStringList Library::getBooksFromDir(std::string dir)
+{
+    auto downloadDir = QDir(QString::fromStdString(dir));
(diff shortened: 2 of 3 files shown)


kelson42 (changes requested):
For me, unfortunately does not work. I have downloaded a ZIM file manuaylly in `/home/kelson/.local/share/kiwix` which is my profile directory but it does not appear in the Kiwix-Desktop after I restarted it... but I can add the file manuallly.

juuz0:
@kelson42 
Which is that zim file?
I tried with a couple of more files and both work for me.

(Maybe its some different language file and will only show up after choosing All files in the left-side panel?)

kelson42:
It is not a filter problem.

kelson42:
@mgautierfr Does it work for you?

mgautierfr:
> Does it work for you?

It works. But it takes time. The function has to parse all the directory and open all new zim files before you see changes in the UI.

src/library.cpp
```diff
@@ -51,6 +51,32 @@ QString Library::openBookFromPath(const QString &zimPath)
...
+    for (auto bookPath : allBooks) {
+        auto fullBookPath = QDir::cleanPath(QString::fromStdString(kiwix::getDataDirectory()) + '/' + bookPath);
+        try {
+            m_library.getBookByPath(fullBookPath.toStdString());
+        }  catch (std::out_of_range& e) {
+            kiwix::Manager manager(&m_library);
+            if(!manager.addBookFromPath(fullBookPath.toStdString())) {
+                qWarning()<<"Couldn't add "<<fullBookPath<<" in the library.";
+            }
+            save();
```
mgautierfr, line 73: It would be better to create only one Manager, and save the library at the end of the process.

src/library.cpp
```diff
@@ -51,6 +51,32 @@ QString Library::openBookFromPath(const QString &zimPath)
...
+
+void Library::syncNewBooksInLibrary(std::string dir)
+{
+    QStringList allBooks = Library::getBooksFromDir(dir);
+    for (auto bookPath : allBooks) {
+        auto fullBookPath = QDir::cleanPath(QString::fromStdString(kiwix::getDataDirectory()) + '/' + bookPath);
+        try {
+            m_library.getBookByPath(fullBookPath.toStdString());
+        }  catch (std::out_of_range& e) {
+            kiwix::Manager manager(&m_library);
```
mgautierfr, line 69: It is better to create a `LibraryManipulator` and create the manager on this manipulator. (As it is made in the `Library` creator).
This way, the manipulator emit the `booksChanged` signal at every new book added and the UI is update as soon as possible.

src/kiwixapp.cpp
```diff
@@ -34,6 +35,10 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
     }
     qInfo() << "Compiled with Qt Version " << QT_VERSION_STR;
     qInfo() << "Runtime Qt Version " << qVersion();
+
+    QFuture<void> future = QtConcurrent::run([=](){
+        m_library.syncNewBooksInLibrary(kiwix::getDataDirectory());
+    });
```
mgautierfr, line 41: We have to be careful with multithreading. The library is not thread safe, so you must protect it from multiple call with a mutex.

mgautierfr (changes requested):
This PR works but there are still few things to change.

@veloman-yunkan is working on live updating the library with issue kiwix/kiwix-tools#243
It would be better than you see with him what is the best way to do this (and not implement twice the same thing).
It may even worth it to delay this PR until the work on kiwix/kiwix-tools#243 is finished its work on the libkiwix side.

@veloman-yunkan, what do you think about this ?

kelson42:
Tested it again... an now it works... I probably made an error the first time I tested.

kelson42:
@juuz0 Thank you for your patience on this. We have discussed this PR with @mgautierfr. What is sure is that:
* This topic is at the crossroad of meany topics
* Even if we tend to agree on the whole concept this is still not printed in the stone
* Your code depends on the capacity of the libkiwix to easily refresh the library and @veloman-yunkan is currently working on this with https://github.com/kiwix/libkiwix/pull/636.

For all these reasons, we would like to wait the code update code in libkiwix is ready. That wait it would avoid you have to reimplement part of it by yourself. Then we could focus on the pure feature aspect of this PR.

juuz0:
@kelson42 Yes thanks for the update! Library reloading will be nice when done :)

juuz0:
Closing in favour of #743

PR #714, opened 2021-11-01
Propagation of library updates to the embedded kiwix server

veloman-yunkan:
Fixes kiwix/libkiwix#629

The problem is solved by fully recreating the name mapper whenever the library changes. An alternative solution was to add new methods `addBook()` and `removeBook()` to `HumanReadableNameMapper` (or to a new subclass of it).

The primary goal was to constrain the fix to a single repo (so that the reported issue is promptly resolved by one PR). However, this or the alternative solution will need to be migrated to libkiwix so that it can be reused for kiwix/kiwix-tools#243.


mgautierfr (approved):
The solution should be moved in libkiwix soon but I agree we can go with this "local" solution and move it to libkiwix with the work on kiwix/kiwix-tools#243

kelson42:
@veloman-yunkan @mgautierfr I just tested, and the symptom described in my ticket kiwix/libkiwix#629 ist still there.

veloman-yunkan:
> @veloman-yunkan @mgautierfr I just tested, and the symptom described in my ticket [kiwix/libkiwix#629](https://github.com/kiwix/libkiwix/issues/629) ist still there.

@kelson42 Something is wrong with your testing. It works for me. Did you build the code yourself? Are you sure that the fix is included in your build?

PR #716, opened 2021-11-03
Added functionality to open new tabs right after current tab.

### LINKED ISSUE #636: Open new tab just right after the current one
When you open (right clic) a link in a new tab, would be better to have it just after the current active one.

First reported in #635

juuz0:
Fixes #636

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -26,7 +26,7 @@ TabBar::TabBar(QWidget *parent) :
 
     connect(app->getAction(KiwixApp::NewTabAction), &QAction::triggered,
             this, [=]() {
-                this->createNewTab(true);
+                this->createNewTab(true, false);
                 auto topWidget = KiwixApp::instance()->getMainWindow()->getTopWidget();
                 topWidget->getSearchBar().setFocus(Qt::MouseFocusReason);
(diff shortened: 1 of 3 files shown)


src/tabbar.cpp
```diff
@@ -127,10 +127,15 @@ void TabBar::setNewTabButton()
     setTabButton(idx, QTabBar::RightSide, Q_NULLPTR);
 }
 
-ZimView* TabBar::createNewTab(bool setCurrent)
+ZimView* TabBar::createNewTab(bool setCurrent, bool adjacentToPreviousTab)
```
mgautierfr, line 130: It may be better to name the argument `adjacentToCurrentTab`

mgautierfr (changes requested):
Just a small change in the parameter name and we can merge.

juuz0:
Should I squash?

mgautierfr:
> Should I squash?

Yes please. The two commits are not two intermediate steps. The second is a fix for the first one. So no need to keep both changes separated (but it was nice to have the two commits to simplify the review)

juuz0:
Alright, done.

mgautierfr (approved):

mgautierfr:
I've rebased and fix the conflict.

PR #717, opened 2021-11-03
Added functionality to always open new URLs in background.

### LINKED ISSUE #637: Open new tab in background
When you open (right clic) a link in a new tab, would be better to just create it and keep the current active one open.

First reported in #635

juuz0:
Fixes #637 

https://user-images.githubusercontent.com/26814569/140111794-d7b91997-2b51-4c4a-a6ac-f7d936f2d02d.mp4


mgautierfr (changes requested):
This doesn't work if you open the tab using the middle click button.
It should behave the same than with the context menu.

mgautierfr (approved):

PR #719, opened 2021-11-12
Improvements to Settings Panel

### LINKED ISSUE #718: Few things to fix/discuss around zoom
In general the zoom works like this:
* There is a default zoom level at 100%
* This default zoom level can be changed in the settings
* This default zoom level applies to any ZIM, ZIM article if not defined differently.
* Each ZIM file can have its own level of zoom which is saved in the settings (but not available via UI)
* This ZIM dedicated ZOOM settings can be changed (and automatically saved in the settings) with the wheel, button, meny entry or shortcuts

To me all of this works properly and as it should. I see only a few things to improve/fix:
* The settings entry should be labeld "Default zoom level" instead of just "Zoom level"
* The settings should apply immediatly to all tabs (which don't have already a manually set zoom level).
* <s>Zoom reset seems not to bring to default zoom level, it seems to bring to 100%</s> works!

juuz0:
Fixes #718

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -65,7 +65,7 @@
     "cancel":"Cancel",
     "apply":"Apply",
     "port-for-local-kiwix-server-setting":"Port for local Kiwix server :",
-    "zoom-level-setting":"Zoom level :",
+    "zoom-level-setting":"Default zoom level :",
     "download-directory-setting":"Download directory :",
     "reset":"Reset",
     "browse":"Browse",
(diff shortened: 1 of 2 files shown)


kelson42 (approved):
From a user perspective, it works as expeted. Thx!

mgautierfr (changes requested):
We are good on the code.

I would prefer a better commit message however.
The current one is a bit to general and do not say what is done.
You can think of the commit title (the first line) as a description of what is done and what will be put in the Changelog.
It would be better to have something like "Change existing tab's zoom when setting's zoom is changed"

---

The same way (but it is to be pedantic, and I will not block the PR for this), as you are doing two different, unrelated things (change the labeling and change the behavior), it would have been nice to have two commits.

juuz0:
Thank you for the suggestions! I have followed them (and also learned how to split a commit into 2 xD)

mgautierfr (approved):

PR #724, opened 2021-11-18
Remove space before colon and question mark in strings

amire80:
Space is necessary before ':' and '?' in the French language,
but it's not necessary in English and in most other languages.


resources/i18n/en.json
```diff
@@ -117,7 +117,7 @@
     "no":"no",
     "no-filter":"no filter",
     "open-link-in-web-browser":"Open link in web browser",
-    "download-dir-dialog-title":"Are you sure you want to change the download directory ?",
-    "download-dir-dialog-msg":"The new download directory path will be :\n{{DIRECTORY}}",
+    "download-dir-dialog-title":"Are you sure you want to change the download directory",
```
mgautierfr, line 120: You have removed the question mark with the space.

mgautierfr (changes requested):

amire80:
:facepalm: 

I put the question mark back. Thanks for the careful review! :)

kelson42:
Actually, and this is something we agreed with @juuz0, to me we should remove the colon character on the settings panel.

kelson42:
Thanks @amire80 for your PR, I have taken the essence of it in #725.

amire80:
Thanks!

PR #730, opened 2021-11-26
Allow multiple files to open from command line

### LINKED ISSUE #720: Open multiple ZIM files from command line is broken
If I launch `./kiwix-desktop Australia.zim`

I get this:
image

We should see the welcome page of the ZIM file and the ZIM file should be added to the local library.

This seems to be a regression, quite sure it was working before.

juuz0:
Fixes #720 

Might have to think about which zim file gets the current tab focus though. Currently it is the last one in arguments

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -17,22 +17,20 @@ int main(int argc, char *argv[])
     KiwixApp a(argc, argv);
     a.setApplicationVersion(version);
     QCommandLineParser parser;
-    QString zimfile;
     parser.setApplicationDescription(QStringLiteral("The Kiwix Desktop is a viewer/manager of ZIM files for GNU/Linux and Microsoft Windows OSes."));
     parser.addHelpOption();
     parser.addVersionOption();
     parser.addPositionalArgument("zimfile", "The zim file");
     parser.process(a);
     auto positionalArguments = parser.positionalArguments();
-    if (positionalArguments.size() >= 1) {
-        zimfile = parser.positionalArguments().at(0);
-    }
     if (a.isRunning()) {
-        a.sendMessage(zimfile);
+        for (QString zimfile : positionalArguments) {
+            a.sendMessage(zimfile);
+        }
         return 0;
     }
     a.init();
-    if (!zimfile.isEmpty()) {
+    for (QString zimfile : positionalArguments) {
         a.openZimFile(zimfile);
     }
     return a.exec();


kelson42:
@juuz0 Thx for the quick PR, I will test it.

kelson42 (changes requested):
Maybe the ticket was not clear enough, but if I call:
```
kiwix-desktop wikipedia_es_history_maxi_2021-10.zim wikipedia_ru_physics_mini_2021-02.zim
```

Then only the welcome page of the `wikipedia_es_history_maxi_2021-10.zim` is shown, and actually it should behaves like if I would run:
```
$kiwix-desktop wikipedia_es_history_maxi_2021-10.zim &
$kiwix-desktop wikipedia_ru_physics_mini_2021-02.zim
```

... wich means that the welcome page of the second ZIM should be open in a new tab.

juuz0:
@kelson42 
Isnt that exactly what the current changes are doing?

https://user-images.githubusercontent.com/26814569/143605350-057d2a12-6d08-4a18-a388-7c708233ba0c.mp4

kelson42:
@juuz0 Somehow my testing was wrong. It works now. Sorry for the burden. Works for me now.

kelson42 (approved):

kelson42:
@mgautierfr Does the code looks good?

src/main.cpp
```diff
@@ -17,23 +17,23 @@ int main(int argc, char *argv[])
...
     parser.addVersionOption();
     parser.addPositionalArgument("zimfile", "The zim file");
     parser.process(a);
     auto positionalArguments = parser.positionalArguments();
-    if (positionalArguments.size() >= 1) {
-        zimfile = parser.positionalArguments().at(0);
-    }
     if (a.isRunning()) {
-        a.sendMessage(zimfile);
+        for(QString zimfile : positionalArguments) {
```
kelson42, line 27: Maybe just put a space after the `for`

src/main.cpp
```diff
@@ -17,23 +17,23 @@ int main(int argc, char *argv[])
...
+            a.sendMessage(zimfile);
+        }
         return 0;
     }
     a.init();
-    if (!zimfile.isEmpty()) {
-        a.openZimFile(zimfile);
+    if (positionalArguments.size() >= 1) {
+        for(QString zimfile : positionalArguments) {
```
kelson42, line 34: Here too!

src/main.cpp
juuz0: Doing. Somehow I always seem to do this mistake :facepalm:

src/main.cpp
```diff
@@ -17,23 +17,23 @@ int main(int argc, char *argv[])
...
-        a.sendMessage(zimfile);
+        for (QString zimfile : positionalArguments) {
+            a.sendMessage(zimfile);
+        }
         return 0;
     }
     a.init();
-    if (!zimfile.isEmpty()) {
-        a.openZimFile(zimfile);
+    if (positionalArguments.size() >= 1) {
```
mgautierfr, line 33: Do we need this check ? If `positionalArguments` is empty, the loop will do nothing.

mgautierfr (changes requested):

mgautierfr:
And there is a typo in the commit message :
> Allow multiple files to **be** open**ed** from command line

juuz0:
Fixed :heavy_check_mark:

kelson42 (approved):

PR #735, opened 2021-12-01
Add server IP Chooser

### LINKED ISSUE #322: Kiwix serve IP address is not configurable
Using the linx appimage version of kiwix 2.0 rc4, the port for the kiwix server can be configured, but not the IP address. I noticed this because I was evaluating the version on a machine with multiple network connections (ethernet and wifi). I was expecting the default to be localhost on 127.0.0.1, which would be the correct default security-concsious setting to not open the port. As it stands, kiwix (arbitrarily?) chose one of the network-facing routes.

Further, in my case kiwix chose a route that was physically disconnected at the time, so because there is no way alter the default, the server is useless on my local network.

### LINKED ISSUE #726: Use new libkiwix server URL help to get the localhost URL
And replace custom code we have currently.

Based on https://github.com/kiwix/libkiwix/pull/622

juuz0:
Fixes #322
Fixes #726 
Needs https://github.com/kiwix/libkiwix/pull/622

### DIFF
--- a/src/localkiwixserver.cpp
+++ b/src/localkiwixserver.cpp
@@ -72,6 +72,7 @@ void LocalKiwixServer::runOrStopServer()
         KiwixApp::instance()->getSettingsManager()->setKiwixServerPort(m_port);
         mp_server->setAddress(ui->IpChooser->currentText().toStdString());
         ui->IpAddress->setText("http://" + m_ipAddress + ":" + QString::number(m_port));
+        ui->IpAddress->setReadOnly(true);
         if (!mp_server->start()) {
             QMessageBox messageBox;
             messageBox.critical(0,gt("error-title"),gt("error-launch-server-message"));
--- a/src/localkiwixserver.cpp
+++ b/src/localkiwixserver.cpp
@@ -26,8 +26,8 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
     connect(ui->KiwixServerButton, SIGNAL(clicked()), this, SLOT(runOrStopServer()));
     connect(ui->OpenInBrowserButton, SIGNAL(clicked()), this, SLOT(openInBrowser()));
     connect(ui->closeButton, &QPushButton::clicked, this, &LocalKiwixServer::close);
-    connect(ui->PortChooser, &QLineEdit::textChanged, ui->PortChooser, [=](const QString &text){
-        if(text.toInt() > 65535 || text.toInt() < 1) {
+    connect(ui->PortChooser, &QLineEdit::textChanged, ui->PortChooser, [=](const QString &text) {
+        if (text.toInt() > 65535 || text.toInt() < 1) {
             QString validText = text;
             validText.chop(1);
             ui->PortChooser->setText(validText);
@@ -46,7 +46,7 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
     for (const auto &interface : interfaces) {
         ui->IpChooser->addItem(interface);
     }
-    ui->IpChooser->setCurrentText(KiwixApp::instance()->getSettingsManager()->getKiwixServerIp());
+    ui->IpChooser->setCurrentText(KiwixApp::instance()->getSettingsManager()->getKiwixServerIpAddress());
     ui->PortChooser->setText(QString::number(m_port));
     ui->PortChooser->setValidator(new QIntValidator(1, 65535, this));
     ui->KiwixServerButton->setStyleSheet("QPushButton {background-color: RoyalBlue;"
--- a/src/localkiwixserver.cpp
+++ b/src/localkiwixserver.cpp
@@ -27,10 +27,13 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
     connect(ui->OpenInBrowserButton, SIGNAL(clicked()), this, SLOT(openInBrowser()));
     connect(ui->closeButton, &QPushButton::clicked, this, &LocalKiwixServer::close);
     connect(ui->PortChooser, &QLineEdit::textChanged, ui->PortChooser, [=](const QString &text) {
-        if (text.toInt() > 65535 || text.toInt() < 1) {
+        if (text.toInt() > 65535) {
             QString validText = text;
             validText.chop(1);
             ui->PortChooser->setText(validText);
+        } else if (text.toInt() < 1) {
+            QString defaultPort = QString::number(KiwixApp::instance()->getSettingsManager()->getKiwixServerPort());
+            ui->PortChooser->setText(defaultPort);
         }
     });
 
@@ -41,8 +44,8 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
         QString ip = QString::fromStdString(interfacePair.second);
         interfaces.push_back(ip);
     }
-    interfaces.push_back(QString("0.0.0.0"));
     std::sort(interfaces.begin(), interfaces.end());
+    interfaces.push_front(QString(gt("all")));
     for (const auto &interface : interfaces) {
         ui->IpChooser->addItem(interface);
     }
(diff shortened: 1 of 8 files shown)


kelson42:
@juuz0 Thank you, I will give a check.

juuz0:
Updated for latest changes in https://github.com/kiwix/libkiwix/pull/622
Good for a check now

kelson42:
@juuz0 It does not compile for me:
```
$ make
g++ -c -pipe -std=c++11 -I/usr/local/include/ -I/usr/local/include -I/usr/include/x86_64-linux-gnu -I/usr/include/p11-kit-1 -O2 -Wall -W -D_REENTRANT -fPIC -DVERSION=2.0.5 -DQT_DEPRECATED_WARNINGS -DQT_NO_DEBUG -DQT_WEBENGINEWIDGETS_LIB -DQT_WEBENGINECORE_LIB -DQT_QUICK_LIB -DQT_PRINTSUPPORT_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_WEBCHANNEL_LIB -DQT_QML_LIB -DQT_NETWORK_LIB -DQT_POSITIONING_LIB -DQT_CORE_LIB -I. -Isubprojects/QtSingleApplication/src -isystem /usr/include/x86_64-linux-gnu/qt5 -isystem /usr/include/x86_64-linux-gnu/qt5/QtWebEngineWidgets -isystem /usr/include/x86_64-linux-gnu/qt5/QtWebEngineCore -isystem /usr/include/x86_64-linux-gnu/qt5/QtQuick -isystem /usr/include/x86_64-linux-gnu/qt5/QtPrintSupport -isystem /usr/include/x86_64-linux-gnu/qt5/QtWidgets -isystem /usr/include/x86_64-linux-gnu/qt5/QtGui -isystem /usr/include/x86_64-linux-gnu/qt5/QtWebChannel -isystem /usr/include/x86_64-linux-gnu/qt5/QtQml -isystem /usr/include/x86_64-linux-gnu/qt5/QtNetwork -isystem /usr/include/x86_64-linux-gnu/qt5/QtPositioning -isystem /usr/include/x86_64-linux-gnu/qt5/QtCore -I. -I. -I/usr/lib/x86_64-linux-gnu/qt5/mkspecs/linux-g++ -o contenttypefilter.o src/contenttypefilter.cpp
In file included from src/contenttypefilter.cpp:2:
src/kiwixapp.h:142:21: error: cannot declare field ‘KiwixApp::m_nameMapper’ to be of abstract type ‘KiwixApp::NameMapperProxy’
  142 |     NameMapperProxy m_nameMapper;
      |                     ^~~~~~~~~~~~
src/kiwixapp.h:110:9: note:   because the following virtual functions are pure within ‘KiwixApp::NameMapperProxy’:
  110 |   class NameMapperProxy : public kiwix::NameMapper {
      |         ^~~~~~~~~~~~~~~
In file included from src/kiwixapp.h:18,
                 from src/contenttypefilter.cpp:2:
/usr/local/include/kiwix/name_mapper.h:36:25: note: 	‘virtual std::string kiwix::NameMapper::getNameForId(const string&) const’
   36 |     virtual std::string getNameForId(const std::string& id) const = 0;
      |                         ^~~~~~~~~~~~
/usr/local/include/kiwix/name_mapper.h:37:25: note: 	‘virtual std::string kiwix::NameMapper::getIdForName(const string&) const’
   37 |     virtual std::string getIdForName(const std::string& name) const = 0;
      |                         ^~~~~~~~~~~~
make: *** [Makefile:1017: contenttypefilter.o] Error 1
```

juuz0:
@kelson42 because of changes in https://github.com/kiwix/kiwix-desktop/pull/732
Rebased now, should work

kelson42:
It compiles now, but I have a dubious warning:
```
src/settingsview.cpp: In member function ‘void SettingsView::init(int, int, const QString&)’:
src/settingsview.cpp:27:29: warning: unused parameter ‘port’ [-Wunused-parameter]
   27 | void SettingsView::init(int port, int zoomPercent, const QString &dir)
      |                         ~~~~^~~~
```

kelson42 (changes requested):
Looks almost perfect, but needs a bit of beautiful.

juuz0:
@kelson42 I agree on the beauty, haha. Will do them changes

juuz0:
Updated @kelson42 
Have still kept borders (although decreased size and used light-grey colour like in the invision designs) because without them it was looking weird. Let me know if they have to be removed :)
Followed all other points, I think :heavy_check_mark:

kelson42 (changes requested):
Few things left:
* Default port should be `8080`
* `99999` input should not be possible at port number. Only values from 1 through 65535 are authorized.

juuz0:
Both done

kelson42 (approved):
Well done

kelson42:
@mgautierfr Can you please have a look to the code?

src/localkiwixserver.cpp
```diff
@@ -24,16 +25,30 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
...
-    connect(KiwixApp::instance()->getSettingsManager(), &SettingsManager::portChanged,
-            this, [=](int port) { m_port = port; });
     connect(ui->closeButton, &QPushButton::clicked, this, &LocalKiwixServer::close);
-    const QHostAddress &localhost = QHostAddress(QHostAddress::LocalHost);
-    for (const QHostAddress &address: QNetworkInterface::allAddresses()) {
-        if (address.protocol() == QAbstractSocket::IPv4Protocol && address != localhost) {
-            m_ipAddress = address.toString();
-            break;
+    connect(ui->PortChooser, &QLineEdit::textChanged, ui->PortChooser, [=](const QString &text){
+        if(text.toInt() > 65535 || text.toInt() < 1) {
```
kelson42, line 30: Please put a space after the `if`

juuz0: Done along with some inconsistency in ipAddress related identifiers

kelson42:
@juuz0 I have rebased and hopefuly, now that kiwix/libkiwix#622 has been merged, the CI will pass.

But I have remarked a small glitch:
image

Why in the selectbox, we have an empty entry, or at least a white margin, at the bottom of the IP list?

juuz0:
@kelson42 was an issue with padding, I removed the padding from there and decreased the Port Edit box's to 2px.

kelson42:
@mgautierfr We really need to move ahead with this PR, please review it quickly.

src/localkiwixserver.cpp
```diff
@@ -24,16 +25,30 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
...
+    });
+
+    const auto interfacesMap = kiwix::getNetworkInterfaces();
+    QVector<QString> interfaces;
+    interfaces.reserve(interfacesMap.size() + 1);
+    for (const auto &interfacePair : interfacesMap) {
+        QString ip = QString::fromStdString(interfacePair.second);
+        interfaces.push_back(ip);
+    }
+    interfaces.push_back(QString("0.0.0.0"));
```
mgautierfr, line 44: I would rename it to "default" or "all".
If the user don't know about IP addresses and just want the "best" value, `0.0.0.0` doesn't help a lot.

src/localkiwixserver.cpp
```diff
@@ -65,8 +78,19 @@ void LocalKiwixServer::openInBrowser()
 void LocalKiwixServer::runOrStopServer()
 {
     if (!m_active) {
+        auto settingsManager = KiwixApp::instance()->getSettingsManager();
+        m_port = ui->PortChooser->text().toInt();
+        if (m_port == 0) {
+            m_port = settingsManager->getKiwixServerPort();
+        }
```
mgautierfr, line 85: Is there a situation where `ui->PortChooser->text().toInt()` returns 0 ?

mgautierfr (changes requested):
> Only values from 1 through 65535 are authorized.

I wonder if we should go even further and prevent port < 1024.
Those ports are "Well-known ports" and most of the time (on linux at least), the user needs to be root to use them. We probably don't want to use them. (See https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers)

src/localkiwixserver.cpp
juuz0: Yep, when the line edit is left empty

mgautierfr: Shouldn't be checked at UI level ?
And if it appears if the line is empty, it is better to check if the line is empty.

juuz0: :+1: Will do

kelson42:
@mgautierfr Not sure about that, port 80 is legit for example. I would avoid that discussion for the moment or at least pospone it in an other ticket.

mgautierfr:
port 80 is legit for a "real" server. Something running as a daemon and probably with some monitoring (and on linux you need root access)
I don't think it is the use case of kiwix-serve through kiwix-desktop. But we can postpone yes.

mgautierfr:
Look good to me. @juuz0 Please rebase/fixup before we merge.

kelson42:
@juuz0 Thx, great to see this PR finally merged and working so well!

juuz0:
Yay! :D

PR #743, opened 2021-12-11
Enable directory monitoring for new zims

### LINKED ISSUE #713: Monitor ZIM files from a user-chosen directory
Some people like to put all of their files at a different directory (usually case with windows users).
If I'm not mistaken, the current way right now is to open each zim file so it adds that in the local library.

A good way could be to :
Add an option in Settings to import all zim files from a directory (user chooses dialog box) [check the load local zim files option]
Settings (1)

In a way, this can also simplify https://github.com/kiwix/kiwix-desktop/wiki/Kiwix-desktop-on-an-USB-Drive-(Portable-version) 

This is a follow-up of https://github.com/kiwix/kiwix-desktop/issues/644#issuecomment-955949075

juuz0:
Fixes #713

### DIFF
--- a/resources/css/_settingsManager.css
+++ b/resources/css/_settingsManager.css
@@ -1,5 +1,5 @@
 QWidget#widget {
-    max-width: 500;
+    max-width: 600;
 }
 
 QLabel,
@@ -23,8 +23,8 @@ QLabel#settingsLabel {
     font-weight: bold;
 }
 QSpinBox,
-QLabel#downloadDirPath {
-    font: bold 12pt "Cantarell";
+QLabel#downloadDirPath, #monitorDirPath {
+    font: bold 12pt;
 }
 QPushButton {
     background-color: white;
@@ -37,4 +37,19 @@ QPushButton {
 QPushButton:hover {
     background-color: #3366cc;
     color: white;
+}
+
+#monitorDirLabel {
+    padding-right: 0;
+}
+
+#monitorHelp {
+    padding: 1;
+    margin: 0;
+    background: #666666;
+    color: white;
+    border-radius: 2px;
+    max-height: 18px;
+    min-width: 14px;
+    font: bold;
 }
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -69,7 +69,10 @@
     "port-for-local-kiwix-server-setting":"Port for local Kiwix server",
     "zoom-level-setting":"Default zoom level",
     "download-directory-setting":"Download directory",
+    "monitor-directory-setting":"Monitor directory",
+    "monitor-directory-default":"Not set",
     "reset":"Reset",
+    "clear":"Clear",
     "browse":"Browse",
     "about-kiwix-desktop-title":"Kiwix Desktop",
     "about-kiwix-desktop-description":"Kiwix enables you to have the whole Wikipedia at hand wherever you go! On a boat, in the middle of nowhere or in jail, Kiwix gives you access to the whole human knowledge. You don't need Internet, everything is stored on your computer.",
@@ -124,5 +127,10 @@
     "download-dir-dialog-msg":"The new download directory path will be:\n{{DIRECTORY}}",
     "invalid-port":"Invalid port",
     "zim-open-fail-title":"Invalid file",
-    "zim-open-fail-text":"The ZIM file <b>{{ZIM}}</b> cannot be opened properly. It will be removed from your library."
+    "zim-open-fail-text":"The ZIM file <b>{{ZIM}}</b> cannot be opened properly. It will be removed from your library.",
+    "monitor-dir-dialog-title":"Are you sure you want to change the monitor directory?",
+    "monitor-dir-dialog-msg":"The new monitor directory path will be:\n{{DIRECTORY}}",
+    "monitor-clear-dir-dialog-title":"Are you sure you want to clear the monitor directory?",
+    "monitor-clear-dir-dialog-msg":"This will stop checking the monitor directory for new ZIM files.",
+    "monitor-directory-tooltip":"All ZIM files in this directory will be automatically added to the library."
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -273,6 +273,10 @@ bool KiwixApp::isCurrentArticleBookmarked()
     return false;
 }
 
+void KiwixApp::setMonitorDir(const QString &dir) {
+    m_settingsManager.setMonitorDir(dir);
+}
+
 #define CREATE_ACTION(ID, TEXT) \
     mpa_actions[ID] = new QAction(TEXT)
 #define SET_SHORTCUT(ID, TEXT, SHORTCUT) \
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -275,6 +275,10 @@ bool KiwixApp::isCurrentArticleBookmarked()
 
 void KiwixApp::setMonitorDir(const QString &dir) {
     m_settingsManager.setMonitorDir(dir);
+    m_library.setMonitorDirZims(QStringList());
+    if (dir != "") {
+        m_library.asyncLoadMonitorDir(dir);
+    }
 }
 
 #define CREATE_ACTION(ID, TEXT) \
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -101,6 +101,11 @@ void KiwixApp::init()
             this->openZimFile(message);
         }
     });
+    QString monitorDir = m_settingsManager.getMonitorDir();
+    if (monitorDir != "") {
+        m_library.setMonitorDirZims(m_library.getLibraryZimsFromDir(monitorDir));
+        m_library.asyncLoadMonitorDir(monitorDir);
+    }
 }
 
 KiwixApp::~KiwixApp()
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -101,9 +101,13 @@ void KiwixApp::init()
             this->openZimFile(message);
         }
     });
+    connect(&m_watcher, &QFileSystemWatcher::directoryChanged, this, [=](QString monitorDir) {
+        m_library.asyncLoadMonitorDir(monitorDir);
+    });
     QString monitorDir = m_settingsManager.getMonitorDir();
     if (monitorDir != "") {
         m_library.setMonitorDirZims(m_library.getLibraryZimsFromDir(monitorDir));
+        m_watcher.addPath(monitorDir);
         m_library.asyncLoadMonitorDir(monitorDir);
     }
 }
@@ -281,7 +285,11 @@ bool KiwixApp::isCurrentArticleBookmarked()
 void KiwixApp::setMonitorDir(const QString &dir) {
     m_settingsManager.setMonitorDir(dir);
     m_library.setMonitorDirZims(QStringList());
+    for (auto path : m_watcher.directories()) {
+        m_watcher.removePath(path);
+    }
     if (dir != "") {
+        m_watcher.addPath(dir);
         m_library.asyncLoadMonitorDir(dir);
     }
 }
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -5,6 +5,8 @@
 #include <kiwix/tools.h>
 
 #include <QtDebug>
+#include <QtConcurrent/QtConcurrentRun>
+
 
 class LibraryManipulator: public kiwix::LibraryManipulator {
   public:
@@ -121,6 +123,45 @@ void Library::save()
 {
     m_library.writeToFile(kiwix::appendToDirectory(m_libraryDirectory.toStdString(),"library.xml"));
     m_library.writeBookmarksToFile(kiwix::appendToDirectory(m_libraryDirectory.toStdString(), "library.bookmarks.xml"));
+}
+
+void Library::setMonitorDirZims(QStringList zimList)
+{
+    m_monitorDirZims = zimList;
+}
+
+void Library::loadMonitorDir(QString monitorDir)
+{
+    QMutex mutex;
+    QMutexLocker locker(&mutex);
+    const QDir dir(monitorDir);
+    QStringList newDirEntries = dir.entryList({"*.zim"});
+    for (auto &str : newDirEntries) {
+        str = monitorDir + QDir::separator() + str;
+    }
+    QSet<QString> newDir = QSet<QString>::fromList(newDirEntries);
+    QStringList oldDirEntries = m_monitorDirZims;
+    QSet<QString> oldDir = QSet<QString>::fromList(oldDirEntries);
+    QStringList addedZims = (newDir - oldDir).values();
+    QStringList removedZims = (oldDir - newDir).values();
+    setMonitorDirZims(newDir.values());
+    auto manipulator = LibraryManipulator(this);
+    auto manager = kiwix::Manager(&manipulator);
+    for (auto book : addedZims) {
+        manager.addBookFromPath(book.toStdString());
+    }
+    for (auto bookPath : removedZims) {
+        removeBookFromLibraryById(QString::fromStdString(m_library.getBookByPath(bookPath.toStdString()).getId()));
+    }
+    emit(booksChanged());
+    save();
+}
+
+void Library::asyncLoadMonitorDir(QString dir)
+{
+    QtConcurrent::run( [=]() {
+        loadMonitorDir(dir);
+    });
 }
 
 const kiwix::Book &Library::getBookById(QString id) const
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -76,7 +76,7 @@ std::shared_ptr<kiwix::Searcher> Library::getSearcher(const QString &zimId)
     return searcher;
 }
 
-QStringList Library::getBookIds()
+QStringList Library::getBookIds() const
 {
     QStringList list;
     for(auto& id: m_library.getBooksIds()) {
@@ -85,7 +85,7 @@ QStringList Library::getBookIds()
     return list;
 }
 
-QStringList Library::listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending)
+QStringList Library::listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending) const
 {
     QStringList list;
     auto bookIds = m_library.filter(filter);
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -130,6 +130,19 @@ void Library::setMonitorDirZims(QStringList zimList)
     m_monitorDirZims = zimList;
 }
 
+QStringList Library::getLibraryZimsFromDir(QString dir) const
+{
+    QStringList zimsInDir;
+    for (auto str : getBookIds()) {
+        auto filePath = QString::fromStdString(getBookById(str).getPath());
+        QDir absoluteDir = QFileInfo(filePath).absoluteDir();
+        if (absoluteDir == dir) {
+            zimsInDir.push_back(filePath);
+        }
+    }
+    return zimsInDir;
+}
+
 void Library::loadMonitorDir(QString monitorDir)
 {
     QMutex mutex;
--- a/src/library.h
+++ b/src/library.h
@@ -33,11 +33,15 @@ public:
     QStringList getBookIds() const;
     QStringList listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending) const;
     const std::vector<kiwix::Bookmark> getBookmarks(bool onlyValidBookmarks = false) const { return m_library.getBookmarks(onlyValidBookmarks); }
+    QStringList getLibraryZimsFromDir(QString dir) const;
+    void setMonitorDirZims(QStringList zimList);
     void addBookToLibrary(kiwix::Book& book);
     void removeBookFromLibraryById(const QString& id);
     void addBookmark(kiwix::Bookmark& bookmark);
     void removeBookmark(const QString& zimId, const QString& url);
     void save();
+    void loadMonitorDir(QString dir);
+    void asyncLoadMonitorDir(QString dir);
     kiwix::Library& getKiwixLibrary() { return m_library; }
 public slots:
     const kiwix::Book& getBookById(QString id) const;
@@ -49,6 +53,7 @@ signals:
 private:
     kiwix::Library m_library;
     QString m_libraryDirectory;
+    QStringList m_monitorDirZims;
 friend class LibraryManipulator;
 };
 
--- a/src/library.h
+++ b/src/library.h
@@ -30,9 +30,9 @@ public:
     QString openBookFromPath(const QString& zimPath);
     std::shared_ptr<kiwix::Reader> getReader(const QString& zimId);
     std::shared_ptr<kiwix::Searcher> getSearcher(const QString& zimId);
-    QStringList getBookIds();
-    QStringList listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending);
-    const std::vector<kiwix::Bookmark> getBookmarks(bool onlyValidBookmarks = false) { return m_library.getBookmarks(onlyValidBookmarks); }
+    QStringList getBookIds() const;
+    QStringList listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending) const;
+    const std::vector<kiwix::Bookmark> getBookmarks(bool onlyValidBookmarks = false) const { return m_library.getBookmarks(onlyValidBookmarks); }
     void addBookToLibrary(kiwix::Book& book);
     void removeBookFromLibraryById(const QString& id);
     void addBookmark(kiwix::Bookmark& bookmark);
--- a/src/settingsmanager.cpp
+++ b/src/settingsmanager.cpp
@@ -73,12 +73,11 @@ void SettingsManager::setZoomFactor(qreal zoomFactor)
     emit(zoomChanged(zoomFactor));
 }
 
-bool SettingsManager::setDownloadDir(QString downloadDir)
+void SettingsManager::setDownloadDir(QString downloadDir)
 {
     m_downloadDir = downloadDir;
     m_settings.setValue("download/dir", downloadDir);
     emit(downloadDirChanged(downloadDir));
-    return true;
 }
 
 void SettingsManager::initSettings()
--- a/src/settingsmanager.cpp
+++ b/src/settingsmanager.cpp
@@ -16,7 +16,7 @@ SettingsView* SettingsManager::getView()
 {
     if (m_view == nullptr) {
         auto view = new SettingsView();
-        view->init(m_zoomFactor * 100, m_downloadDir);
+        view->init(m_zoomFactor * 100, m_downloadDir, m_monitorDir);
         connect(view, &QObject::destroyed, this, [=]() { m_view = nullptr; });
         m_view = view;
     }
@@ -80,10 +80,18 @@ void SettingsManager::setDownloadDir(QString downloadDir)
     emit(downloadDirChanged(downloadDir));
 }
 
+void SettingsManager::setMonitorDir(QString monitorDir)
+{
+    m_monitorDir = monitorDir;
+    m_settings.setValue("monitor/dir", monitorDir);
+    emit(monitorDirChanged(monitorDir));
+}
+
 void SettingsManager::initSettings()
 {
     m_kiwixServerPort = m_settings.value("localKiwixServer/port", 8080).toInt();
     m_zoomFactor = m_settings.value("view/zoomFactor", 1).toDouble();
     m_downloadDir = m_settings.value("download/dir", QString::fromStdString(kiwix::getDataDirectory())).toString();
     m_kiwixServerIpAddress = m_settings.value("localKiwixServer/ipAddress", QString("0.0.0.0")).toString();
+    m_monitorDir = m_settings.value("monitor/dir", QString("")).toString();
 }
--- a/src/settingsmanager.h
+++ b/src/settingsmanager.h
@@ -31,7 +31,7 @@ public slots:
     void setKiwixServerPort(int port);
     void setKiwixServerIpAddress(QString ipAddress);
     void setZoomFactor(qreal zoomFactor);
-    bool setDownloadDir(QString downloadDir);
+    void setDownloadDir(QString downloadDir);
 private:
     void initSettings();
 
--- a/src/settingsmanager.h
+++ b/src/settingsmanager.h
@@ -22,16 +22,16 @@ public:
     bool settingsExists(const QString &key);
     QVariant getSettings(const QString &key);
     qreal getZoomFactorByZimId(const QString &id);
+    int getKiwixServerPort() const { return m_kiwixServerPort; }
+    QString getKiwixServerIpAddress() const { return m_kiwixServerIpAddress; }
+    qreal getZoomFactor() const { return m_zoomFactor; }
+    QString getDownloadDir() const { return m_downloadDir; }
 
 public slots:
     void setKiwixServerPort(int port);
-    int getKiwixServerPort() { return m_kiwixServerPort; };
     void setKiwixServerIpAddress(QString ipAddress);
-    QString getKiwixServerIpAddress() { return m_kiwixServerIpAddress; };
     void setZoomFactor(qreal zoomFactor);
-    qreal getZoomFactor() { return m_zoomFactor; };
     bool setDownloadDir(QString downloadDir);
-    QString getDownloadDir() { return m_downloadDir; }
 private:
     void initSettings();
 
--- a/src/settingsmanager.h
+++ b/src/settingsmanager.h
@@ -26,12 +26,14 @@ public:
     QString getKiwixServerIpAddress() const { return m_kiwixServerIpAddress; }
     qreal getZoomFactor() const { return m_zoomFactor; }
     QString getDownloadDir() const { return m_downloadDir; }
+    QString getMonitorDir() const { return m_monitorDir; }
 
 public slots:
     void setKiwixServerPort(int port);
     void setKiwixServerIpAddress(QString ipAddress);
     void setZoomFactor(qreal zoomFactor);
     void setDownloadDir(QString downloadDir);
+    void setMonitorDir(QString monitorDir);
 private:
     void initSettings();
 
@@ -39,6 +41,7 @@ signals:
     void portChanged(int port);
     void zoomChanged(qreal zoomFactor);
     void downloadDirChanged(QString downloadDir);
+    void monitorDirChanged(QString monitorDir);
 
 private:
     QSettings m_settings;
@@ -47,6 +50,7 @@ private:
     QString m_kiwixServerIpAddress;
     qreal m_zoomFactor;
     QString m_downloadDir;
+    QString m_monitorDir;
 };
 
 #endif // SETTINGSMANAGER_H
--- a/src/settingsview.cpp
+++ b/src/settingsview.cpp
@@ -24,25 +24,31 @@ SettingsView::SettingsView(QWidget *parent)
     ui->resetButton->setText(gt("reset"));
     ui->browseButton->setText(gt("browse"));
 }
+
 void SettingsView::init(int zoomPercent, const QString &dir)
 {
     ui->zoomPercentSpinBox->setValue(zoomPercent);
     ui->downloadDirPath->setText(dir);
 }
-bool SettingsView::confirmDialogDownloadDir(const QString& dir)
+
+bool SettingsView::confirmDialog( QString messageText, QString messageTitle)
 {
-    auto text = gt("download-dir-dialog-msg");
-    text = text.replace("{{DIRECTORY}}", dir);
     QMessageBox msgBox(
         QMessageBox::Question, //Icon
-        gt("download-dir-dialog-title"), //Title
-        text, //Text
+        messageTitle, //Title
+        messageText, //Text
         QMessageBox::Ok | QMessageBox::Cancel //Buttons
     );
     msgBox.setDefaultButton(QMessageBox::Ok);
 
     int ret = msgBox.exec();
     return (ret == QMessageBox::Ok);
+}
+
+bool SettingsView::confirmDialogDownloadDir(const QString &dir) {
+    auto messageText = gt("download-dir-dialog-msg");
+    messageText = messageText.replace("{{DIRECTORY}}", dir);
+    return confirmDialog(messageText, gt("download-dir-dialog-title"));
 }
 
 void SettingsView::resetDownloadDir()
--- a/src/settingsview.cpp
+++ b/src/settingsview.cpp
@@ -16,21 +16,32 @@ SettingsView::SettingsView(QWidget *parent)
     connect(ui->zoomPercentSpinBox, QOverload<int>::of(&QSpinBox::valueChanged), this, &SettingsView::setZoom);
     connect(ui->browseButton, &QPushButton::clicked, this, &SettingsView::browseDownloadDir);
     connect(ui->resetButton, &QPushButton::clicked, this, &SettingsView::resetDownloadDir);
+    connect(ui->monitorBrowse, &QPushButton::clicked, this, &SettingsView::browseMonitorDir);
+    connect(ui->monitorClear, &QPushButton::clicked, this, &SettingsView::clearMonitorDir);
     connect(KiwixApp::instance()->getSettingsManager(), &SettingsManager::downloadDirChanged, this, &SettingsView::onDownloadDirChanged);
+    connect(KiwixApp::instance()->getSettingsManager(), &SettingsManager::monitorDirChanged, this, &SettingsView::onMonitorDirChanged);
     connect(KiwixApp::instance()->getSettingsManager(), &SettingsManager::zoomChanged, this, &SettingsView::onZoomChanged);
     ui->settingsLabel->setText(gt("settings"));
     ui->zoomPercentLabel->setText(gt("zoom-level-setting"));
     ui->downloadDirLabel->setText(gt("download-directory-setting"));
+    ui->monitorDirLabel->setText(gt("monitor-directory-setting"));
     ui->resetButton->setText(gt("reset"));
     ui->browseButton->setText(gt("browse"));
+    ui->monitorClear->setText(gt("clear"));
+    ui->monitorBrowse->setText(gt("browse"));
+    ui->monitorHelp->setText("<b>?</b>");
+    ui->monitorHelp->setToolTip(gt("monitor-directory-tooltip"));
 }
 
-void SettingsView::init(int zoomPercent, const QString &dir)
+void SettingsView::init(int zoomPercent, const QString &downloadDir, const QString &monitorDir)
 {
     ui->zoomPercentSpinBox->setValue(zoomPercent);
-    ui->downloadDirPath->setText(dir);
+    ui->downloadDirPath->setText(downloadDir);
+    if (monitorDir == QString()) {
+        ui->monitorClear->hide();
+    }
+    ui->monitorDirPath->setText(monitorDir);
 }
-
 bool SettingsView::confirmDialog( QString messageText, QString messageTitle)
 {
     QMessageBox msgBox(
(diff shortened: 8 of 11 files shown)


kelson42:
@mgautierfr Would you be able please to make a first pass and explain how to use libkiwix to allow this refresh please?

juuz0:
@kelson42 Are the UI changes ok?

kelson42:
> @kelson42 Are the UI changes ok?

Few remarks:
* I have tried and it complains that there is no `library.xml`. I don't understand why. The feature should monitor `.zim` files not `.xml` files.
* The button "reset" shoudl better called "clear"
* We should have somehow a hint (maybe a small question mark badge and when we are over a tooltipe with an explanation sentence is displayed)
* If there is nothing defined, the "clear/reset" button should not be displayed.

juuz0:
@kelson42 Recommend tooltip message?

kelson42:
"All ZIM files in this directory will be automatically added to the library".

juuz0:
Did points 2,3,4 for now.
Will check 1

juuz0:
This might require changes on how to approach but does all the work mentioned in the issue for now.

@kelson42 ping!

kelson42 (changes requested):
Few remarks:
* Question mark icon could be a bit bigger, it is difficult to point on it.
* Triggering the tooltip immediatly when the mouse is over (and deactivate) when not over it immediatly would be better
* I have put monitoring on "Downloads" folder and then ZIM files where loaded properly in the local library, but then I downloaded with the Web browser a new ZIM file and it did not appear.
* Same with a ZIM file aready in "Downloads" that is renamed with an other extension. The file were still to be seen in the local library and if I click open the whole kiwix-desktop crash! (so an other bug).

To conclude, the monitoring part of the feature seems to not work. This seems to be still to code.

juuz0:
Thank you for the review, quick question.
> I have put monitoring on "Downloads" folder and then ZIM files where loaded properly in the local library, but then I downloaded with the Web browser a new ZIM file and it did not appear.

When you say you downloaded and it didnt appear, do you mean after restart of kiwix-desktop or while kiwix-desktop was running (haven't implemented that yet oops :))

kelson42:
@juuz0 while running, it shoudl imediatky appera in the library.

juuz0:
@kelson42 yes correct, file system watcher, I'll implement that 👍

kelson42:
@juuz0 Any news on this PR?

juuz0:
> Same with a ZIM file aready in "Downloads" that is renamed with an other extension. The file were still to be seen in the local library and if I click open the whole kiwix-desktop crash! (so an other bug).
Related to https://github.com/kiwix/kiwix-desktop/pull/760 I think

> I have put monitoring on "Downloads" folder and then ZIM files where loaded properly in the local library, but then I downloaded with the Web browser a new ZIM file and it did not appear.
Done

> Question mark icon could be a bit bigger, it is difficult to point on it.
Done

> Triggering the tooltip immediatly when the mouse is over (and deactivate) when not over it immediatly would be better
I've used the inbuilt method for tooltip so not really sure how to make it fast

kelson42:
> > Same with a ZIM file aready in "Downloads" that is renamed with an other extension. The file were still to be seen in the local library and if I click open the whole kiwix-desktop crash! (so an other bug).
> 
> Related to #760 I think

@juuz0 This still does not work. The problem here is not that it crashes (this is indeed handled in #760), the problem is that this is that the book is still visible in the library although the file has been removed! Actually the otherway around does not work either: adding a new ZIM file in the directory does not make the file automatically appearing in the library. It appears only after a reboot, but it should appears immediatly. How do you test this feature?

juuz0:
> The problem here is not that it crashes (this is indeed handled in #760), the problem is that this is that the book is still visible in the library although the file has been removed.

Ah, okay gotcha

> Actually the otherway around does not work either: adding a new ZIM file in the directory does not make the file automatically appearing in the library. It appears only after a reboot, but it should appears immediatly. How do you test this feature?

Weird, it should work.
I have tested by putting Monitoring in Downloads and then downloading a file - gets added
Cut paste something into the monitor directory - gets added.
Will send a screencast

juuz0:
This is the required functionality, right?

https://user-images.githubusercontent.com/26814569/149726988-114b4fde-4520-4cc1-9884-514f7f1251ea.mp4

resources/css/_settingsManager.css
```diff
@@ -23,8 +23,8 @@ QLabel#settingsLabel {
     font-weight: bold;
 }
 QSpinBox,
-QLabel#downloadDirPath {
-    font: bold 12pt "Cantarell";
```
kelson42, line 27: Is that normal that the font has disappeared here?

resources/i18n/en.json
```diff
@@ -120,5 +123,11 @@
     "open-link-in-web-browser":"Open link in web browser",
     "download-dir-dialog-title":"Are you sure you want to change the download directory?",
     "download-dir-dialog-msg":"The new download directory path will be:\n{{DIRECTORY}}",
+    "monitor-dir-dialog-title":"Are you sure you want to change the monitor directory?",
+    "monitor-dir-dialog-msg":"The new monitor directory path will be:\n{{DIRECTORY}}",
+    "monitor-clear-dir-dialog-title":"Are you sure you want to clear the monitor directory?",
+    "monitor-clear-dir-dialog-msg":"This will stop checking the monitor directory for new zim files.",
```
kelson42, line 129: "This will stop checking the monitor directory for new ZIM files."

src/kiwixapp.cpp
```diff
@@ -95,6 +96,19 @@ void KiwixApp::init()
...
         }
     });
+    connect(&m_watcher, &QFileSystemWatcher::directoryChanged, this, [=](QString monitorDir) {
+        m_library.loadMonitorDir(monitorDir);
+    });
+    QMutex mutex;
+    mutex.lock();
+    QFuture<void> future = QtConcurrent::run( [=]() {
+        QString monitorDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
+        if(monitorDir != "") {
```
kelson42, line 106: space after `if`

src/settingsview.cpp
```diff
@@ -17,29 +17,42 @@ SettingsView::SettingsView(QWidget *parent)
...
     ui->settingsLabel->setText(gt("settings"));
     ui->serverPortLabel->setText(gt("port-for-local-kiwix-server-setting"));
     ui->zoomPercentLabel->setText(gt("zoom-level-setting"));
     ui->downloadDirLabel->setText(gt("download-directory-setting"));
+    ui->monitorDirLabel->setText(gt("monitor-directory-setting"));
     ui->resetButton->setText(gt("reset"));
     ui->browseButton->setText(gt("browse"));
+    ui->monitorClear->setText(gt("clear"));
+    auto monitorDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
+    if(monitorDir == QString()) {
```
kelson42, line 35: space after `if`

src/settingsview.cpp
```diff
@@ -72,11 +85,48 @@ void SettingsView::browseDownloadDir()
...
+    if (confirmDirDialog(dir, gt("download-dir-dialog-msg"), gt("download-dir-dialog-title"))) {
         KiwixApp::instance()->getSettingsManager()->setDownloadDir(dir);
     }
 }
 
+void SettingsView::browseMonitorDir()
+{
+    const auto &monitorDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
+    QString previousDir;
+    if(monitorDir == "") {
```
kelson42, line 97: space after `if`

kelson42 (approved):
Support deletion of ZIM file in monitored directory still to be done.

kelson42 (changes requested):
Sorry, approved to quickly.

resources/css/_settingsManager.css
juuz0: Everywhere else, there was no font attribute. So I think it's globally added already, no need to add again

juuz0:
> Support deletion of ZIM file in monitored directory still to be done.
Should work now

kelson42:
@veloman-yunkan This is ready for code review.

veloman-yunkan:
> @veloman-yunkan This is ready for code review.

@kelson42 Do you want me to review this PR? I am asking because I am not assigned as a reviewer here.

src/library.cpp
```diff
@@ -120,6 +120,25 @@ void Library::save()
...
+    auto manipulator = LibraryManipulator(this);
+    auto manager = kiwix::Manager(&manipulator);
+    std::vector<std::string> paths{kiwix::getDataDirectory() + "/" + "library.xml"};
+    manager.reload(paths);
+    QDirIterator it(dir, {"*.zim"}, QDir::Files);
+    QString fullBookPath;
+    while (it.hasNext()) {
+        fullBookPath = it.next();
+        if(!manager.addBookFromPath(fullBookPath.toStdString())) {
+            qWarning()<<"Couldn't add "<<fullBookPath<<" in the library.";
```
veloman-yunkan, line 134: When a new ZIM file appears in MonitorDir, isn't this warning printed for all books that were present in the library before that?

veloman-yunkan (changes requested):
Please clean your commit history by rebasing&fixing-up before addressing my comments. Then address my comments as new or fixup commits where appropriate.

src/kiwixapp.cpp
```diff
@@ -51,6 +52,12 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
     QFontDatabase::addApplicationFont(":/fonts/Selawik/selawksl.ttf");
     QFontDatabase::addApplicationFont(":/fonts/Selawik/selawk.ttf");
     setFont(QFont("Selawik"));
+    QMutex mutex;
+    mutex.lock();
+    QFuture<void> future = QtConcurrent::run( [=]() {
+        m_library.loadMonitorDir(KiwixApp::instance()->getSettingsManager()->getMonitorDir());
+    });
+    mutex.unlock();
```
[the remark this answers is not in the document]
juuz0: This https://doc.qt.io/qt-5/qmutexlocker.html?
Never knew, will use it thanks

src/library.cpp
juuz0: It doesn't print warnings. I kind of assumed addBookFromPath would already check for existing book

src/settingsview.cpp
```diff
@@ -102,23 +119,21 @@ void SettingsView::browseMonitorDir()
...
-        msgBox.setText(gt("monitor-directory-invalid"));
-        msgBox.exec();
     }
 }
 
 void SettingsView::clearMonitorDir()
 {
     auto dir = QString();
-    if (confirmDirDialog(dir, gt("monitor-clear-dir-dialog-msg"), gt("monitor-clear-dir-dialog-title"))) {
+    if (confirmDialogMonitorDir(dir, true)) {
```
veloman-yunkan, line 133: Since in this case you don't need to replace the `{{DIRECTORY}}` parameter in the message you can call `confirmDirDialog()` here directly:

```
if (confirmDirDialog(gt("monitor-clear-dir-dialog-msg"), gt("monitor-clear-dir-dialog-title"))) {
```

src/settingsview.cpp
```diff
@@ -58,14 +56,33 @@ bool SettingsView::confirmDirDialog(const QString& dir, QString messageText, QSt
...
+    if (clearDir) {
+        messageText = gt("monitor-clear-dir-dialog-msg");
+        messageTitle = gt("monitor-clear-dir-dialog-title");
+    } else {
+        messageText = gt("monitor-dir-dialog-msg");
+        messageText = messageText.replace("{{DIRECTORY}}", dir);
+        messageTitle = gt("monitor-dir-dialog-title");
+    }
+    return confirmDialog(messageText, messageTitle);
+}
```
veloman-yunkan, line 76: In my previous review I missed that for the monitor-directory we have two separate messages, and one of them isn't parameterized. This helper isn't justified for the latter case where `confirmDialog()` can be called directly. Please drop the `clearDir` argument and make this wrapper similar to `confirmDialogDownloadDir()`.

src/settingsview.cpp
```diff
@@ -102,23 +119,21 @@ void SettingsView::browseMonitorDir()
     if (dir == monitorDir || dir.isEmpty()) {
         return;
     }
-    if (confirmDirDialog(dir, gt("monitor-dir-dialog-msg"), gt("monitor-dir-dialog-title"))) {
+    if (confirmDialogMonitorDir(dir, false)) {
         KiwixApp::instance()->getSettingsManager()->setMonitorDir(dir);
         ui->monitorClear->show();
         KiwixApp::instance()->getLibrary()->loadMonitorDir(dir);
         KiwixApp::instance()->getWatcher()->addPath(dir);
```
veloman-yunkan, line 126: Shouldn't you `KiwixApp::instance()->getWatcher()->removePath(previousDir);` here too (when `previousDir` is not empty)?

src/library.cpp
veloman-yunkan: I was confused here, but still my feeling that something is wrong remains. In such an implementation, whenever a new ZIM file appears, all books are re-read. @kelson42 is that acceptable?

src/kiwixapp.cpp
```diff
@@ -93,6 +100,19 @@ void KiwixApp::init()
             this->openZimFile(message);
         }
     });
+    connect(&m_watcher, &QFileSystemWatcher::directoryChanged, this, [=](QString monitorDir) {
+        m_library.loadMonitorDir(monitorDir);
+    });
```
veloman-yunkan, line 99: After rebasing and fixing-up this change appears earlier than the commit in which `m_watcher` is introduced. My personal protection against such issues after non-trivial history clean-up is to replay the history and make sure that each commit compiles (and doesn't break the unit tests).

veloman-yunkan (changes requested):

src/library.cpp
kelson42: @veloman-yunkan @juuz0 No, if this can be avoided. Opening a ZIM file (and its Xapian indexes) is not a cheap operation. Therefore we should only do the necessary.

@veloman-yunkan You already had a similar challenge with monitoring a library in kiwix-serve. Maybe you can give a hint or two to @juuz0 here about the best path to follow.

juuz0: QFileSystemWatcher doesn't provide any other useful information

Comparing directory files before and after any changes in directory is another way. Not sure about any other

veloman-yunkan: > @veloman-yunkan You already had a similar challenge with monitoring a library in kiwix-serve. Maybe you can give a hint or two to @juuz0 here about the best path to follow.

Monitoring a library XML file is different from monitoring a directory for new or removed ZIM files. @juuz0 already knows how to do it:

> Comparing directory files before and after any changes in directory is another way.

kelson42: @juuz0 If I understand properly, QFileSystemWatcher is not able to tell you what has changed, which files have been removed, updated or even deleted. So a kind of diff will be needed. For performance reason this should be achieved without having to read the files themself, so each file shoudl be listed and identified based in filename, size and modification tine. The a diff shoukd be done. Without been prescriptive, a subclassing seems apropriate here,

juuz0:
New method adopted, please check :)
(Didn't subclass because it looked too extra for something which can be done in `loadMonitorDir()` function and also needed some settings information)

src/settingsview.cpp
```diff
@@ -119,22 +113,27 @@ void SettingsView::browseMonitorDir()
     if (dir == monitorDir || dir.isEmpty()) {
         return;
     }
-    if (confirmDialogMonitorDir(dir, false)) {
+    if (confirmDialogMonitorDir(dir)) {
         KiwixApp::instance()->getSettingsManager()->setMonitorDir(dir);
         ui->monitorClear->show();
-        KiwixApp::instance()->getLibrary()->loadMonitorDir(dir);
         KiwixApp::instance()->getWatcher()->addPath(dir);
+        auto previousDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
```
veloman-yunkan, line 120: `previousDir` must be obtained *before* the call to `setMonitorDir()`

src/library.cpp
```diff
@@ -123,25 +125,45 @@ void Library::save()
...
-            emit(booksChanged());
+    emit(booksChanged());
+    if (!newZims.empty()) {
+        QString fullBookPath;
+        for (auto book : newZims) {
+            fullBookPath = monitorDir + QDir::separator() + book;
+            if (!manager.addBookFromPath(fullBookPath.toStdString())) {
+                qWarning()<<"Couldn't add "<<fullBookPath<<" in the library.";
+            } else {
+                emit(booksChanged());
```
veloman-yunkan, line 151: Isn't it better to emit this signal only once in the end?

src/settingsmanager.h
```diff
@@ -25,15 +25,17 @@ class SettingsManager : public QObject
     int getKiwixServerPort() const { return m_kiwixServerPort; };
     QString getKiwixServerIpAddress() const { return m_kiwixServerIpAddress; };
     qreal getZoomFactor() const { return m_zoomFactor; };
-    QString getDownloadDir() const { return m_downloadDir; }
-    QString getMonitorDir() const { return m_monitorDir; }
+    QString getDownloadDir() const { return m_downloadDir; };
+    QString getMonitorDir() const { return m_monitorDir; };
```
veloman-yunkan, line 29: Instead of adding semicolons on the end-of-line, you'd better removed such superfluous punctuation on the other neighbouring lines

src/settingsview.cpp
```diff
@@ -119,22 +113,27 @@ void SettingsView::browseMonitorDir()
...
+        if(previousDir != "") {
+            KiwixApp::instance()->getWatcher()->removePath(previousDir);
+        }
+        KiwixApp::instance()->getLibrary()->loadMonitorDir(dir);
```
veloman-yunkan, line 124: 1. This call to thread-unsafe `loadMonitorDir()` (instead of the thread-safe `loadMonitorDirInThread()`) can become a source of random (though, probably, extremely rare) glitches (when it happens to be executed concurrently with a call of itself triggered by `QFileSystemWatcher::directoryChanged`).
2. `SettingsManager::m_monitorDirZims` must be cleared when changing the monitory directory, otherwise inside `loadMonitorDir()` called from here you will be comparing the contents of two different directories (which may have unexpected and hardly desirable effects)

veloman-yunkan (changes requested):
For the next review please rebase & fixup your commit history so that I can review the cleaned-up PR. Don't forget to make sure that each commit in the final history compiles.

src/settingsview.cpp
juuz0: Wow, that fixes something I was concerned about, thank you!

src/settingsview.cpp
juuz0: Thanks, missed that :}

juuz0:
The last 2 fixups are the new changes, and all previous commits compile individually too :+1:

src/library.cpp
```diff
@@ -123,6 +125,45 @@ void Library::save()
...
+}
+
+void Library::loadMonitorDirInThread(QString dir)
+{
+    QMutex mutex;
+    QMutexLocker locker(&mutex);
+    QFuture<void> future = QtConcurrent::run( [=]() {
+        loadMonitorDir(dir);
+    });
+}
```
veloman-yunkan, line 163: Such combination of `QMutex` and `QFuture` is meaningless. The `QFuture` destructor [neither waits nor cancels the asynchronous computation](https://doc.qt.io/qt-5/qfuture.html#dtor.QFuture). Therefore when this function terminates (and the mutex is unlocked) `loadMonitorDir()` may (and mostly will) still be running.

I think the mutex should be moved inside `loadMonitorDir()`.

I also recommend renaming `loadMonitorDirInThread()` to `loadMonitorDirAsync()` or `asyncLoadMonitorDir()`.

src/library.cpp
```diff
@@ -123,6 +125,43 @@ void Library::save()
...
+    manager.reload(paths);
+    if (!newZims.empty()) {
+        QString fullBookPath;
+        for (auto book : newZims) {
+            fullBookPath = monitorDir + QDir::separator() + book;
+            if (!manager.addBookFromPath(fullBookPath.toStdString())) {
+                qWarning()<<"Couldn't add "<<fullBookPath<<" in the library.";
+            }
+        }
+    }
```
veloman-yunkan, line 151: The `if` is unnecessary. Please remove it and declare `fullBookPath` inside the loop.

src/settingsmanager.cpp
```diff
@@ -81,10 +81,26 @@ bool SettingsManager::setDownloadDir(QString downloadDir)
...
+}
+
 void SettingsManager::initSettings()
 {
     m_kiwixServerPort = m_settings.value("localKiwixServer/port", 8080).toInt();
     m_zoomFactor = m_settings.value("view/zoomFactor", 1).toDouble();
     m_downloadDir = m_settings.value("download/dir", QString::fromStdString(kiwix::getDataDirectory())).toString();
     m_kiwixServerIpAddress = m_settings.value("localKiwixServer/ipAddress", QString("0.0.0.0")).toString();
+    m_monitorDir = m_settings.value("monitor/dir", QString("")).toString();
+    m_monitorDirZims = m_settings.value("monitor/zims", QStringList()).toStringList();
```
veloman-yunkan, line 105: Why do you need to store the list of ZIM files in settings?

src/settingsview.cpp
```diff
@@ -73,6 +97,47 @@ void SettingsView::browseDownloadDir()
...
+
+void SettingsView::clearMonitorDir()
+{
+    auto dir = QString();
+    if (confirmDialog(gt("monitor-clear-dir-dialog-msg"), gt("monitor-clear-dir-dialog-title"))) {
+        ui->monitorClear->hide();
+        auto previousDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
+        KiwixApp::instance()->getWatcher()->removePath(previousDir);
+        KiwixApp::instance()->getSettingsManager()->setMonitorDir(dir);
+        KiwixApp::instance()->getSettingsManager()->setMonitorDirZims(QStringList());
```
veloman-yunkan, line 137: What about resetting/clearing the list of ZIM files in `setMonitorDir()`?

src/settingsmanager.h
```diff
@@ -22,22 +22,18 @@ class SettingsManager : public QObject
...
 public slots:
     void setKiwixServerPort(int port);
-    int getKiwixServerPort() { return m_kiwixServerPort; };
     void setKiwixServerIpAddress(QString ipAddress);
-    QString getKiwixServerIpAddress() { return m_kiwixServerIpAddress; };
     void setZoomFactor(qreal zoomFactor);
-    qreal getZoomFactor() { return m_zoomFactor; };
     bool setDownloadDir(QString downloadDir);
-    QString getDownloadDir() { return m_downloadDir; }
     bool setMonitorDir(QString monitorDir);
```
veloman-yunkan, line 37: This commit ("Move getters in settingsmanager out of slots") looks strange. Consider `getKiwixServerPort()`. The commit says that there were two (const non-slot and non-const slot) versions of it. The non-const variant was dropped, and for the const variant there was only a small punctuation fix.

veloman-yunkan (changes requested):
Reviewing the cleaned-up version uncovered a couple of more issues, but those seem to be the last ones.

src/settingsmanager.cpp
juuz0: need to check on startup also, part of this PR is also to fix https://github.com/kiwix/kiwix-desktop/issues/644

veloman-yunkan: OK. I was under the wrong impression that `m_settings.value(X, V)` sets the setting `X` to `V` (I now understand that `V` is the default value for the case when `X` is not yet defined). Then persisting the list of ZIM files from the monitored directory makes sense, though with the following reservation:

Ideally (and in accordance with the [DRY principle](https://en.wikipedia.org/wiki/Don't_repeat_yourself)) on startup we should build the list of known files from the monitored directory by filtering the library XML file. The current solution can be an acceptable hack if the right approach is significantly more difficult to implement. Then, at least, the reason should be noted in a comment.

kelson42 (changes requested):
@juuz0 There is a problem with 2 deb packages which don't compile, see https://github.com/kiwix/kiwix-desktop/runs/4955340548?check_suite_focus=true

kelson42:
@juuz0 The `QSet` problem is probably that the constructor you use has been introduced in Qt 5.14, see https://doc.qt.io/qt-5/qset.html#QSet-2. Older Ubuntu version might have an older minor version of Qt. Would you be able to get rid of this constructor?

juuz0:
@kelson42 Yep, right. If I recall correctly, I used that because Qt was complaining some deprecated warning, will do the changes by tomorrow probably.

kelson42:
@juuz0 I have rebased on latest master. Any problem to fix the `QSet` constructor call?

juuz0:
@kelson42 Thank you for the rebase, got too busy with college :/ But back on this today! :)

src/settingsmanager.cpp
juuz0: Thank you, it makes it so much clear

src/settingsmanager.h
juuz0: I must've dropped the non-const variant by mistake. But I don't really know what you suggest to do here (except bringing the non const variant back?). Any hints?

veloman-yunkan: > I must've dropped the non-const variant by mistake.

Non-const slot variants were converted to const non-slots based on [my suggestion](https://github.com/kiwix/kiwix-desktop/pull/743#discussion_r788023785). Then you also performed some [cosmetic clean-up](https://github.com/kiwix/kiwix-desktop/pull/743#discussion_r792698362). During history cleanup those two changes were combined in a way that doesn't accurately reflect what has happened.

src/settingsmanager.cpp
```diff
@@ -81,26 +81,20 @@ bool SettingsManager::setDownloadDir(QString downloadDir)
     return true;
 }
 
-bool SettingsManager::setMonitorDir(QString monitorDir)
+bool SettingsManager::setMonitorDir(QString monitorDir, QStringList zimList)
 {
     m_monitorDir = monitorDir;
     m_settings.setValue("monitor/dir", monitorDir);
+    m_monitorDirZims = zimList;
```
veloman-yunkan, line 88: `SettingsManager` doesn't seem the right place for keeping track of the ZIM files in the monitor dir. I think it should belong to `Library`.

src/kiwixapp.cpp
```diff
@@ -93,11 +93,20 @@ void KiwixApp::init()
...
     });
     QString monitorDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
+    QStringList zimsInMonitorDir;
+    for (auto str : m_library.getBookIds()) {
+        auto filePath = QString::fromStdString(m_library.getBookById(str).getPath());
+        QDir absoluteDir = QFileInfo(filePath).absoluteDir();
+        if (absoluteDir == monitorDir) {
+            zimsInMonitorDir.push_back(filePath);
+        }
+    }
```
veloman-yunkan, line 106: Please make this a member function of `Library` (and the list of zim files from the monitored directory should be a data-member of `Library` too).

veloman-yunkan (changes requested):
I will have to review this PR in its entirety again, so please clean-up/rewrite the history via a rebase.

src/library.cpp
```diff
@@ -123,6 +125,64 @@ void Library::save()
...
+    if (!addedZims.empty()) {
+        for (auto book : addedZims) {
+            manager.addBookFromPath(book.toStdString());
+        }
+    }
+    if (!removedZims.empty()) {
+        for (auto bookPath : removedZims) {
+            removeBookFromLibraryById(QString::fromStdString(m_library.getBookByPath(bookPath.toStdString()).getId()));
+        }
+    }
```
veloman-yunkan, line 174: The `if`-s are redundant - iterating over an empty list is a no-op anyway

src/library.cpp
```diff
@@ -123,6 +125,64 @@ void Library::save()
...
+    for (auto &str : newDirEntries) {
+        str = monitorDir + QDir::separator() + str;
+    }
+    QSet<QString> newDir = QSet<QString>::fromList(newDirEntries);
+    QStringList oldDirEntries = getMonitorDirZims();
+    QSet<QString> oldDir = QSet<QString>::fromList(oldDirEntries);
+    QSet<QString> addedZimSet = newDir - oldDir;
+    QStringList addedZims = addedZimSet.values();
+    QSet<QString> removedZimSet = oldDir - newDir;
+    QStringList removedZims = removedZimSet.values();
```
veloman-yunkan, line 161: Eliminate the intermediate variable `removedZimSet` by combining these two lines into one.

src/library.cpp
```diff
@@ -123,6 +125,64 @@ void Library::save()
...
+    const QDir dir(monitorDir);
+    QStringList newDirEntries = dir.entryList({"*.zim"});
+    for (auto &str : newDirEntries) {
+        str = monitorDir + QDir::separator() + str;
+    }
+    QSet<QString> newDir = QSet<QString>::fromList(newDirEntries);
+    QStringList oldDirEntries = getMonitorDirZims();
+    QSet<QString> oldDir = QSet<QString>::fromList(oldDirEntries);
+    QSet<QString> addedZimSet = newDir - oldDir;
+    QStringList addedZims = addedZimSet.values();
```
veloman-yunkan, line 159: Eliminate the intermediate variable `addedZimSet` by combining these two lines into one.

src/library.cpp
```diff
@@ -123,6 +125,64 @@ void Library::save()
...
+    }
+    emit(booksChanged());
+    save();
+}
+
+void Library::asyncLoadMonitorDir(QString dir)
+{
+    QFuture<void> future = QtConcurrent::run( [=]() {
+        loadMonitorDir(dir);
+    });
```
veloman-yunkan, line 183: The named variable doesn't buy as anything since we are not going to use it.

src/library.h
```diff
@@ -33,11 +33,16 @@ class Library : public QObject
     QStringList getBookIds();
     QStringList listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending);
     const std::vector<kiwix::Bookmark> getBookmarks(bool onlyValidBookmarks = false) { return m_library.getBookmarks(onlyValidBookmarks); }
+    QStringList getMonitorDirZims() const { return m_monitorDirZims; }
+    QStringList getLibraryZimsFromDir(QString dir);
```
veloman-yunkan, line 37: This method can be `const`

src/settingsmanager.h
```diff
@@ -22,6 +22,11 @@ class SettingsManager : public QObject
     bool settingsExists(const QString &key);
     QVariant getSettings(const QString &key);
     qreal getZoomFactorByZimId(const QString &id);
+    int getKiwixServerPort() const { return m_kiwixServerPort; }
+    QString getKiwixServerIpAddress() const { return m_kiwixServerIpAddress; }
+    qreal getZoomFactor() const { return m_zoomFactor; }
+    QString getDownloadDir() const { return m_downloadDir; }
+    QString getMonitorDir() const { return m_monitorDir; }
```
veloman-yunkan, line 29: This `const`-ification of `getXXX()` methods should be in a commit of its own (preferably as the first commit of this PR so that new accessors are directly added as `const` methods in the next commit).

src/settingsview.cpp
```diff
@@ -42,14 +42,12 @@ void SettingsView::init(int zoomPercent, const QString &downloadDir, const QStri
     ui->downloadDirPath->setText(downloadDir);
     ui->monitorDirPath->setText(monitorDir);
 }
-bool SettingsView::confirmDirDialog(const QString& dir, QString messageText, QString messageTitle)
+bool SettingsView::confirmDialog( QString messageText, QString messageTitle)
```
veloman-yunkan, line 45: It would be nice to have this refactoring (paired with the introduction of the `confirmDialogDownloadDir()` function) in a separate commit.

src/settingsview.cpp
```diff
@@ -114,9 +114,14 @@ void SettingsView::browseMonitorDir()
...
     if (confirmDialogMonitorDir(dir)) {
+        auto previousDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
         KiwixApp::instance()->getSettingsManager()->setMonitorDir(dir);
         KiwixApp::instance()->getLibrary()->setMonitorDirZims(QStringList());
         ui->monitorClear->show();
+        KiwixApp::instance()->getWatcher()->addPath(dir);
+        if (previousDir != "") {
+            KiwixApp::instance()->getWatcher()->removePath(previousDir);
+        }
         KiwixApp::instance()->getLibrary()->asyncLoadMonitorDir(dir);
```
veloman-yunkan, line 125: Instead of exposing the directory watcher from `KiwixApp` (the `getWatcher()` method) you can encapsulate this entire operation in a new method `KiwixApp::setMonitorDir(QString path)`.

veloman-yunkan (changes requested):
I propose to structure this PR as follows:

1. Start it with a couple of refactoring/clean-up commits (moving getters out of the slots sections and making them `const`, the `confirmDirDialog() → confirmDialog()` change etc).
2. Then add new functionality directly in a clean way (still as a series of small logically self-contained commits).

src/settingsview.cpp
```diff
@@ -58,14 +56,27 @@ bool SettingsView::confirmDirDialog(const QString& dir, QString messageText, QSt
     return (ret == QMessageBox::Ok);
 }
 
+bool SettingsView::confirmDialogDownloadDir(QString dir) {
+    auto messageText = gt("download-dir-dialog-msg");
+    messageText = messageText.replace("{{DIRECTORY}}", dir);
+    return confirmDialog(messageText, gt("download-dir-dialog-title"));
+}
```
veloman-yunkan, line 63: The current ordering of commits is not optimal
1. first `confirmDialogDownloadDir()` is replaced with a more general purpose (yet imperfect) `confirmDirDialog()` (which is later converted to `confirmDialog()`)
2. then `confirmDialogDownloadDir()` is restored as a small wrapper around `confirmDialog()`

A better sequencing is to
1. Extract `confirmDialog()` out of `confirmDialogDownloadDir()` (this is just a preparatory refactoring commit)
2. Use `confirmDialog()` for monitor dir in the next commit

src/settingsview.cpp
```diff
@@ -127,6 +128,7 @@ void SettingsView::clearMonitorDir()
     if (confirmDialog(gt("monitor-clear-dir-dialog-msg"), gt("monitor-clear-dir-dialog-title"))) {
         ui->monitorClear->hide();
         KiwixApp::instance()->getSettingsManager()->setMonitorDir(dir);
+        KiwixApp::instance()->setMonitorDir("");
         KiwixApp::instance()->getLibrary()->setMonitorDirZims(QStringList());
```
veloman-yunkan, line 132: `KiwixApp::instance()->getSettingsManager()->setMonitorDir(dir);` and `KiwixApp::instance()->getLibrary()->setMonitorDirZims(QStringList());` can be moved into `KiwixApp::setMonitorDir()`.

src/library.cpp
veloman-yunkan: This comment was not addressed (or the commit was lost)

src/library.cpp
veloman-yunkan: This comment was not addressed (or the commit was lost)

src/library.cpp
veloman-yunkan: This comment was not addressed (or the commit was lost)

src/library.cpp
veloman-yunkan: This comment was not addressed (or the commit was lost)

veloman-yunkan (changes requested):
@juuz0 Thanks and please bear with me a little more. I see that you are becoming an important contributor in kiwix-desktop that's why I am putting more rigor in the review of this PR than I would have done for a one-shot contribution.

A recurring pitfall while submitting one's work is when you don't review it yourself (this occurs to me too). Please before requesting the next review of this PR put on a hat of a reviewer and inspect all the changes as someone else's work.

src/library.h
```diff
@@ -30,9 +30,9 @@ class Library : public QObject
     QString openBookFromPath(const QString& zimPath);
     std::shared_ptr<kiwix::Reader> getReader(const QString& zimId);
     std::shared_ptr<kiwix::Searcher> getSearcher(const QString& zimId);
-    QStringList getBookIds();
-    QStringList listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending);
-    const std::vector<kiwix::Bookmark> getBookmarks(bool onlyValidBookmarks = false) { return m_library.getBookmarks(onlyValidBookmarks); }
+    QStringList getBookIds() const;
+    QStringList listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending) const;
+    const std::vector<kiwix::Bookmark> getBookmarks(bool onlyValidBookmarks = false) const { return m_library.getBookmarks(onlyValidBookmarks); }
```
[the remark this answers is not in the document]
juuz0: I'm sorry, i had a thought but then somehow still didn't follow it (probably because it looked too small). I understand now, thank you :)

src/settingsview.cpp
juuz0: > Use confirmDialog() for monitor dir in the next commit

Can this change be included the bigger commit (Add monitor directory entry in Settings panel)?

veloman-yunkan: That's what I meant

src/settingsview.cpp
juuz0: Do you mean the calls to those functions or full logic?

juuz0: I was asking if you mean to remove `SettingsManager::setMonitorDir` and `Library::setMonitorDirZims` from the app and use `KiwixApp::setMonitorDir` only OR just use them two functions in `KiwixApp::setMonitorDir`, but now I think you meant the latter

src/settingsview.cpp
```diff
@@ -90,6 +139,11 @@ void SettingsView::onDownloadDirChanged(const QString &dir)
     ui->downloadDirPath->setText(dir);
 }
 
+void SettingsView::onMonitorDirChanged(const QString &dir)
+{
+    ui->monitorDirPath->setText(dir);
+}
```
veloman-yunkan, line 145: There is incoherence between treatment of `ui->monitorClear` and `ui->monitorDirPath`. The former is updated in `SettingsView::clearMonitorDir()` and `SettingsView::browseMonitorDir()` but  omitted from this slot, while the opposite is true for `ui->monitorDirPath`.

You should localize all updates to the GUI side of monitor dir either in `SettingsView::clearMonitorDir()` and `SettingsView::browseMonitorDir()` (but then the `SettingsManager::monitorDirChanged()` signal and the `SettingsView::onMonitorDirChanged()` slot should be dropped), or in `SettingsView::onMonitorDirChanged()`.

src/library.cpp
veloman-yunkan: This comment is still not addressed

src/library.cpp
veloman-yunkan: This comment is still not addressed

src/library.h
```diff
@@ -33,11 +33,15 @@ class Library : public QObject
     QStringList getBookIds() const;
     QStringList listBookIds(const kiwix::Filter& filter, kiwix::supportedListSortBy sortBy, bool ascending) const;
     const std::vector<kiwix::Bookmark> getBookmarks(bool onlyValidBookmarks = false) const { return m_library.getBookmarks(onlyValidBookmarks); }
+    QStringList getMonitorDirZims() const { return m_monitorDirZims; }
```
veloman-yunkan, line 36: This function is actually not needed since it is only used (once) in `Library`

src/kiwixapp.cpp
```diff
@@ -92,6 +92,11 @@ void KiwixApp::init()
             this->openZimFile(message);
         }
     });
+    QString monitorDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
```
veloman-yunkan, line 98: Inside a `KiwixApp` member function `KiwixApp::instance() == this`

veloman-yunkan (changes requested):
Looks like this is the last request for changes. Please implement the fixes as fixup commits so that I don't need to review the entire PR again.

src/kiwixapp.cpp
```diff
@@ -92,6 +92,15 @@ void KiwixApp::init()
...
         }
     });
+    connect(&m_watcher, &QFileSystemWatcher::directoryChanged, this, [=](QString monitorDir) {
+        m_library.asyncLoadMonitorDir(monitorDir);
+    });
+    QString monitorDir = m_settingsManager.getMonitorDir();
+    if (monitorDir != "") {
+        m_library.setMonitorDirZims(m_library.getLibraryZimsFromDir(monitorDir));
+        m_library.asyncLoadMonitorDir(monitorDir);
+        m_watcher.addPath(monitorDir);
```
veloman-yunkan, line 102: With the current order of these two lines there is a small chance that a file appears or is deleted in the monitor directory after its contents has been read but before it started being watched.

src/settingsmanager.h
```diff
@@ -22,23 +22,26 @@ class SettingsManager : public QObject
...
 public slots:
     void setKiwixServerPort(int port);
-    int getKiwixServerPort() { return m_kiwixServerPort; };
     void setKiwixServerIpAddress(QString ipAddress);
-    QString getKiwixServerIpAddress() { return m_kiwixServerIpAddress; };
     void setZoomFactor(qreal zoomFactor);
-    qreal getZoomFactor() { return m_zoomFactor; };
     bool setDownloadDir(QString downloadDir);
-    QString getDownloadDir() { return m_downloadDir; }
+    bool setMonitorDir(QString monitorDir);
```
veloman-yunkan, line 36: Return type `bool` (as copy-pasted from `setDownloadDir()`) is meaningless with the current implementation. Change it to `void` (and fix the return type of ``setDownloadDir()` too, but in a separate non-fixup commit).

veloman-yunkan (changes requested):
I was about to approve this PR but noticed two more absolutely minor issues. Please fix them too and you will immediately have the green light. Promise.

src/settingsmanager.h
veloman-yunkan: > this commit should be before "Add monitor directory entry" right?

Yes

> Also, can I rebase the current 3 fixups?

Please don't rebase now. Let me see the last two fixes and then you will rebase the entire PR right before it is merged.

veloman-yunkan (commented):
Technically this is an approval. Please rebase and fixup.

veloman-yunkan (commented):
You didn't apply the fixups correctly. The correct command is (note the `--autosquash` option) 

```
git rebase -i --autosquash master
```
If you don't provide the `--autosquash` option, you have to reorder the commits and mark them as fixups for rebase manually. Please search for "fixup" in the manual page of [git rebase](http://git-scm.com/docs/git-rebase) and read the relevant sections.

Now you will have to reset the HEAD of your `monitorDirectory` branch to its previous state (find it out with `git reflog`) and repeat the rebase. Note that if you mess up, recovering will be a little more difficult. Read [this post](http://effectif.com/git/recovering-lost-git-commits) and look through the manual page of [git reset](http://git-scm.com/docs/git-reset) (and in general consult the git documentation from time to time; git is a powerful but also quite a complex tool, so those who want to master it must regularly study its reference manual).

juuz0:
That was a nice mess up, indeed I forgot `autosquash` somehow. Thanks for the links, I'll read them. But I do have a backup since I pushed those 5 fixups, can I rebase that correctly and force push it?

veloman-yunkan:
> That was a nice mess up, indeed I forgot `autosquash` somehow. Thanks for the links, I'll read them. But I do have a backup since I pushed those 5 fixups, can I rebase that correctly and force push it?

It's good that you have a backup. But let's keep the backup as a backup plan.

Don't hurry. I will be online until this is fixed. Basically you should be able to recover by reading just http://effectif.com/git/recovering-lost-git-commits. In my a previous comment I just wanted to emphasize that serious tools deserve serious attitude by also paying attention to their full documentation.

veloman-yunkan (approved):
@juuz0 Congratulations to us both for bringing this PR to a nearly perfect shape!

I hope that you have the same feeling and don't perceive me as a perverse faultfinder :smile:

juuz0:
@veloman-yunkan Congratulations indeed xD. I can't stress enough how much I've learned in this PR alone, much thankful :)

kelson42 (approved):
I rebase on `master` and will merge. Good job!

Issue #450, opened 2020-05-19
Books - ISBN


Ubuntu-2020:
In **Wikipedia** books looks like this: ISBN [978-0-8032-3944-9](https://en.wikipedia.org/wiki/Special:BookSources/978-0-8032-3944-9).
In **Kiwix** books looks like this: ISBN 978-0-8032-3944-9.
I think it will be good if in the **Kiwix** ISBN of the book will look the same as in **Wikipedia**.

PR #746, opened 2021-12-15
Code clean up: illiminate KiwixApp::SideBarType property

asashnov:
While working on #540 I discovered it difficult to understand why do we need KiwixApp::currentSideType  QObject property, and all the complicated logic around it:

* QObject property which can be accessed via JS / QML Script  KiwixApp::currentSide  (but it's not used from JS code)
* three-state enum KiwixApp::SideBarType
* KiwixApp::setSideBar(type)
* KiwixApp::toggleSideBar(type)
* signal  KiwixApp::currentSideTypeChanged
* Tabbar::onCurrentChanged(tab index) used these methods above

Actually, it's MainWindow who owns TabBar and SideBar, should know what to do with them. But it didn't.

Fixed by this PR.

Instead of complicated three-state logic straight-forward (and much shorter) code is written in MainWindow.

The behavior is the same as was before:
* ContentList is shown with the library tab
* after switching back to Settings or Zim document tab  Reading list hides if it was shown before.

If we don't like the current behavior it's better to change it by creating a new issue.
This PR doesn't change the behavior, just clean up the code.

The approach used by CREATE_ACTION_ONOFF_ICON_SHORTCUT  can be extended to ToggleFullscreenAction action also: QIcon may have state: https://doc.qt.io/qt-5/qicon.html#State-enum
and the action icon is chosen appropriately to its state, no need to manually change the icon every time QAction is toggled.

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -84,8 +84,6 @@ void KiwixApp::init()
     mp_tabWidget = mp_mainWindow->getTabBar();
     mp_tabWidget->setContentManagerView(mp_manager->getView());
     mp_tabWidget->setNewTabButton();
-    mp_mainWindow->getSideContentManager()->setContentManager(mp_manager);
-    setSideBar(CONTENTMANAGER_BAR);
     postInit();
     mp_errorDialog = new QErrorMessage(mp_mainWindow);
     setActivationWindow(mp_mainWindow);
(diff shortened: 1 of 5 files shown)


kelson42:
@asashnov which ticket is this PR related? I can not #450.

asashnov:
Reference is fixed, it must be #540. Sorry for the typo.

kelson42:
@mgautierfr Can you please review this soon, this is waiting for so long!

PR #751, opened 2021-12-20
Part 2 to fix #540: don't emit tabbar signal from zimview

### LINKED ISSUE #540: Codebase: emiting a signal forcedly from outside is a bad practice

asashnov:
Partial fix #540: stop emitting TabBar::webActionEnabledChanged from ZimView
as it's bad practice according to QObject documentation.

Dismiss way-round signal passing through KiwixApp::currentTitleChanged
Now it's local to MainWindow and its siblings (TabBar, ZimViews, TopWidget).

TabBar, TopWidget, SideBar- connect signals in MainWindow constructor
after MainWindow mp_ui->setupUi() completed.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -38,7 +38,7 @@ MainWindow::MainWindow(QWidget *parent) :
     connect(app->getAction(KiwixApp::KiwixServeAction), &QAction::triggered,
             mp_localKiwixServer, &QDialog::show);
 
-    connect(app, &KiwixApp::currentTitleChanged, this, [=](const QString& title) {
+    connect(mp_ui->tabBar, &TabBar::currentTitleChanged, this, [=](const QString& title) {
         if (!title.isEmpty() && !title.startsWith("zim://"))
             setWindowTitle(title + " - Kiwix");
         else
@@ -63,6 +63,14 @@ MainWindow::MainWindow(QWidget *parent) :
             mp_ui->mainToolBar, &TopWidget::updateBackForwardButtons);
     connect(mp_ui->tabBar, &TabBar::libraryPageDisplayed,
             this, &MainWindow::when_libraryPageDisplayed);
+
+    connect(mp_ui->tabBar, &TabBar::currentTitleChanged,
+            &(mp_ui->mainToolBar->getSearchBar()), &SearchBar::on_currentTitleChanged);
+
+    // This signal emited more often than the history really updated
+    // but for now we have no better signal for it.
+    connect(mp_ui->tabBar, &TabBar::currentTitleChanged,
+            mp_ui->mainToolBar, &TopWidget::updateBackForwardButtons);
 
     mp_ui->contentmanagerside->setContentManager(app->getContentManager());
     mp_ui->sideBar->setCurrentWidget(mp_ui->contentmanagerside);
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -72,6 +72,9 @@ MainWindow::MainWindow(QWidget *parent) :
     connect(mp_ui->tabBar, &TabBar::currentTitleChanged,
             mp_ui->mainToolBar, &TopWidget::updateBackForwardButtons);
 
+    connect(mp_ui->tabBar, &TabBar::webActionEnabledChanged,
+            mp_ui->mainToolBar, &TopWidget::handleWebActionEnabledChanged);
+
     mp_ui->contentmanagerside->setContentManager(app->getContentManager());
     mp_ui->sideBar->setCurrentWidget(mp_ui->contentmanagerside);
 }
--- a/src/tabbar.h
+++ b/src/tabbar.h
@@ -68,6 +68,7 @@ private:
 private slots:
     void onTabMoved(int from, int to);
     void onCurrentChanged(int index);
+    void onWebviewHistoryActionChanged(QWebEnginePage::WebAction action, bool enabled);
 };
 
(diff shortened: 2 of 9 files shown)


kelson42:
@asashnov Can you please rebase and remove the draft?

kelson42:
@veloman-yunkan Would you be able please to review this PR, unfortunately @mgautierfr is super busy these days.

veloman-yunkan (commented):
The set of changes in this PR is not so easy to understand and validate when viewed as one commit. I think that this PR can - and therefore must - be split into several smaller commits (with explanations in the commit messages where appropriate).

src/tabbar.h
```diff
@@ -68,6 +68,7 @@ public slots:
 private slots:
     void onTabMoved(int from, int to);
     void onCurrentChanged(int index);
+    void when_zimview_history_action_changed(QWebEnginePage::WebAction action, bool enabled);
```
veloman-yunkan, line 71: Usage of snake\_case here is inconsistent with camelCased function names above. I see that this is not the first such inconsistency but let's not fall victim to the broken windows effect. Also, probably it is better to replace 'when' with 'on'.

src/mainwindow.cpp
```diff
@@ -63,6 +63,15 @@ MainWindow::MainWindow(QWidget *parent) :
...
     connect(mp_ui->tabBar, &TabBar::libraryPageDisplayed,
             this, &MainWindow::when_libraryPageDisplayed);
+
+    connect(mp_ui->tabBar, &TabBar::currentTitleChanged,
+            &(mp_ui->mainToolBar->getSearchBar()), &SearchBar::on_currentTitleChanged);
+
+    // This signal emited more often than the history really updated
+    // but for now we have no better signal for it.
+    connect(mp_ui->tabBar, &TabBar::currentTitleChanged,
+            mp_ui->mainToolBar, &TopWidget::updateBackForwardButtons);
```
veloman-yunkan, line 73: Several lines above there is

```
    connect(mp_ui->tabBar, &QTabBar::currentChanged,
    mp_ui->mainToolBar, &TopWidget::updateBackForwardButtons);
```

Are they both justified? If yes, shouldn't `SearchBar::on_currentTitleChanged` be connected to `QTabBar::currentChanged`, too?

veloman-yunkan (changes requested):
Some commit titles exceed the recommended length limit and look ugly when split with an ellipsis

Looks like the commit titled '#540: don't emit TabBar::webActionEnabledChanged signal from ZimView' can be split further into two more-or-less independent changes.

src/mainwindow.cpp
asashnov: > Are they both justified? If yes, shouldn't `SearchBar::on_currentTitleChanged` be connected to `QTabBar::currentChanged`, too?

SearchBar::on_currentTitleChanged connection mistakenly fell in the other commit, fixed.

asashnov:
> Some commit titles exceed the recommended length limit and look ugly when split with an ellipsis
Fixed. Thanks for pointing this out. My further commits will be better from start.

> Looks like the commit titled '#540: don't emit TabBar::webActionEnabledChanged signal from ZimView' can be split further into two more-or-less independent changes.
Fixed. Split changes to commits reworked.

src/mainwindow.cpp
```diff
@@ -38,7 +38,7 @@ MainWindow::MainWindow(QWidget *parent) :
     connect(app->getAction(KiwixApp::KiwixServeAction), &QAction::triggered,
             mp_localKiwixServer, &QDialog::show);
 
-    connect(app, &KiwixApp::currentTitleChanged, this, [=](const QString& title) {
+    connect(mp_ui->tabBar, &TabBar::currentTitleChanged, this, [=](const QString& title) {
         if (!title.isEmpty() && !title.startsWith("zim://"))
             setWindowTitle(title + " - Kiwix");
         else
```
veloman-yunkan, line 44: There are three separate `connect(mp_ui->tabBar, &TabBar::currentTitleChanged, ...)` statements in the `MainWindow` constructor. Isn't it better to introduce a new private slot `MainWindow::onCurrentTabTitleChanged()` and localize the handling of that event entirely in one function?

veloman-yunkan (commented):
Only one question. If the answer to it is negative then this is an approval.

src/mainwindow.cpp
asashnov: I incline to leave it as separate 3 signal-slot connections.

veloman-yunkan (approved):

PR #759, opened 2022-01-03
Better (with dependences version) --version

kelson42:


src/kiwixapp.cpp
```diff
@@ -444,3 +445,11 @@ void KiwixApp::updateNameMapper()
 {
   m_nameMapper.update();
 }
+
+void KiwixApp::printVersions(std::ostream& out) {
+  out << version.toStdString() << std::endl;
+  out << "+ libqt " << QT_VERSION_STR << " (" << qVersion() << ")" << std::endl << std::endl;
```
veloman-yunkan, line 451: IMHO, this line will be confusing. The previous way of reporting the build-time and run-time versions was clearer.

kelson42: @veloman-yunkan Better now?

veloman-yunkan (approved):

PR #760, opened 2022-01-03
Add exception handling for book modification 

### LINKED ISSUE #758: Error by opening a file is not handled properly
Steps:
* Have a file in the local library
* Delete or rename that file in the filesystem
* Click "open" on that file in the local library

I get:
* A crash with:
```
$ ./kiwix-desktop
Compiled with Qt Version  5.12.8
Runtime Qt Version  5.12.8
add widget
terminate called after throwing an instance of 'std::runtime_error'
  what():  error 2 opening file "/home/kelson/Downloads/5.zim
Aborted (core dumped)
```

I would like:
* A proper error dialog with an error... and probably the entry should be deleted from the local library.

juuz0:
Fixes #758

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -15,6 +15,7 @@
     "home-page":"Home page",
     "main-menu":"Main menu",
     "print":"Print",
+    "print-page-error": "An error has occured while printing.",
     "new-tab":"New tab",
     "close-tab":"Close tab",
     "close":"Close",
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -121,5 +121,7 @@
     "open-link-in-web-browser":"Open link in web browser",
     "download-dir-dialog-title":"Are you sure you want to change the download directory?",
     "download-dir-dialog-msg":"The new download directory path will be:\n{{DIRECTORY}}",
-    "invalid-port":"Invalid port"
+    "invalid-port":"Invalid port",
+    "zim-open-fail-title":"Invalid file",
+    "zim-open-fail-text":"The ZIM file <b>{{ZIM}}</b> cannot be open properly. It will be removed from your library."
 }
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -48,6 +48,9 @@ QString Library::openBookFromPath(const QString &zimPath)
 
     kiwix::Manager manager(&m_library);
     auto id =  manager.addBookFromPathAndGetId(zimPath.toStdString());
+    if (id == "") {
+        throw std::invalid_argument("invalid zim file");
+    }
     save();
     emit(booksChanged());
     return QString::fromStdString(id);
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -10,6 +10,7 @@
 #include <QUrl>
 #include <QDir>
 #include <QStorageInfo>
+#include <QMessageBox>
 
 ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader, QObject *parent)
     : QObject(parent),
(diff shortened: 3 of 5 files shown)


juuz0:
Have encountered a small bug, fixing

resources/i18n/en.json
```diff
@@ -121,5 +121,6 @@
     "open-link-in-web-browser":"Open link in web browser",
     "download-dir-dialog-title":"Are you sure you want to change the download directory?",
     "download-dir-dialog-msg":"The new download directory path will be:\n{{DIRECTORY}}",
-    "invalid-port":"Invalid port"
+    "invalid-port":"Invalid port",
+    "zim-open-fail":"The zim file has been modified on disk and cannot be loaded properly.\nPlese undo the changes and re-open it."
```
kelson42, line 125: "The ZIM file %1 cannot be open properly. It will be removed from your library."

kelson42 (changes requested):
@juuz0 Beside changing the message, it looks good to me. Thx!

kelson42:
@juuz0 I wonder why starting `/kiwix-desktop file_which_does_not_exist.zim` does not generate the same error? Is there a good reason? Could we have the same behaviour?

resources/i18n/en.json
juuz0: What should be shown at %1? ZIM title or full path?

juuz0:
> @juuz0 I wonder why starting /kiwix-desktop file_which_does_not_exist.zim does not generate the same error? Is there a good reason? Could we have the same behaviour
Yes we should do that but the possibilities here are a bit different and so different message (file may/may not be in library) should be displayed
`kiwix-desktop justSomeRandomWords`
`kiwix-desktop aValidZimFile.zim`
`kiwix-desktop anInvalidZimFile.zim`

What do you think about this?

resources/i18n/en.json
kelson42: fullpath

juuz0:
>  I wonder why starting /kiwix-desktop file_which_does_not_exist.zim does not generate the same error? Is there a good reason? Could we have the same behaviour?

I have found that this is already implemented except missing a `throw` statement

kelson42 (changes requested):
It I launch with a non existing file from the command line I get something like this:
image

This is not the same box style like if the error is displayed from library and there is a checkbox "Show this message again which should not be there".

juuz0:
Consistent now, although I don't really like this copying almost the same code for MessageBox again (but its twice only so maybe fine!) 
Maybe re-implementing `KiwixApp::showMessage()` (currently uses default message box hence the "show this message again") is better...Would go with your inputs

src/tabbar.cpp
```diff
@@ -154,7 +154,11 @@ void TabBar::openUrl(const QUrl& url, bool newTab)
         webView = createNewTab(true, true)->getWebView();
     }
     QUITIFNULL(webView);
-    webView->setUrl(url);
+    try {
+        webView->setUrl(url);
+    } catch (const std::exception& e) {
+        throw;
+    }
```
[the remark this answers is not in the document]
juuz0: I had some more code there, removed later and forgot to remove this. Thanks! :)

src/library.cpp
```diff
@@ -48,6 +48,9 @@ QString Library::openBookFromPath(const QString &zimPath)
 
     kiwix::Manager manager(&m_library);
     auto id =  manager.addBookFromPathAndGetId(zimPath.toStdString());
+    if (id == "") {
+        throw std::invalid_argument("invalid zim file");
+    }
```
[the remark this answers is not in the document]
juuz0: Agreed, main cause of all this was `addBookFromPathAndGetId` not throwing expected exception.

kelson42 (approved):

mgautierfr (approved):

PR #774, opened 2022-01-24
set zoom factor after page has finished loading

### LINKED ISSUE #770: Zooming function is functionally flawed
OS: elementary OS, Ubuntu 20.04-based
Kiwix-desktop version: 2.0.5, flatpak edition from flathub

My main issue is that, using the search function, the zoom level gets reset to 100% by Kiwix.
Therefore, even if I have set the default zoom at 130%, that will apply only to the homepage of the Wiki loaded on Kiwix, and the hyperlinks clicked from there. But it will *not* work for any page belonging to the same Wiki when accessed through the search function.

For all users who search wiki pages, rather than navigating them through hyperlinks only, this is a major obstacle to usage of the software.
Please refer to the video below for a demonstration of the behaviour

https://user-images.githubusercontent.com/16632292/150642804-668b4277-a4c7-47c6-915b-642a942a920a.mp4

juuz0:
Fixes #770

### DIFF
--- a/src/webview.cpp
+++ b/src/webview.cpp
@@ -40,6 +40,29 @@ void WebViewForwardMenu::showEvent(QShowEvent *)
     setGeometry(geo);
 }
 
+QString getZimIdFromUrl(QUrl url)
+{
+    return url.host().split('.')[0];
+}
+
+QString getResultTypeFromUrl(QUrl url)
+{
+    return url.host().split('.')[1];
+}
+
+void WebView::applyCorrectZoomFactor() {
+    auto url = this->url();
+    auto settingsManager = KiwixApp::instance()->getSettingsManager();
+    qreal zoomFactor;
+    const bool isSearchResultsView = QUrlQuery(url).hasQueryItem("pattern") && (getResultTypeFromUrl(url) == "search");
+    if (isSearchResultsView) {
+        zoomFactor = settingsManager->getZoomFactor();
+    } else {
+        auto zimId = getZimIdFromUrl(url);
+        zoomFactor = settingsManager->getZoomFactorByZimId(zimId);
+    }
+    this->setZoomFactor(zoomFactor);
+}
 
 WebView::WebView(QWidget *parent)
     : QWebEngineView(parent)
@@ -48,6 +71,18 @@ WebView::WebView(QWidget *parent)
     QObject::connect(this, &QWebEngineView::urlChanged, this, &WebView::onUrlChanged);
     connect(this->page(), &QWebEnginePage::linkHovered, this, [=] (const QString& url) {
         m_linkHovered = url;
+    });
+
+    /* In Qt 5.12, the zoom factor is not correctly passed after a fulltext search
+     * Bug Report: https://bugreports.qt.io/browse/QTBUG-51851
+     * This rezooms the page to its correct zoom (default/by ZIM ID) after loading is finished.
+     * If the page is search results, we put the default zoom factor
+     * If in Qt 6.x, the bug is fixed this code can be removed.
+     */
+    connect(this, &QWebEngineView::loadFinished, this, [=] (bool ok) {
+        if (ok) {
+            applyCorrectZoomFactor();
+        }
     });
 }
 
@@ -125,7 +160,7 @@ QWebEngineView* WebView::createWindow(QWebEnginePage::WebWindowType type)
 }
 
 void WebView::onUrlChanged(const QUrl& url) {
-    auto zimId = url.host().split('.')[0];
+    auto zimId = getZimIdFromUrl(url);
     if (m_currentZimId == zimId ) {
         return;
     }
--- a/src/webview.h
+++ b/src/webview.h
@@ -69,6 +69,7 @@ private slots:
 
 private:
     void addHistoryItemAction(QMenu *menu, const QWebEngineHistoryItem &item, int n) const;
+    void applyCorrectZoomFactor();
 };
 
 #endif // WEBVIEW_H


juuz0:
I couldn't get a good reason on why that was happening (it didnt happen, if the search bar article was opened in a different tab) but this looks like the culprit, known bug :
https://bugreports.qt.io/browse/QTBUG-51992

kelson42 (changes requested):
The search result page should have the zoom from the "default zoom" settings

kelson42:
It seems we can not avoid this zoom changing problem because of https://bugreports.qt.io/browse/QTBUG-51851

kelson42 (approved):
Beside the rezooming bug (upstream) after clicking on a search result, it looks good. @mgautierfr Please make the code review.

kelson42:
@juuz0 Would you be able please to make a comment in the code with a link to the upstream ticket to explain why you rezzoom after having loaded the content

juuz0:
@kelson42 Added, rebased

kelson42:
@veloman-yunkan Changing reviewe in an attemos to move forward with this PR :)

src/webview.cpp
```diff
@@ -49,6 +49,25 @@ WebView::WebView(QWidget *parent)
...
+    connect(this, &QWebEngineView::loadFinished, this, [=] (bool ok) {
+        if (ok) {
+            auto url = this->url();
+            auto settingsManager = KiwixApp::instance()->getSettingsManager();
+            qreal zoomFactor;
+            QUrlQuery query(url);
+            if (query.hasQueryItem("pattern")) {
+               zoomFactor = settingsManager->getZoomFactor();
+            } else {
+                auto zimId = url.host().split('.')[0];
```
veloman-yunkan, line 65: I feel suspicious about a ZIM id obtained in such a way. I see that it is a copy from `WebView::onUrlChanged()`. I think that we have a bug there. For example, the URL `http://library.kiwix.org/phet_hy_2021-08` will have `library` extracted from it as `zimId` whereas my impression was that we are interested in the `phet_hy_2021-08` part.

@kelson42 Will you please confirm that my understanding is correct?

veloman-yunkan (changes requested):

src/webview.cpp
mgautierfr: On kiwix-desktop, the url format is `zimid.zim/article/path`.
This is somehow a stable fact as the custom url handler is installed to respond to all `zim` tld (Other request are not catch, and so we can transfer to native browser for external url)
Extracting the zimId from the url could (should) be move to a dedicated function, but it is a correct way to obtain the zimId (In the customSchemeHandler, we simply remove the 4 last chars)

src/webview.cpp
```diff
@@ -49,6 +49,25 @@ WebView::WebView(QWidget *parent)
...
+    /* This rezooms the page to its correct zoom(default/by ZIM ID) after loading is finished.
+    If the page is result of fulltext search, use the default zoom
+    Needed because of https://bugreports.qt.io/browse/QTBUG-51851 */
+    connect(this, &QWebEngineView::loadFinished, this, [=] (bool ok) {
+        if (ok) {
+            auto url = this->url();
+            auto settingsManager = KiwixApp::instance()->getSettingsManager();
+            qreal zoomFactor;
+            QUrlQuery query(url);
+            if (query.hasQueryItem("pattern")) {
```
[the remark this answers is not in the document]
juuz0: Probably, my reasoning was in `suggestionlistworker.cpp`, it was always putting pattern query item

kelson42:
@juuz0 Whats up with this PR?

src/webview.cpp
```diff
@@ -40,6 +40,19 @@ void WebViewForwardMenu::showEvent(QShowEvent *)
...
 
+void WebView::applyCorrectZoomFactor() {
+    auto url = this->url();
+    auto settingsManager = KiwixApp::instance()->getSettingsManager();
+    qreal zoomFactor;
+    QUrlQuery query(url);
+    if (query.hasQueryItem("pattern")) {
+        zoomFactor = settingsManager->getZoomFactor();
+    } else {
+        auto zimId = url.host().split('.')[0];
```
veloman-yunkan, line 51: Please follow @mgautierfr's hint in his [earlier comment](https://github.com/kiwix/kiwix-desktop/pull/774#discussion_r803707578)

> Extracting the zimId from the url could (should) be move to a dedicated function

src/webview.cpp
```diff
@@ -40,6 +40,19 @@ void WebViewForwardMenu::showEvent(QShowEvent *)
     setGeometry(geo);
 }
 
+void WebView::applyCorrectZoomFactor() {
+    auto url = this->url();
+    auto settingsManager = KiwixApp::instance()->getSettingsManager();
+    qreal zoomFactor;
+    QUrlQuery query(url);
+    if (query.hasQueryItem("pattern")) {
```
veloman-yunkan, line 48: The intent of this check will be more explicit (and the code more readable) if you encapsulate it into a helper function `WebView::isSearchResultsView()` or introduce an auxiliary variable in this function:

```
const bool isSearchResultsView = QUrlQuery(url).hasQueryItem("pattern");
if ( isSearchResultsView ) {
    ...
```

Besides, I still think that relying only on the presence of a parameter named  `pattern` in the URL query string is a weak criterion that can easily break if we happen to introduce another type of a URL with a similar parameter. Probably it makes sense to also check that the host component of the URL ends with ".search".

veloman-yunkan (changes requested):

src/webview.cpp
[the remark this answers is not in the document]
kelson42: @juuz0 Should we close this discussion? What is the status?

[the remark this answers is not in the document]
juuz0: @kelson42 Fixed following the discussion in https://github.com/kiwix/kiwix-desktop/pull/774#discussion_r814166559

veloman-yunkan (approved):

PR #368, opened 2020-01-31
Fix menu in fullscreen on Windows

jetownfeve21:
fix #305

PR #783, opened 2022-02-07
Fix contentmanager/mainWindow initialisation on windows.

### LINKED ISSUE #775: [REGRESSION] Searching for "All files" crashes application on Windows
On Windows 10, running Kiwix-desktop 2.1.0 

Clicking on All files in the library causes the application to close unexpectedly.

mgautierfr:
- The (pretty old) PR https://github.com/kiwix/kiwix-desktop/pull/368 set the full screen window border on windows.
But it has a test if the native window is null and return if yes. This is somehow buggy as the native window exists only if the widget has been shown and it is not the case.
- The PR https://github.com/kiwix/kiwix-desktop/pull/680 add a history on back/forward buttons but connect the mainWindow signal *after* the previous check. Because of the early return on windows, the signal was never set on Windows. It was already a bug, but not a crash as the bug was simply that the buttons were not updated.
- The PR https://github.com/kiwix/kiwix-desktop/pull/746 cleanup the code and move the `setContentManager` add the end of the previous code. On windows it was never called and as soon as we try to use it (by clicking on "All files" (https://github.com/kiwix/kiwix-desktop/blob/master/src/contentmanagerside.cpp#L16)) kiwix-desktop crash.

This PR "simply" move the configuration of the fullscreen border at the correct position (after the widget is shown) and remove the early exit in the constructor. Every thing is configured correctly.
A build of this PR is available here : http://tmp.kiwix.org/ci/fix_windows_contentmanager/kiwix-desktop_windows_x64_2022-02-07.zip
Fix #775

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -16,6 +16,9 @@
 #include <QPrintDialog>
 #include <thread>
 #include <QMessageBox>
+#ifdef Q_OS_WIN
+#include <QtPlatformHeaders\QWindowsWindowFunctions>
+#endif
 
 ////////////////////////////////////////////////////////////////////////////////
 // KiwixApp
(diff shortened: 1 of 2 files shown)


kelson42:
@mgautierfr Thank you very much for the PR
@Popolechien If you could check the ZIP provided by Matthieu please?
@asashnov @juuz0 Considering the nature of the bug, I would like to merge this PR within 2 days if no problem is found, so we can make a new release.

juuz0:
Works good on my side (also fixes the side filter panel showing up in Settings)

juuz0 (approved):

Popolechien:
Late to the party but works for me. The library, however, did not load up any files but I don't know if that is bug or feature.

kelson42:
It should

juuz0:
Just a hunch but maybe its a language filter issue? Browse by language -> All, if that shows

Popolechien:
Nope. Tested on 2.0.5 and it's not working either (pretty sure it did before). My internet works, got no filter or weird settings.

Issue #594, opened 2021-02-12
Buttons in toolbar drawn incorrectly in right-to-left layout


asashnov:
Found on current 'master' (ca776b46)

### Steps to reproduce:

1. Run Kiwix application with right-to-left interface layout by adding -reverse CLI option
2. See the buttons in toolbar
3. Open any .zim file from library
4. See the buttons in toolbar again

### Actual result:

![Screenshot from 2021-02-12 20-10-29](https://user-images.githubusercontent.com/5078434/107772817-52083880-6d6f-11eb-9631-1b11fefc2f0e.png)

![Screenshot from 2021-02-12 20-10-38](https://user-images.githubusercontent.com/5078434/107772834-57fe1980-6d6f-11eb-8fe1-3d9e78bac266.png)

### Expected result:

Graphical items in toolbar should not overwrite each other, no duplicate drawings.

### Additional info

There is another issue about Back/Forward buttons order: #595

kelson42:
@asashnov and the icons should be on the right!!!

asashnov:
@kelson42 inside tab headers the order of icon and close button also reverted- that's right, the expected behaviour of QTabbar and other programs, for example:

![web_phishing_zoom75](https://user-images.githubusercontent.com/5078434/107801708-023b6880-6d93-11eb-833f-59b18ff942f2.png)

Source: https://help.eset.com/ees/6/ar-EG/index.html?idh_config_antiphish.htm

asashnov:
Actually the example above shows the opposit of what I wanted to say.

Qt manual says (https://doc.qt.io/qt-5/qtquick-positioning-righttoleft.html):
> The general rule of thumb is that content (like photos, videos and maps) is not mirrored, but positioning of the content (like application layouts and the flow of visual elements) is mirrored.
> ...
> Also, there are some special cases you may need to take into account where right-to-left language speakers are used to left-to-right positioning, for example when using number dialers in phones and media play, pause, rewind and forward buttons in music players.

But it says nothing clearly about tabs header.

So I'm confused here. I think we can rely on the default behaviour of QTabbar now, and may `setLayoutDirection(Qt::LeftToRight);` in Tabbar constructor if users ask us about it:

![Screenshot from 2021-02-13 00-47-23](https://user-images.githubusercontent.com/5078434/107803746-9a3a5180-6d95-11eb-8456-6905b78b5f5b.png)

kelson42:
[...]

PR #797, opened 2022-02-20
Fix right-to-left layout in SearchBar

asashnov:
There was a hacky implementation which created QPushButton as a child of QLineEdit, but without using any Layout (actually, QLineEdit has no documented facilities to contain any other widget).
So, somehow in old implementation Qt placed QPushButton in the left side of QLineEdit. QLineEdit itself didn't knew about it, so in CSS left padding 40px; was needed. With '-reverse' CLI option (right-to-left layout) padding 40px remained to be on the left. Also, depending on which text we enter in QLineEdit (English or Arabic) the text can also be started from the left or the right side of text area in QLineEdit, so otherwise we had to detect text layout and update CSS paddings on the fly.

The new implementation is more elegant and uses the documented facility of QLineEdit to contain an QAction. So no hacks with CSS padding is need. Also, because QPushButton was a separate QWidget, clicking on it didn't pass the focus into QLineEdit, but clicking on QAction icon does focusIn. So focus interation code also has to be changed.

Fixes: #594

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -2,37 +2,29 @@
 
 #include <QCompleter>
 #include <QFocusEvent>
+#include <QCursor>
 
 #include "kiwixapp.h"
 #include "suggestionlistworker.h"
 
-SearchButton::SearchButton(QWidget *parent) :
-    QPushButton(parent),
-    m_searchMode(true)
-{
-    setFlat(true);
-    setIcon(QIcon(":/icons/search.svg"));
-    connect(this, &QPushButton::clicked, this, &SearchButton::on_buttonClicked);
-}
 
-void SearchButton::set_searchMode(bool searchMode)
+
+void SearchBar::set_searchMode(bool searchMode)
 {
     m_searchMode = searchMode;
     if (m_searchMode) {
-        setIcon(QIcon(":/icons/search.svg"));
-        setIconSize(QSize(27, 27));
+        m_action->setIcon(QIcon(":/icons/search.svg"));
     } else {
         auto kiwixApp = KiwixApp::instance();
         if (kiwixApp->isCurrentArticleBookmarked()) {
-            setIcon(QIcon(":/icons/reading-list-active.svg"));
+            m_action->setIcon(QIcon(":/icons/reading-list-active.svg"));
         } else {
-            setIcon(QIcon(":/icons/reading-list.svg"));
+            m_action->setIcon(QIcon(":/icons/reading-list.svg"));
         }
-        setIconSize(QSize(25, 25));
     }
 }
 
-void SearchButton::on_buttonClicked()
+void SearchBar::on_action_triggered()
 {
     if (m_searchMode)
         return;
@@ -60,7 +52,7 @@ void SearchButton::on_buttonClicked()
 SearchBar::SearchBar(QWidget *parent) :
     QLineEdit(parent),
     m_completer(&m_completionModel, this),
-    m_button(this)
+    m_action(new QAction(this))
 {
     mp_typingTimer = new QTimer(this);
     mp_typingTimer->setSingleShot(true);
@@ -70,6 +62,11 @@ SearchBar::SearchBar(QWidget *parent) :
     m_completer.setCaseSensitivity(Qt::CaseInsensitive);
     m_completer.setMaxVisibleItems(16);
     setCompleter(&m_completer);
+
+    connect(m_action, &QAction::triggered,
+            this, &SearchBar::on_action_triggered);
+    addAction(m_action, QLineEdit::LeadingPosition);
+    set_searchMode(true);
 
     QFile styleFile(":/css/popup.css");
     styleFile.open(QIODevice::ReadOnly);
@@ -121,30 +118,41 @@ void SearchBar::on_currentTitleChanged(const QString& title)
     } else {
         setText("");
     }
-    m_button.set_searchMode(title.isEmpty());
+    set_searchMode(title.isEmpty());
     m_title = title;
 }
 
 void SearchBar::focusInEvent( QFocusEvent* event)
 {
+    if (event->reason() == Qt::MouseFocusReason) {
+        if (QApplication::widgetAt(QCursor::pos()) != this) {
+            // drop focus, mouse clicked on one of inner widget (m_action)
+            clearFocus();
+            return;
+        }
+    }
+
     setReadOnly(false);
     if (event->reason() == Qt::MouseFocusReason && text() == m_title) {
         clear();
     }
     if (event->reason() == Qt::ActiveWindowFocusReason ||
         event->reason() == Qt::MouseFocusReason) {
+        /* need to connect every time because QLineEdit::focusOutEvent does:
+         * QObject::disconnect(d->control->completer(), 0, this, 0);
+         */
         connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
         this, QOverload<const QModelIndex &>::of(&SearchBar::openCompletion));
     }
     QLineEdit::focusInEvent(event);
-    m_button.set_searchMode(true);
+    set_searchMode(true);
 }
 
 void SearchBar::focusOutEvent(QFocusEvent* event)
 {
     setReadOnly(true);
     if (event->reason() == Qt::MouseFocusReason && text().isEmpty()) {
-        m_button.set_searchMode(false);
+        set_searchMode(false);
         setText(m_title);
     }
     deselect();
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -32,23 +32,14 @@ QToolButton {
 
 SearchBar {
     background-color: white;
-    padding: 2px 2px 2px 40px;
+    padding: 2px;
     max-height: 38px;
-    margin: 2px 5px;
     color: #666;
     font-size: 16px;
     border: 1px solid #ccc;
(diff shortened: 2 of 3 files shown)


asashnov:
Testing notes:

Screenshot from 2022-02-20 16-24-42

In Search mode 'Search' is on the left, but this only because 'Search' is in Enlish:
Screenshot from 2022-02-20 16-33-11

When run with 'ar' (Arabic) LANG the 'Search' is in Arabic, and also right-to-left:

Screenshot from 2022-02-20 16-31-04

veloman-yunkan:
@kelson42 I added you as a reviewer so that you can verify that the change is functionally acceptable. I will review the code after you confirm that it works fine.

kelson42 (changes requested):
@asashnov Thx for your PR, this is a lot better, but IMO there is problem with the alignment of content within the searchbox itself, see my comment at https://github.com/kiwix/kiwix-desktop/issues/594#issuecomment-1050797097

src/searchbar.cpp
```diff
@@ -97,6 +44,52 @@ SearchBar::SearchBar(QWidget *parent) :
...
+
+    m_action->setEnabled(true); // enable toggling bookmark for the current tab
+
+    auto kiwixApp = KiwixApp::instance();
+    if (kiwixApp->isCurrentArticleBookmarked()) {
+        m_action->setIcon(QIcon(":/icons/reading-list-active.svg"));
+    } else {
+        m_action->setIcon(QIcon(":/icons/reading-list.svg"));
+    }
 }
```
mgautierfr, line 93: I wonder if we could use two different action and set the action of the QLineEdit when we switch mode instead of reconfiguring the only one action.

src/searchbar.cpp
```diff
@@ -97,6 +44,52 @@ SearchBar::SearchBar(QWidget *parent) :
...
+     * showing of text cursor.
+     *
+     * This workaround was found in QLineEdit sources:
+     * d->control->setBlinkingCursorEnabled(true)
+     * is called only from two places:
+     *  - from focusInEvent() - but we have to ignore it: at the time the event arrived
+     *    it could be action toggling next;
+     *  - from event() if (e->type() == QEvent::Show) - we use it here:
+     */
+    QApplication::postEvent(this, new QShowEvent());
```
mgautierfr, line 77: This look like a issue for Qt
When the blinking of the cursor is disable ?
Should it be enable "itself" ? If not QLineEdit need a new method.

We will still need the workaround here (if we have properly used Qt) but it would be nice to create a issue on Qt side.

src/searchbar.cpp
```diff
@@ -113,41 +106,100 @@ void SearchBar::clearSuggestions()
...
-        connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
-        this, QOverload<const QModelIndex &>::of(&SearchBar::openCompletion));
+
+    setBookmarkMode(); // already in bookmark mode; called just to set the right (on/off) icon
+    library->save();
+}
+
+void SearchBar::on_delayed_search_activation()
+{
+    if (m_action_toggled) {
```
mgautierfr, line 177: When this is set to true ?

mgautierfr (changes requested):

src/searchbar.cpp
asashnov: 2 different actions would be indeed much simpler, but I tried to keep only one icon like it done in Apple Safari

src/searchbar.cpp
asashnov: I don't think it's Qt bug. We just want our own custom behaviour.

It somes from Apple Safari:

When there is no opened page, we see 'Search' icon in the address bar:

Screenshot from 2022-03-05 01-02-41

When some page is open, we see 'Add to reading list' button:

Screenshot from 2022-03-05 01-04-17

But once we start typing, it turns into 'Search' icon again:

Screenshot from 2022-03-05 01-10-57

So this is why we want only one action, not two.

src/searchbar.cpp
asashnov: Good catch. I have forgotten it somehow, returned back.

asashnov:
@kelson42 

> @asashnov Thx for your PR, this is a lot better, but IMO there is problem with the alignment of content within the searchbox itself, see my comment at https://github.com/kiwix/kiwix-desktop/issues/594#issuecomment-1050797097 :
>
> The context of the search box:
> * The text should be aligned right (like the suggestions)
> * Any magnify logo should be on the right as well

When run with LANG=ar   (Arabic, no -reverse CLI option is needed) content of search box is already as described:

Screenshot from 2022-03-05 01-36-45

kelson42 (changes requested):
From your latest comment I actually remark that I already had all the information in your earlier comment. Sorry you had to repeat yourself. Your explanation seems indeed appropriate and this is probably OK even if I don't understand why the `-reverse` option does not behave like `-lang=ar` option... but I guess there is a logical explanation.

That said, this PR introduces a serious regression for me. If I type quickly in the searchbox the search box is emptied without good reason. We need to be carefuly about this PR, I see a lot of code changes and it has already taken quite a few PR in the past to have this searchbox working correctly.

asashnov:
`-reverse` changes the layout, but keeps menu texts in English. So with just `-reverse` option we cannot see fully mirrored controls because strings in English continue to be aligned left (in context menus for example). This is why `-lang=ar` is better for testing this PR

I've seen a lot of work in past on this. Indeed, it's a trick to make QLineEdit to change the icon inside it accordingly. I tried to keep my changes small, but the over all code readability decreased a lot, so I decided to make it more easy to read in future. I ordered the code from top to bottom as it executed in runtime, and commented.

@kelson42 the issue with quick typing- did you test on MacOS?

kelson42:
@asashnov No, I tested on Ubuntu, this code repo does not run on macOS.

mgautierfr (changes requested):
This PR break suggestions:

- Open a zim file
- Start searching for a term
- Select a suggestion article (either by clicking on it with mouse or select with keyboard)
- A full text search on the term is launch (on master, the suggested article is opened)


>  I tried to keep my changes small, but the over all code readability decreased a lot, so I decided to make it more easy to read in future. I ordered the code from top to bottom as it executed in runtime, and commented.

Reordering the methods is ok. But it would be easier (for reviewer) to put the code move in a separated commit. Here we have to compare different code in different place (and it is not so easy 🙂) The same way, it is better to have new commits for the new changes you've made. If you want to have all the changes in the same commit at the end, you can name the new commit with `fixup! <title of the commit to merge with>`. But it will help the reviewing process as I don't have to review all the changes but only what change since the last review.

kelson42:
@asashnov Any news?

src/searchbar.cpp
```diff
@@ -113,41 +106,109 @@ void SearchBar::clearSuggestions()
...
+        return;
     }
-    QLineEdit::focusInEvent(event);
-    m_button.set_searchMode(true);
+
+    // bookmark wasn't toggled, enter into search mode
+    setSearchMode();
+
+    connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
+            this, QOverload<const QModelIndex &>::of(&SearchBar::openCompletion));
```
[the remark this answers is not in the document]
asashnov: Seems yes. When I move this connection to constructor, it stops working.

asashnov:
This is why we need to connect the signal every time:

From Qt sources, qlineedit.cpp:
```c
void QLineEdit::focusInEvent(QFocusEvent *e)
...
    if (d->control->completer()) {
        d->control->completer()->setWidget(this);
        QObject::connect(d->control->completer(), SIGNAL(activated(QString)),
                         this, SLOT(setText(QString)));
...

void QLineEdit::focusOutEvent(QFocusEvent *e)
... 
    if (d->control->completer()) {
        QObject::disconnect(d->control->completer(), 0, this, 0);
    }
```

asashnov:
This is a completely new fix, not based on the previous changes I made.

This change is trying to be minimal. And it use reliable way of distinguishing between a click on action icon, or inside the text area inside QLineEdit by calling QApplication::widgetAt()  (it returns the widget for action when clicked on action button).

So this PR should not have any side effects.

Also, added a comment why we need to connect every time QCompleter::activated() signal.

kelson42:
@asashnov Thx for the new attempt. Will check within a few days.

kelson42:
@asashnov From a testing POV it looks good to me!
@mgautierfr Could you please make the technical code review?

kelson42:
@mgautierfr Can you review again this PR?

asashnov:
rebased on current master

kelson42:
@mgautierfr please review

asashnov:
rebased on the current master.

@mgautierfr Strange indeed that in your experience LANG=ar didn't have the effect. I specified LANG value in QtCreator's program run environment:

Screenshot from 2022-06-15 01-40-53

In the current version of code the placement of user interface items is totaly up to Qt, so once Qt configured correctly it will place UI items in most appropriate way for the current language.

When run as specified above, the placement of controls is right-to-left:

Screenshot from 2022-06-15 01-44-06

kelson42:
@mgautierfr Don't forget to review pleade

kelson42:
@mgautierfr Still waiting your review. If only a few details have to be fixed, please carry one yourself. This PR is really old meanwhile.

kelson42:
This PR is not needed anymore

PR #791, opened 2022-02-14
Correctly set the "acceptTags" when we filter by tags.

mgautierfr:
In case of "other" filter, we were setting the filter with
`filter.rejectTags(tags)` and `filter.acceptTags(tags)` in the same time.
Which obviously return 0 results.

We must call `filter.acceptTags(tags)` only if we use a specific category,
not all the time (including with "other" category).

Fix #786

PR #798, opened 2022-02-23
Fix content-type filtering

### LINKED ISSUE #792: Content Type filters are not working
image

mgautierfr:
The commit cb63bb2 was buggy and deactivate filtering by contentType.
We now handle two different lists to store tags we want and tags we don't.

Related to #791 
Fix #792

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -392,30 +392,30 @@ void ContentManager::setSearch(const QString &search)
 QStringList ContentManager::getBookIds()
 {
     kiwix::Filter filter;
-    std::vector<std::string> tags;
+    std::vector<std::string> acceptTags, rejectTags;
     if (m_categoryFilter != "all" && m_categoryFilter != "other") {
-        tags.push_back("_category:"+m_categoryFilter.toStdString());
-        filter.acceptTags(tags);
+        acceptTags.push_back("_category:"+m_categoryFilter.toStdString());
     }
     if (m_categoryFilter == "other") {
         for (auto& category: S_CATEGORIES) {
             if (category.first != "other" && category.first != "all") {
-                tags.push_back("_category:"+category.first.toStdString());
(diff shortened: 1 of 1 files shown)


juuz0 (approved):

src/contentmanager.cpp
```diff
@@ -392,30 +392,30 @@ void ContentManager::setSearch(const QString &search)
...
     for (auto &contentTypeFilter : m_contentTypeFilters) {
         auto state = contentTypeFilter->checkState();
         auto filter = contentTypeFilter->getName();
         if (state == Qt::PartiallyChecked) {
-            tags.push_back("_" + filter.toStdString() +":yes");
+            acceptTags.push_back("_" + filter.toStdString() +":yes");
         } else if (state == Qt::Checked) {
-            tags.push_back("_" + filter.toStdString() +":no");
+            acceptTags.push_back("_" + filter.toStdString() +":no");
         }
```
veloman-yunkan, line 414: I wonder why a partially checked state is translated to a "_contenttype:yes" tag and a fully checked state to a "_contenttype:no" tag

mgautierfr: The button is a 3 states button (`yes`, `no` and `no filtering`).
The 3 states available from Qt are `Checked`, `PartiallyChecked` and `Unchecked`.
We set the button ui here : https://github.com/kiwix/kiwix-desktop/blob/master/src/contenttypefilter.cpp#L8-L12

kelson42 (approved):
@mgautierfr Thank you! LGTM

src/contentmanager.cpp
veloman-yunkan: Such mapping of various checked states to filter states is a little confusing. Anyway, your explanation allows to conclude that everything is correct here.

veloman-yunkan (approved):

PR #800, opened 2022-02-25
Fix for monitor directory on Windows

### LINKED ISSUE #793: Latest kiwix-desktop crashes on using monitor directory
Steps to reproduce:

- Choose a monitor directory from Settings
- kiwix-desktop will fail instantly/after some use.

Windows logs show it's a heap corruption

juuz0:
Fixes #793

1. The books weren't changing (adding/removing) because of a mess in creating path. Specifically, I used QDir::separator(), while the path was unix based (on windows too), using QDir::separator() made the path like: `dir/subdir/sub-subdir\sub-sub-subdir`

2. The crash only happened when monitor directory was set in the same directory as a library.xml (download directory), this was because it was calling save() at end  (as rightly pointed by @veloman-yunkan in the issue), this would send a directoryChanged() signal whenever library.xml modified. Going in an infinite loop. Now, we don't call save(), it is already called in the library destructor() so not needed. Doesn't compromise any functionality in monitoring.

3. Whenever download or monitor directory was changed, we used UNIX styled paths, we now use paths with native separators in them

A build for testing is available at: http://tmp.kiwix.org/ci/juuz0_windows_build/2022-02-25/kiwix-desktop_windows_x64_2022-02-25.zip

### DIFF
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -150,7 +150,7 @@ void Library::loadMonitorDir(QString monitorDir)
     const QDir dir(monitorDir);
     QStringList newDirEntries = dir.entryList({"*.zim"});
     for (auto &str : newDirEntries) {
-        str = monitorDir + QDir::separator() + str;
+        str = QDir::toNativeSeparators(monitorDir + "/" + str);
     }
     QSet<QString> newDir = QSet<QString>::fromList(newDirEntries);
     QStringList oldDirEntries = m_monitorDirZims;
--- a/src/library.cpp
+++ b/src/library.cpp
@@ -167,7 +167,6 @@ void Library::loadMonitorDir(QString monitorDir)
         removeBookFromLibraryById(QString::fromStdString(m_library.getBookByPath(bookPath.toStdString()).getId()));
     }
     emit(booksChanged());
-    save();
 }
 
 void Library::asyncLoadMonitorDir(QString dir)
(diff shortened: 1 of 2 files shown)


kelson42:
@juuz0 Congrats for the fix, I will give a try.

veloman-yunkan (approved):

mgautierfr:
> The PR message is bigger than the fixes! :P

Welcome to the real developers' world. Spending hours searching for a bug and fixing it with one line.

mgautierfr (commented):
In the setting the download directory is written with `\` and the directory to monitor is written with `/`. It would be nice to have the same way to writing (using the native norme `\` on windows, `/` elsewhere)

But I will not block for this as the PR is fixing a important bug.

juuz0:
> In the setting the download directory is written with \ and the directory to monitor is written with /. It would be nice to have the same way to writing (using the native norme \ on windows, / elsewhere)
But I will not block for this as the PR is fixing a important bug.


f4df4022b6e50b12bb627aba83c50ff9610537e8 should have fixed that too, should do at init() also then. Doing

kelson42:
> In the setting the download directory is written with `\` and the directory to monitor is written with `/`. It would be nice to have the same way to writing (using the native norme `\` on windows, `/` elsewhere)

Really important to me, because currently really misleading for users.

kelson42 (changes requested):
Works fine now for me, but please fix the way the monitoring path is displayed inthe settings.

juuz0:
It is already fixed actually, if you change monitor directory, you will see `\` in path.
But the already saved path (from previous kiwix-desktop versions will have `/`).
So can you please confirm if you see corrrect path on changing it?

kelson42:
@juuz0 I confirm, sorry about them misunderstanding

kelson42 (approved):

Issue #802, opened 2022-02-25
Not loading online article list


RatorLP:
tried it on 2.0, 2.1.0 and today's nightly, as well as two different computers, kiwix desktop doesn't seem to load the online library. Running Windows 10 20H2 on one and 21H2 on the other machine, even tried it with a completely disabled firewall over 2 different networks from different ISPs, just for the funzies, but it didn't work either. I'm kinda lost here, don't have any idea where this might be coming from. I assumed some server downtime or sth, but the web version seems to work fine, so i'm clueless.

When opening "All files"  (or "Alle Dateien" in german) the blue spinner shows, after a while it disappears, but nothing else happens.
![kiwix1](https://user-images.githubusercontent.com/46613254/155806670-85f518da-e024-44b6-8cb8-781b0bd52e1b.png)
![kiwix2](https://user-images.githubusercontent.com/46613254/155806682-9db77f83-8525-48ad-85ed-3753b7b4fb41.png)

RatorLP:
maybe i'm just missing something in the process, in that case this is to be treated as an UI intinuitivity issue

RatorLP:
> @RatorLP Try with latest nightky build http://download.kiwix.org/nightly/2022-02-25/kiwix-desktop_windows_x64_2022-02-25.zip. Shoukd work.

That's exactly the build I tried

kelson42:
@RatorLP Works for me now with my computer (Linux). In the console window you can see the API URL which is loaded. If you load it in the your browser, does it load properly?

RatorLP:
@kelson42 about the update, yeah the 20H2 machine is a bit behind, but 21H2 ist the most recent version of win10, there is nothing newer. 

After a couple of reboots it loads the lib on the 20H2 PC, but does this:

https://user-images.githubusercontent.com/46613254/155845663-6d4363fd-7009-499f-9b70-d458c55d1e72.mp4

juuz0:
@RatorLP can you show settings of kiwix desktop?
Top right, 3 dots -> Settings

kelson42:
@RatorLP Your latest screenshot is wrong because the URL you have given is wrong (and not the one Kiwix Desktop requested). I should be http://library.kiwix.org/catalog/search?lang=deu&count=-1, not http://library.kiwix.org/catalog/search?lang=deu?count=-1

juuz0:
[...]

PR #803, opened 2022-02-26
Move asyncLoadMonitorDir in ContentManager

juuz0:
This gives it the ability to use m_local and only make changes when local library is in use.
Not causing repeated refreshes while downloading - if the monitor directory is set same as download directory

The problem this fixes is:
- Put monitor directory same as download directory.
- Go to "All files"
- Download a large enough file
- There should be many re-freshes

Because, while downloading, file is changing a lot and causes file watcher to detect causing many calls of loadMonitorDir() leading to emitBooksChanged()

Solution:

Don't do any monitoring while we are using remote library, the "Download" function already adds books, shouldn't interfere in that

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -35,6 +35,9 @@ void ContentManager::setLocal(bool local) {
         return;
     }
     m_local = local;
+    if (m_local) {
+        asyncLoadMonitorDir(KiwixApp::instance()->getSettingsManager()->getMonitorDir());
+    }
     emit(filterParamsChanged());
 }
 
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -11,6 +11,7 @@
 #include <QDir>
 #include <QStorageInfo>
 #include <QMessageBox>
+#include <QtConcurrent/QtConcurrentRun>
 
 ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader, QObject *parent)
     : QObject(parent),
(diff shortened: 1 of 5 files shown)


juuz0:
@kelson42 No, I can't reproduce #802 with http://download.kiwix.org/nightly/2022-02-26/kiwix-desktop_windows_x64_2022-02-26.zip. I should mention that the changes from https://github.com/kiwix/kiwix-desktop/pull/800  are introduced in nightly, 2022-02-26 (which OP hasn't tried), they could've fixed it.

This PR fixes an OS independent issue :)

kelson42 (changes requested):
I believe to not understand all the rationals behind this PR. It seems to be an optimization of the code which is in `master` to save the software many refresh efforts (when the local library is not displayed).

On the principle, I can only praise such an effort and I have tested it. Unfortunately this bring at least one regression:
* Monitore the Download directory and put a `toto.zim` file
* It appears in the local library properly
* Go to online library
* Rename `toto.zim` to `toto.zim.hidden`
* Got to local library. The `toto.zim` entry is still there and it should not anymore (wchih is the case with `master`).

With the little I understand, the principle of the this optimisation seems somehow weak. Maybe a clearer split should be made between (1) update the internal library (2) update the visual part.

juuz0:
> Got to local library. The toto.zim entry is still there and it should not anymore (wchih is the case with master).

Fixed

kelson42 (changes requested):
Seems now totally broken to me: does not detect new files and does not detect removed files

juuz0:
@kelson42 Superseded by #811

PR #816, opened 2022-03-10
Add check for fileName in eraseBookFilesFromComputer

### LINKED ISSUE #809: Stack Overflow downloading error, all files in folder deleted
I ran into a very annoying error that deleted all my files in download directory.

Steps to reproduce:

1. Download the appimage https://download.kiwix.org/release/kiwix-desktop/kiwix-desktop_x86_64.appimage
2. Run it.
3. Switch to all files.
4. Change the language to English.
5. Sort by size (descending).
6. On the "Stack Overflow" line, click "Download".
7. In the window with the error, click "OK".

kiwix

All files in the same directory as kiwix-desktop_x86_64_2.0.5.appimage will be removed. Appimage too. Directories are not deleted.

Console output:
```
$ ./kiwix-desktop_x86_64_2.1.0-1.appimage 
add widget
"2fe8fd32b5ec077f"
"a9ad212078bf2828"
Downloading "https://library.kiwix.org:443/catalog/search?lang=eng&count=0"
```

Expected behavior:

1. Stack Overflow must be loaded.
2. Files should not be deleted.

```
$ uname -a: 
Linux geek-pc 5.13.0-30-generic #33~20.04.1-Ubuntu SMP Mon Feb 7 14:25:10 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

juuz0:
Fix #809

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -258,9 +258,12 @@ QString ContentManager::downloadBook(const QString &id)
     return QString::fromStdString(download->getDid());
 }
 
-void ContentManager::eraseBookFilesFromComputer(const QString dirPath, const QString filename)
+void ContentManager::eraseBookFilesFromComputer(const QString dirPath, const QString fileName)
 {
-    QDir dir(dirPath, filename);
+    if (fileName == "*") {
+        return;
+    }
+    QDir dir(dirPath, fileName);
     for(const QString& file: dir.entryList()) {
         dir.remove(file);
     }
@@ -272,8 +275,8 @@ void ContentManager::eraseBook(const QString& id)
     tabBar->closeTabsByZimId(id);
     kiwix::Book book = mp_library->getBookById(id);
     QString dirPath = QString::fromStdString(kiwix::removeLastPathElement(book.getPath()));
-    QString filename = QString::fromStdString(kiwix::getLastPathElement(book.getPath())) + "*";
-    eraseBookFilesFromComputer(dirPath, filename);
+    QString fileName = QString::fromStdString(kiwix::getLastPathElement(book.getPath())) + "*";
+    eraseBookFilesFromComputer(dirPath, fileName);
     mp_library->removeBookFromLibraryById(id);
     mp_library->save();
     emit mp_library->bookmarksChanged();


juuz0:
This change adds the check which should fix the problem reported but I don't understand why is there a need to add `*` in `ContentManager::eraseBook`, this line:
`QString fileName = QString::fromStdString(kiwix::getLastPathElement(book.getPath())) + "*";`

kelson42:
@juuz0 Thank you for the super quick patch. I suspect this is needed to delete all files in case of a splitted ZIM file.

kelson42:
@juuz0 The bug to me is more that under certain circunstances `filename` is empty string! Why this is the case?

juuz0:
> The bug to me is more that under certain circunstances filename is empty string! Why this is the case?

Not a clear idea but as you mentioned probably because the download has failed and this book is not added in library then?

> For the moment, I believe it would be better to just not call eraseBookFilesFromComputer in case of book.getPath() returning nothing.

I considered this but `eraseBookFilesFromComputer` is also used in `cancelBook` so to have a safeguard at both places, I went with this.
As an alternative, I feel just not calling `ContentManager::eraseBook` on a download fail should work without any other changes since for failed downloads, `book.getPath()` (assumption) will always return an empty string?

kelson42:
> > The bug to me is more that under certain circunstances filename is empty string! Why this is the case?
> 
> Not a clear idea but as you mentioned probably because the download has failed and this book is not added in library then?

We can leave it like it is then, if this seems legit to call
 
> > For the moment, I believe it would be better to just not call eraseBookFilesFromComputer in case of book.getPath() returning nothing.
> 
> I considered this but `eraseBookFilesFromComputer` is also used in `cancelBook` so to have a safeguard at both places, I went with this. As an alternative, I feel just not calling `ContentManager::eraseBook` on a download fail should work without any other changes because for failed downloads, `book.getPath()` (assumption) will always return an empty string? Changing this in `_contentManager.js`:
> 
> ```
> else if (d.status == "error") {
>       clearInterval(downloadUpdaters[id]);
>       Vue.delete(app.downloads, id);
>       alert("Error: download failed.");
>       contentManager.eraseBook(id); <------ remove this
>       return;
>     }
> ```

There is quite a few things to say here IMO:
* First your download can have error after having started to save data on the filesystem, in that case you have to clean it.
* `cancelBook` and `eraseBook` are in a way quite similar and share similarities. Not sure we have the best code factoring here...
* Neither the best method naming...
* `eraseBookFilesFromComputer()` is a tool, you should let him do his job... if something does not use it right it not its fault. The safeguard is anyway pretty weak, think about `*.*` for example.
* If the download has an error (so download is not completed), I don't understand anyway why `eraseBook()` is called and not `cancelBook()`!

Actually quite a few things are NOK to me. If all my remarks bother you ;) then just leave it, will try to propose an other PR later this week.

WildLynxDev:
I think doing file deletions with * is a bad thing. 
There is so many ways it could go rogue, not only with empty book name. 
Any software that is not a file manager should remove own files only by exact name. There were a lot of stories about uninstallers clearing disk C because of some directory walking algorithm gone nuts.

src/contentmanager.cpp
```diff
@@ -272,8 +275,8 @@ void ContentManager::eraseBook(const QString& id)
     tabBar->closeTabsByZimId(id);
     kiwix::Book book = mp_library->getBookById(id);
     QString dirPath = QString::fromStdString(kiwix::removeLastPathElement(book.getPath()));
-    QString filename = QString::fromStdString(kiwix::getLastPathElement(book.getPath())) + "*";
-    eraseBookFilesFromComputer(dirPath, filename);
```
mgautierfr, line 276: It is probably to `eraseBookFilesFromComputer` to split the bookPath and erase book properly.

The signature of the method should probably be:
`bool eraseBookFilesFromComputer(const Book& book)`

mgautierfr:
> book.getPath() (assumption) will always return an empty string?

The path of the book is set once the download is finished (https://github.com/kiwix/kiwix-desktop/blob/master/src/contentmanager.cpp#L165). Before that, the book is a copy of what is the remote library and, obviously, it hasn't a (local) path.
So yes, for a failed or canceled download, `getPath()` is always empty.

We probably have to:
- The the path of the book as soon as we start the download.
- The the path valid (https://github.com/kiwix/libkiwix/blob/master/include/book.h#L111) when the download succeed.
- Use the signature I propose for `eraseBookFilesFromComputer` to make the method take a `Book` reference.
- Add a check in the method to be sur that `book.getPath()` is not empty.
- Do not use `*` globing. We have to remove the path of the book. If there is something else to remove (meta4 file ?) we know it and we must try to remove it only.
- For split zim file, we may try to compose what would be the filenames (`.zimaa`, `.zimab`, ...) and search for them. But I'm not sure about this. kiwix-desktop don't use split files, so if the zim file is split, it is a zim file added by the user, and I'm in favor of not remove it (https://github.com/kiwix/kiwix-desktop/issues/479#issuecomment-654317770)

mgautierfr:
This fixed issue is really important.
I prefer a proper solution to this PR, but it does the work.
I will merge this PR as it is and make a 2.2.1 release.
Please @juuz0, continue to work on the issue to have a better fix.

juuz0:
:+1: I will follow all inputs above

PR #830, opened 2022-04-11
Do not crash if we are searching on a zim file without fulltext index.

### LINKED ISSUE #826: when search using fullindex,application crashed.
I just encounter this crash when work on another project.

use the zim files provided here (download dictionary1.1.0.zim)
https://sourceforge.net/projects/telugu-dictionary-zim-file/files/

trigger a fulltext search 
image

the application will just exit without any warning.

kiwix desktop version :2.2.1

mgautierfr:
Fix #826


juuz0 (approved):
LGTM

PR #831, opened 2022-04-12
[WIP] Replace deprecated functions

### LINKED ISSUE #805: Stop using deprecated libkiwix methods
libkiwix `10.x.y` have deprecated many methods and function which are still in used in Kiwix Desktop. This has two consequences:
* Many warning at Kiwix Desktop compilation time
* Stop us to removed these old/deprecated primitives in future libkiwix releases.

Would be beneficial to remove the usage of these deprecated methods and use modern/new API of libkiwix.

The effort has already been made in Kiwix Tools, see https://github.com/kiwix/kiwix-tools/pull/535 and this PR might help to know what to do with Kiwix Desktop.

juuz0:
This replaces the deprecated Reader and Favicon functions from libkiwix

kiwix::Reader functions are converted to use zim::Archive
Favicon functions are converted to use Illustration
Fix #805


src/library.cpp
```diff
@@ -67,11 +67,27 @@ std::shared_ptr<kiwix::Reader> Library::getReader(const QString &zimId)
...
+        if (archive) {
+            if (!searcher) {
+                searcher = std::make_shared<zim::Searcher>(*archive);
+            } else {
+                searcher->addArchive(*archive);
+            }
+        }
     } catch(std::out_of_range& e) {
         return nullptr;
     }
```
mgautierfr, line 93: If we don't catch exception, this can be simply rewritten with :
```
auto archive = m_library.getArchiveById(zimId);
return std::make_shared<zim::Searcher>(*archive);
```
and when kiwix/libkiwix#729 will be merged (and released), we can simply use `m_library.getSearcherById(zimId)`.

mgautierfr:
> Removing the reader in suggestionlistworker.cpp should be the last thing to do, to complete this PR. But I'm stuck on how to get suggestions with an archive (wouldn't exposing getSuggestions() from libkiwix internalServer just make it easier? Points to https://github.com/kiwix/libkiwix/issues/740) so would appreciate help on that.

As say the small comment in https://github.com/kiwix/libkiwix/blob/master/src/server/internalServer.cpp#L408-L440, searching in the title list using "normalized" case should be done in libzim.
But it is not a easy task (this is also why we introduced xapian tittle index)

I'm partisan to only use `SuggestionSearcher` in kiwix-desktop. It already fallback to title search (but without case normalization).
For old zim file (before 2019-08) it would be a regression as we would not do case normalization title search (and so suggestion may be less accurate).
If this is really a issue, we should do it where it must be done, in libzim (but I'm not sure it worth the work)

juuz0:
> For old zim file (before 2019-08) it would be a regression as we would not do case normalization title search (and so suggestion may be less accurate).

For a zim like https://sourceforge.net/projects/telugu-dictionary-zim-file/files/ (from 2016), this will lead to no suggestions, won't that be a major regression for old files?

kelson42:
From a user perspective it looks good. I don't see DEPRECATED warnings during the compilation too. @mgautierfr Please have a look to the code.

src/suggestionlistworker.cpp
```diff
@@ -47,7 +70,7 @@ void SuggestionListWorker::run()
             suggestionList.append(m_text + " (" + gt("fulltext-search") + ")");
             urlList.append(url);
         }
-    } else {
+    } catch (std::out_of_range& e) {
```
mgautierfr, line 73: The `else` was here for the second `if (reader)`, and the second if was about fulltext search.
Now, the `catch` is about the only one `try` so the following comment is a bit odd.
It would be better to have:
```
// Impossible to find the requested archive (bug ?)
// We could propose a suggestion to do multi-zim search with:
// url.setHost("library.search");
// but we don't have a correct UI to select on which zim search, how to display results, ...
// So do nothing for now

File: src/urlschemehandler.cpp [hunk 3]
```diff
@@ -8,11 +8,41 @@
...
+  auto final_entry = entry;
+  while (final_entry.isRedirect() && loopCounter--) {
+    final_entry = final_entry.getRedirectEntry();
+  }
+  // Prevent infinite loops.
+  if (final_entry.isRedirect()) {
+    throw zim::EntryNotFound("bad entry");
+  }
+  return final_entry;
+}
```
mgautierfr, line 44: `zim::Entry::getItem(true)` follows the redirection (until we have a item) and then return the associated item. We could use it directly instead this intermediate function.

src/webview.cpp
```diff
@@ -167,12 +167,16 @@ void WebView::onUrlChanged(const QUrl& url) {
...
-    if (!reader) {
+    std::shared_ptr<zim::Archive> archive;
+    try {
+        archive = app->getLibrary()->getArchive(m_currentZimId);
+    } catch (std::out_of_range& e) {
         return;
     }
     std::string favicon, _mimetype;
-    reader->getFavicon(favicon, _mimetype);
+    auto item = archive->getIllustrationItem(48);
```
mgautierfr, line 177: We must handle case where archive has not 48x48 illustration

mgautierfr (changes requested):

src/readinglistbar.cpp
```diff
@@ -31,12 +31,15 @@ void ReadingListBar::setupList()
...
+        std::shared_ptr<zim::Archive> archive;
+        try {
+            archive = library->getArchive(QString::fromStdString(bookmark.getBookId()));
+        } catch (std::out_of_range& e) {
             continue;
-        std::string content;
-        std::string mimeType;
-        reader->getFavicon(content, mimeType);
+        }
+        auto illustration = archive->getIllustrationItem(48);
```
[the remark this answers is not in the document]
juuz0: How do you handle it?
Currently, I have just left the catch block empty.

src/contentmanager.cpp
```diff
@@ -235,7 +258,8 @@ QString ContentManager::downloadBook(const QString &id)
     }();
     auto downloadPath = KiwixApp::instance()->getSettingsManager()->getDownloadDir();
     QStorageInfo storage(downloadPath);
-    if (book.getSize() > storage.bytesAvailable()) {
+    auto bytesAvailable = storage.bytesAvailable();
+    if (bytesAvailable == -1 || book.getSize() > (unsigned) bytesAvailable) {
```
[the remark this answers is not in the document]
juuz0: https://doc.qt.io/qt-5/qstorageinfo.html#bytesAvailable says it returns `-1` if the object is invalid. That should be enough confirmation i think. Though I will change it to `bytesAvailable < 0` if you do not like

src/readinglistbar.cpp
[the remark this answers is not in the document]
mgautierfr: It depends of the use case.
We probably want to display nothing (in the library or in the tab icon) if the zim has no illustration. It seems that lefting the catch block empty is enough.
(In the library we display a "broken" image. It would be better to not display an image at all. But it is not the purpose of this PR)

src/contentmanager.cpp
```diff
@@ -80,13 +80,33 @@ QStringList ContentManager::getBookInfos(QString id, const QStringList &keys)
         ADD_V("url", getUrl);
         ADD_V("name", getName);
         ADD_V("origId", getOrigId);
-        ADD_V("faviconMimeType", getFaviconMimeType);
         ADD_V("downloadId", getDownloadId);
-        ADD_V("faviconUrl", getFaviconUrl);
         if (key == "favicon") {
-            auto s = b->getFavicon();
+            auto s = b->getIllustration(48)->getData();
```
mgautierfr, line 86: There is not try around this `getIllustration`

src/suggestionlistworker.cpp
```diff
@@ -48,13 +30,20 @@ void SuggestionListWorker::run()
...
-        searchSuggestionsSmart(archive, m_text.toStdString(), 15, suggestions);
-        for (auto& suggestion: suggestions) {
-            QString path = QString("/") + QString::fromStdString(suggestion.getPath());
-            url.setPath(path);
-            suggestionList.append(QString::fromStdString(suggestion.getTitle()));
-            urlList.append(url);
+        int suggestionsCount = 15;
+        auto prefix = m_text.toStdString();
+        auto suggestionSearcher = zim::SuggestionSearcher(*archive);
+        if (archive->hasTitleIndex()) {
```
mgautierfr, line 36: Why limiting suggestion to when archive has a title index ?
If archive has no title index, suggestion may indeed not be the best (as we search in the title listing) but it is better than nothing.

src/suggestionlistworker.cpp
```diff
@@ -48,13 +30,20 @@ void SuggestionListWorker::run()
...
+            auto suggestionSearch = suggestionSearcher.suggest(prefix);
+            const auto suggestions = suggestionSearch.getResults(0, suggestionsCount);
+            for (auto current : suggestions) {
+                kiwix::SuggestionItem suggestion(current.getTitle(), lcAll(current.getTitle()),
```
mgautierfr, line 40: Do we really need to use a `kiwix::SuggestionItem` ?
Could we feed suggestionList, url and urlList directly from `current` ?

mgautierfr (changes requested):

kelson42:
@juuz0 Any update?

mgautierfr:
@juuz0 I've added few fixup commit which should fix my remarks.
Please have a look to them before we merge this branch in master.

src/suggestionlistworker.cpp
```diff
@@ -8,6 +9,12 @@ SuggestionListWorker::SuggestionListWorker(const QString& text, int token, QObje
 {
 }
 
+std::string lcAll(std::string title)
```
juuz0, line 12: We can remove this too, not used anywhere now

juuz0:
@mgautierfr LGTM and works too. Only thing remaining is my comment above

src/suggestionlistworker.cpp
mgautierfr: Removed. Thanks

mgautierfr (approved):

PR #858, opened 2022-07-08
Fix compilation after removing of wrapping in libkiwix

mgautierfr:
Following the merge of kiwix/libkiwix#789 we have to adapt a bit the code here.


juuz0 (approved):
LGTM and everything works fine :+1:

PR #859, opened 2022-07-12
Hide top widget and tabbar in fullscreen

### LINKED ISSUE #571: Fix fullscreen if there is no current webview (zim)

juuz0:
This change hides the topwidget and tabbar when fullscreen shortcut/button is pressed.
If the mouse is taken to the top of screen, they are shown till the time it is there.
Fixes #571

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -22,6 +22,7 @@ MainWindow::MainWindow(QWidget *parent) :
     mp_ui->tabBar->setStackedWidget(mp_ui->mainView);
 
     auto app = KiwixApp::instance();
+    addAction(KiwixApp::instance()->getAction(KiwixApp::ToggleFullscreenAction));
 
     connect(app->getAction(KiwixApp::ExitAction), &QAction::triggered,
             this, &QMainWindow::close);
@@ -75,11 +76,45 @@ MainWindow::~MainWindow()
 }
 
 void MainWindow::toggleFullScreen() {
-    if (isFullScreen())
+    if (isFullScreen()) {
+        QApplication::instance()->removeEventFilter(this);
+        showTabAndTop();
         showNormal();
-    else
+    }
+    else {
+        QApplication::instance()->installEventFilter(this);
+        hideTabAndTop();
         showFullScreen();
+    }
 }
+
+void MainWindow::hideTabAndTop() {
+    getTabBar()->hide();
+    getTopWidget()->hide();
+}
+
+void MainWindow::showTabAndTop() {
+    getTabBar()->show();
+    getTopWidget()->show();
+}
+
+bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
+{
+    if (event->type() == QEvent::MouseMove && isFullScreen())
+    {
+        const auto mouseEvent = static_cast<QMouseEvent*>(event);
+        const int tabRegion = getTabBar()->height() + getTopWidget()->height() + 30;
+        // We don't have to check for visibilty as calling hide() on a hidden widget, or show() on a non-hidden widget is a no-op
+        if (mouseEvent->y() == 0) {
+            showTabAndTop();
+        } else if(mouseEvent->y() >= tabRegion) {
+            hideTabAndTop();
+        }
+        return true;
+    }
+    return false;
+}
+
 
 void MainWindow::when_ReadingList_toggled(bool state)
 {
--- a/src/mainwindow.h
+++ b/src/mainwindow.h
@@ -27,11 +27,14 @@ public:
 
 protected:
     void keyPressEvent(QKeyEvent *event);
+    bool eventFilter(QObject* object, QEvent* event) override;
 
 private slots:
     void toggleFullScreen();
     void when_ReadingList_toggled(bool state);
     void when_libraryPageDisplayed(bool showed);
+    void hideTabAndTop();
+    void showTabAndTop();
 
 private:
     Ui::MainWindow *mp_ui;


kelson42 (approved):
From a user perspective, the PR fix the problem the right way.

kelson42:
@veloman-yunkan Considering that this PR is already more than a week old that @mgautierfr will be away for a while, could you please review the code of this PR?

src/mainwindow.cpp
```diff
@@ -67,6 +68,8 @@ MainWindow::MainWindow(QWidget *parent) :
 
     mp_ui->contentmanagerside->setContentManager(app->getContentManager());
     mp_ui->sideBar->setCurrentWidget(mp_ui->contentmanagerside);
+    QApplication::instance()->installEventFilter(this);
```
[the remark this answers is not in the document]
juuz0: I tried mouse move events directly earlier, but they just never fired without a mouse click (mouse tracking was enabled).
[Here's a similar problem](https://stackoverflow.com/questions/32714105/mousemoveevent-is-not-called).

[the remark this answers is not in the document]
veloman-yunkan: Indeed. It looks that the mouse events are delivered to the innermost widget, necessitating the event filtering approach. However, I still recommend having event filtering enabled only while in full-screen mode).

Besides, there is no sense in calling `setMouseTracking()` on the next line.

src/mainwindow.cpp
```diff
@@ -75,12 +76,44 @@ MainWindow::~MainWindow()
...
+}
+
+bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
+{
+    if (event->type() == QEvent::MouseMove && isFullScreen())
+    {
+        const auto mouseEvent = static_cast<QMouseEvent*>(event);
+        if (mouseEvent->y() == 0) {
+            showTabAndTop();
+        } else if(mouseEvent->y() >= 150) {
```
veloman-yunkan, line 108: How did you select this value? In my environment it is noticeably below the bottom of the tab bar.

juuz0: The noticeably below was intentional - that's how Firefox does it also. Though value selection was not optimal, found another way this time.

src/mainwindow.cpp
```diff
@@ -75,12 +76,45 @@ MainWindow::~MainWindow()
...
+}
+
+void MainWindow::showTabAndTop() {
+    getTabBar()->show();
+    getTopWidget()->show();
+}
+
+bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
+{
+    const int tabRegion = getTabBar()->height() + getTopWidget()->height() + 30;
```
veloman-yunkan, line 103: Compute this value inside the body of the `if` statement.

src/mainwindow.cpp
```diff
@@ -75,12 +76,45 @@ MainWindow::~MainWindow()
...
+    if (event->type() == QEvent::MouseMove && isFullScreen())
+    {
+        const auto mouseEvent = static_cast<QMouseEvent*>(event);
+        if (mouseEvent->y() == 0) {
+            showTabAndTop();
+        } else if(mouseEvent->y() >= tabRegion) {
+            hideTabAndTop();
+        }
```
veloman-yunkan, line 111: Please state in a comment the (so far) implicit assumption that calling `hide()` on a hidden widget or `show()` on a non-hidden widget is a cheap no-op.

veloman-yunkan (changes requested):

src/mainwindow.cpp
```diff
@@ -75,12 +76,46 @@ MainWindow::~MainWindow()
...
+    getTopWidget()->show();
+}
+
+bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
+{
+    if (event->type() == QEvent::MouseMove && isFullScreen())
+    {
+        const auto mouseEvent = static_cast<QMouseEvent*>(event);
+        const int tabRegion = getTabBar()->height() + getTopWidget()->height() + 30;
+        // We don't have to check for visibilty as calling hide() on a hidden widget, or show() on a non-hidden widget is a no-op
```
veloman-yunkan, line 107: Since you left out the word "cheap" I would like to clarify that keeping `show()`ing a visible widget on every mouse move in the bigger part of the screen is acceptable only if that operation has negligible performance overhead rather than just being [idempotent](https://en.wikipedia.org/wiki/Idempotence). An example to the contrary would be if `QWidget::show()` unconditionally enqueued a repaint event.

veloman-yunkan (approved):

src/mainwindow.cpp
juuz0: > if that operation has negligible performance overhead rather than just being [idempotent](https://en.wikipedia.org/wiki/Idempotence).

Noted, thank you.

> An example to the contrary would be if QWidget::show() unconditionally enqueued a repaint event.

True, I checked into code and found out that there's a [check in setVisible](https://code.woboq.org/qt5/qtbase/src/widgets/kernel/qwidget.cpp.html#_ZN7QWidget10setVisibleEb) already. That should be fine to qualify it as a no-op I think?

veloman-yunkan: Yes.

PR #865, opened 2022-07-23
Don't monitor files while in All files view.

juuz0:
When downloading a file, with monitor directory set same as download directory, the changes in fs may send a double booksChanged signal leading to:
image

The first is due to adding/removing books while the second is the explicit booksChanged.
This change stops monitoring while in All files view.
To save the edge case when the user themselves adds books in monitor directory (while Kiwix is in All files view), a connection is added to Local Files button - to call loadMonitorDir().

### DIFF
--- a/src/contentmanager.h
+++ b/src/contentmanager.h
@@ -26,6 +26,7 @@ public:
     void setCurrentLanguage(QString language);
     void setCurrentCategoryFilter(QString category);
     void setCurrentContentTypeFilter(QList<ContentTypeFilter*>& contentTypeFilter);
+    bool isLocalLibrary() const { return m_local; }
 
 private:
     Library* mp_library;
--- a/src/contentmanagerside.cpp
+++ b/src/contentmanagerside.cpp
@@ -22,6 +22,13 @@ ContentManagerSide::ContentManagerSide(QWidget *parent) :
     connect(mp_ui->localFileButton, &QRadioButton::toggled,
             this, [=](bool checked) { mp_ui->localFileButton->setStyleSheet(
                     checked ?"*{font-weight: bold}" : "");});
+    connect(mp_ui->localFileButton, &QRadioButton::toggled,
+            this, [=](bool checked) {
+        if (checked) {
+            QString monitorDir = KiwixApp::instance()->getSettingsManager()->getMonitorDir();
+            KiwixApp::instance()->getLibrary()->asyncLoadMonitorDir(monitorDir);
+        }
+    });
     mp_ui->localFileButton->setStyleSheet("*{font-weight: bold}");
 
     mp_ui->allFileButton->setText(gt("all-files"));
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -102,7 +102,8 @@ void KiwixApp::init()
         }
     });
     connect(&m_watcher, &QFileSystemWatcher::directoryChanged, this, [=](QString monitorDir) {
-        m_library.asyncLoadMonitorDir(monitorDir);
+        if (getContentManager()->isLocalLibrary())
+            m_library.asyncLoadMonitorDir(monitorDir);
     });
     QString monitorDir = m_settingsManager.getMonitorDir();
     if (monitorDir != "") {


kelson42:
@juuz0 Thx for this PR. It's the first time I hear about that. Is that a regression? Why this bug, which does not look to be a detail, is detected now?

juuz0:
@kelson42 this behaviour:

- is inconsistent
- only happens when monitor directory is same as download directory

Think these reasons are why it was undetected till now.

I've found this on nightly from June & July and did get the behaviour in almost every after some downloads, inconsistent as I said.

kelson42:
@juuz0 I still wonder because both of us had tested the combination download/monitor directory being the same.

juuz0:
Yeah, really. I should try checking before June versions also.

... are you able to reproduce the issue on master though?

src/contentmanager.h
```diff
@@ -26,6 +26,7 @@ class ContentManager : public QObject
     void setCurrentLanguage(QString language);
     void setCurrentCategoryFilter(QString category);
     void setCurrentContentTypeFilter(QList<ContentTypeFilter*>& contentTypeFilter);
+    bool isLocalLibrary() { return m_local; }
```
veloman-yunkan, line 29: Can (and, therefore, should) be `const`

veloman-yunkan (changes requested):
My feeling is that the issue is being solved the wrong way. When I try to add a ZIM file to the library twice no duplicate entry is created. So why is it different for the monitor-directory scenario? I think that the solution to this and similar issues should fully belong to the `Library` class which must prevent appearance of duplicate entries.

juuz0:
> My feeling is that the issue is being solved the wrong way. When I try to add a ZIM file to the library twice no duplicate entry is created. So why is it different for the monitor-directory scenario?

The issue here is not duplicate entries. While downloading, the early emitting of booksChanged (just before download is completed) leads to a refresh of books (assumption) which causes that download-progress-circle and Open button to show up together

veloman-yunkan:
> The issue here is not duplicate entries. While downloading, the early emitting of booksChanged (just before download is completed) leads to a refresh of books (assumption) which causes that download-progress-circle and Open button to show up together

You should have put that in the description - while trying to decode the problem from the screenshot I thought that you were after the double appearance of WikiSummaries.

veloman-yunkan:
> While downloading, the early emitting of booksChanged (just before download is completed) leads to a refresh of books (assumption) which causes that download-progress-circle and Open button to show up together

My experience with kiwix-desktop is quite superficial so I still don't fully understand the problem but my feeling that it is being solved the wrong way persists. If the bug is about how the booksChanged signal is handled then it should be hiding in another place and your fix is only a workaround. Please state your understanding of the issue in more detail.

kelson42:
@juuz0 What is the status here?

juuz0:
I'll come up with an update by tomorrow morning

juuz0:
**Pre requisite:**

1. Set monitor directory same as download directory.

**Steps of reproduction:**

1. Go to All files section
2. Download a ZIM
3. At the end of download, it should show like this (Open button along with pause/close download buttons)
image


**Possible Reason:**

1. While the file is being download, it is constantly modified hence directoryChanged signals are sent for the watcher to catch.
2. This change is caught and `loadMonitorDir` is called but since the download is not complete, book can't be added. No change.
3. This catch-the-signal-and-try-to-add game happens all the time while file is being downloaded but no change happens.
4. When download is (almost/full) done, the book is added by monitorDir, which also sends a Library::booksChanged signal. This leads to refresh -> this confusion of Open and download status.

**Possible Solutions:**

1. We don't monitor while we are in All files view (current PR)
2. (simpler) Change (file is library.cpp)
```
    if (needsRefresh) {
        setMonitorDirZims(newDir.values());
        emit(booksChanged());
    }
```
to
```
    setMonitorDirZims(newDir.values());
    if (needsRefresh) {
        emit(booksChanged());
    }
```
Basically, always set the new directory values so we don't go on the loop of catch-the-signal-and-try-to-add.

Ping: @kelson42 @veloman-yunkan

veloman-yunkan:
> 4\. When download is (almost/full) done, the book is added by monitorDir, which also sends a Library::booksChanged signal. This leads to refresh -> this confusion of Open and download status.

IMHO, to get at the root cause of the issue you must drill down into what happens here. I still don't get why pause/close download buttons are displayed *after* the download has completed.

kelson42:
We should resume on this PR. I will try to reproduce the problem and give a helpful input.

kelson42:
@juuz0 Could you please rebase (and fix the conflicts) so we can finally finish this PR? Or maybe the PR is not relevent anymore at all?

kelson42:
@juuz0 I can not reproduce anymore the problem. Do you? Can you please update us?

juuz0:
@kelson42 I can't either. This probably got fixed while i was working on the library

mgautierfr:
Maybe it this [change](https://github.com/kiwix/kiwix-desktop/commit/5ee18f1b40abbaa61661a0f3566cbb0657d0320d) which has fixed the issue ?

PR #867, opened 2022-08-06
Add shortcuts for navigating tabs.

### LINKED ISSUE #866: Feature Request: Cycle through tabs with <ctrl>+<tab> and <ctrl>+<shift>+<tab>
It'd be nice to be able to cycle through tabs with `<ctrl>+<tab>` and `<ctrl>+<shift>+<tab>`. Those two keyboard shortcuts are pretty fundamental to my workflow in typical browsers, and it's been pretty jarring to not have that available when eg browsing Wikipedia offline.

(Sorry if this isn't the right place for feature requests.)

juuz0:
Added:
- CTRL + TAB shortcut for going to the next tab.
- CTRL + SHIFT + TAB shortcut for going to previous tab.

Fixes #866

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -127,6 +127,14 @@ void TabBar::setNewTabButton()
     setTabButton(idx, QTabBar::RightSide, Q_NULLPTR);
 }
 
+int TabBar::realTabCount() const
+{
+    // The last tab is "+" in TabBar, but that isn't a real tab which displays any content hence the real count is tab count - 1
+    if (count() < 1)
+        return 0;
+    return count() - 1;
+}
+
 ZimView* TabBar::createNewTab(bool setCurrent, bool adjacentToCurrentTab)
 {
     auto tab = new ZimView(this, this);
@@ -134,7 +142,7 @@ ZimView* TabBar::createNewTab(bool setCurrent, bool adjacentToCurrentTab)
     if(adjacentToCurrentTab) {
         index = currentIndex() + 1;
     } else {
-        index = count() - 1; // for New Tab Button
+        index = realTabCount(); // for New Tab Button
     }
     mp_stackedWidget->insertWidget(index, tab);
     index = insertTab(index, "");
@@ -255,7 +263,7 @@ void TabBar::closeTabsByZimId(const QString &id)
 void TabBar::closeTab(int index)
 {
     // the last tab is + button, cannot be closed
-    if (index == this->count() - 1)
+    if (index == this->realTabCount())
         return;
 
     setSelectionBehaviorOnRemove(index);
@@ -289,8 +297,8 @@ void TabBar::onCurrentChanged(int index)
         return;
 
     // if somehow the last tab (+ button) became active, switch to the previous
-    if (index >= (count() - 1)) {
-        setCurrentIndex(count() - 2);
+    if (index >= realTabCount()) {
+        setCurrentIndex(realTabCount() - 1);
         return;
     }
 
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -24,6 +24,8 @@ TabBar::TabBar(QWidget *parent) :
     connect(this, &QTabBar::currentChanged, this, &TabBar::onCurrentChanged, Qt::QueuedConnection);
     auto app = KiwixApp::instance();
 
+    connect(app->getAction(KiwixApp::NextTabAction), &QAction::triggered, this, &TabBar::moveToNextTab);
+    connect(app->getAction(KiwixApp::PreviousTabAction), &QAction::triggered, this, &TabBar::moveToPreviousTab);
     connect(app->getAction(KiwixApp::NewTabAction), &QAction::triggered,
             this, [=]() {
                 this->createNewTab(true, false);
@@ -133,6 +135,18 @@ int TabBar::realTabCount() const
     if (count() < 1)
         return 0;
     return count() - 1;
+}
+
+void TabBar::moveToNextTab()
+{
+    const int index = currentIndex();
+    setCurrentIndex(index == realTabCount() - 1 ? 0 : index + 1);
+}
+
+void TabBar::moveToPreviousTab()
+{
+    const int index = currentIndex();
+    setCurrentIndex(index <= 0 ? realTabCount() - 1 : index - 1);
 }
 
 ZimView* TabBar::createNewTab(bool setCurrent, bool adjacentToCurrentTab)
--- a/src/tabbar.h
+++ b/src/tabbar.h
@@ -64,7 +64,10 @@ private:
     QScopedPointer<FullScreenWindow> m_fullScreenWindow;
 
     void setSelectionBehaviorOnRemove(int index);
-
+    // The "+" (new tab) button is implemented as a tab (that is always placed at the end).
+    // This function returns the count of real tabs.
+    int realTabCount() const;
+    
 private slots:
     void onTabMoved(int from, int to);
     void onCurrentChanged(int index);
--- a/src/tabbar.h
+++ b/src/tabbar.h
@@ -58,6 +58,8 @@ public slots:
     void closeTab(int index);
     void fullScreenRequested(QWebEngineFullScreenRequest request);
     void on_webview_titleChanged(const QString& title);
+    void moveToNextTab();
+    void moveToPreviousTab();
 
 private:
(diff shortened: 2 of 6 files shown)


kelson42 (approved):
From a user perspective, works as expected. Thank you @juuz0

src/mainmenu.cpp
```diff
@@ -44,6 +44,8 @@ MainMenu::MainMenu(QWidget *parent) :
     m_viewMenu.ADD_ACTION(ZoomInAction);
     m_viewMenu.ADD_ACTION(ZoomOutAction);
     m_viewMenu.ADD_ACTION(ZoomResetAction);
+    m_viewMenu.ADD_ACTION(NextTabAction);
+    m_viewMenu.ADD_ACTION(PreviousTabAction);
```
veloman-yunkan, line 48: Do we really need these in the menu?

src/tabbar.cpp
```diff
@@ -127,6 +129,24 @@ void TabBar::setNewTabButton()
...
 }
 
+void TabBar::moveToNextTab()
+{
+    auto index = currentIndex() + 1;
+    if (index == count() - 1) {
+        index = 0; // if we are on last tab, move to first tab (library).
+    }
+    setCurrentIndex(index);
+}
```
veloman-yunkan, line 139: If I understand correctly, this jumps from the last-but-one tab to the first tab (skips the last tab).

src/tabbar.cpp
```diff
@@ -127,6 +129,24 @@ void TabBar::setNewTabButton()
...
+
+void TabBar::moveToPreviousTab()
+{
+    auto index = currentIndex() - 1;
+    if (index < 0) {
+        index = count() - 1; // if we are on first tab, move to last tab.
+    }
```
veloman-yunkan, line 146: This relies on the fact that `currentIndex()` returns a signed type. Better rewrite it in a way that would work for an unsigned type too (check before decrementing).

veloman-yunkan (changes requested):

src/mainmenu.cpp
juuz0: Let's ask @kelson42

src/tabbar.cpp
juuz0: Yes, the last tab in TabBar is "+" icon

src/mainmenu.cpp
kelson42: Probably not, how it is in Firefox?

src/tabbar.cpp
veloman-yunkan: Then make it explicit (with the help of a comment)

veloman-yunkan: A more explicit way of addressing this subtlety (instead of having to mention and explain it at every location where it matters) is to introduce a function `realTabCount()` and document the issue in only one place

src/tabbar.cpp
veloman-yunkan: In the light of https://github.com/kiwix/kiwix-desktop/pull/867#discussion_r939687653, shouldn't you move from the first tab to the last-but-one tab?

juuz0: > In the light of https://github.com/kiwix/kiwix-desktop/pull/867#discussion_r939687653, shouldn't you move from the first tab to the last-but-one tab?

Correct, that was wrong code but the behaviour was saved by the check in `void TabBar::onCurrentChanged(int index)`

Fixed anyway. Didn't like to be dependent on that check

src/tabbar.cpp
juuz0: Done, thanks!

src/mainmenu.cpp
juuz0: Not available in Firefox, removed.

src/tabbar.cpp
```diff
@@ -127,6 +129,32 @@ void TabBar::setNewTabButton()
     setTabButton(idx, QTabBar::RightSide, Q_NULLPTR);
 }
 
+int TabBar::realTabCount()
+{
+    // The last tab is "+" in TabBar, but that isn't a real tab which displays any content hence the real count is tab count - 1
+    if (count() - 1 < 0)
+        return 0;
```
veloman-yunkan, line 136: `count() - 1` runs the risk of underflowing if `count()` returns an unsigned type. `count() < 1` is a mathematically equivalent but computationally safer way of expressing the same idea.

src/tabbar.cpp
```diff
@@ -127,6 +129,32 @@ void TabBar::setNewTabButton()
...
+    auto index = currentIndex() + 1;
+    if (index == realTabCount()) {
+        index = 0;
+    }
+    setCurrentIndex(index);
+}
+
+void TabBar::moveToPreviousTab()
+{
+    auto index = currentIndex() - 1;
```
veloman-yunkan, line 151: You didn't address https://github.com/kiwix/kiwix-desktop/pull/867#discussion_r939554592

src/tabbar.cpp
```diff
@@ -127,6 +129,32 @@ void TabBar::setNewTabButton()
     setTabButton(idx, QTabBar::RightSide, Q_NULLPTR);
 }
 
+int TabBar::realTabCount()
+{
+    // The last tab is "+" in TabBar, but that isn't a real tab which displays any content hence the real count is tab count - 1
```
veloman-yunkan, line 134: Better move (or repeat) this comment at the declaration of `realTabCount()`

src/tabbar.h
```diff
@@ -58,12 +58,15 @@ public slots:
...
     void on_webview_titleChanged(const QString& title);
+    void moveToNextTab();
+    void moveToPreviousTab();
 
 private:
     QStackedWidget*     mp_stackedWidget;
     QScopedPointer<FullScreenWindow> m_fullScreenWindow;
 
     void setSelectionBehaviorOnRemove(int index);
+    int realTabCount();
```
veloman-yunkan, line 69: Make `const`

veloman-yunkan (changes requested):
If you can find applications of `realTabCount()` in pre-existing code, please  introduce that function (and immediately utilize it) in a separate commit.

src/tabbar.cpp
juuz0: Repeated.

src/tabbar.cpp
juuz0: Done now (though I had a hard time understanding what to do, let me know please if there's a better way!)

src/tabbar.cpp
juuz0: Eagle eye

src/tabbar.cpp
```diff
@@ -135,6 +137,24 @@ int TabBar::realTabCount() const
...
+    setCurrentIndex(index);
+}
+
+void TabBar::moveToPreviousTab()
+{
+    int index = currentIndex() <= 0 ? -1 : currentIndex() - 1;
+    if (index == -1) {
+        index = realTabCount() - 1;
+    }
+    setCurrentIndex(index);
```
veloman-yunkan, line 155: Why so complex? Why do you set `index` to `-1` if you know that you are going to immediately change it to `realTabCount() - 1`? This code can be simplified, for example, to:

```c++
const int i = currentIndex();
setCurrentIndex(i <= 0 ? realTabCount() - 1 : i - 1);
```

Whichever variant you choose, please use the same style for both `moveToNextTab` and `moveToPreviousTab()`.  

BTW, can `currentIndex()` return a negative value (at least in a context where `moveToPreviousTab()` can be called)?

veloman-yunkan (changes requested):

src/tabbar.h
```diff
@@ -58,13 +58,15 @@ public slots:
...
+    void moveToNextTab();
+    void moveToPreviousTab();
 
 private:
     QStackedWidget*     mp_stackedWidget;
     QScopedPointer<FullScreenWindow> m_fullScreenWindow;
 
     void setSelectionBehaviorOnRemove(int index);
-    // The last tab is "+" in TabBar, but that isn't a real tab which displays any content hence the real count is tab count - 1
+    // Gives the count of tabs which display any content
```
veloman-yunkan, line 69: 1. Please change to

> // The "+" (new tab) button is implemented as a tab (that is always placed at the end).
> // This function returns the count of real tabs.

(and do it in the first commit)

veloman-yunkan (changes requested):

veloman-yunkan (approved):

PR #870, opened 2022-08-12
Hide ZIMIt books from library

### LINKED ISSUE #780: Zim files requiring Service Workers should be hidden from the library if SW are not available
Following up on [Libkiwix#696](../../libkiwix/issues/696)

Currently zimit-generated files can only be read on Kiwix-Android and Kiwix-serve (K-Desktop has a workaround but it's not main stuff). Kiwix-desktop cannot read these files at the moment, yet it is likely that at least some users will invest data transfers (and costs) into downloading files they ultimately cannot access.

Since zimit-generated files are tagged as such, it would make sense that they be hidden from the app's internal library if we know it cannot be read.

juuz0:
This change hides books which contain a _sw tag (created by ZIMIt).
This is a temporary fix until support of SW based ZIMs is added.
Fixes #780

### DIFF
--- a/resources/js/_contentManager.js
+++ b/resources/js/_contentManager.js
@@ -24,11 +24,13 @@ function setTranslations(translations) {
 
 const BOOK_KEYS = ["id", "name", "path", "url", "size", "description", "title", "tags", "date", "faviconUrl", "faviconMimeType", "downloadId"];
 function addBook(values) {
-  var b = createDict(BOOK_KEYS, values);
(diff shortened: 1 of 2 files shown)


kelson42:
@juuz0 This should better have been done witha proper API request.

mgautierfr:
Agree with @kelson42, it is to the contentManager exclude zimit books.
We should exclude the books in `ContentManager::getBookIds()` not in the part getting information about a book.

PR #880, opened 2022-09-07
Version 2.3.0

mgautierfr:


mgautierfr:
Debian packaging is broken. But we need to do a release. We will fix packaging later.

PR #886, opened 2022-09-18
Improve zim file picker

kelson42:


juuz0:
LGTM, are split ZIM files only of the type **zimaa**? (I thought zimab, zimac, etc were also possible).

kelson42:
Only .zimaa can be open (this is the begin of the zim file).

PR #887, opened 2022-09-18
Filter out service worker based content from the online library

kelson42:
Fixes https://github.com/kiwix/kiwix-desktop/issues/780 and superseeds #870

### DIFF
--- a/src/opdsrequestmanager.cpp
+++ b/src/opdsrequestmanager.cpp
@@ -12,24 +12,32 @@ OpdsRequestManager::OpdsRequestManager()
 void OpdsRequestManager::doUpdate(const QString& currentLanguage, const QString& categoryFilter)
 {
     QUrlQuery query;
+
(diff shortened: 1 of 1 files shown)


kelson42:
@mgautierfr  @veloman-yunkan Can I merge this?

juuz0 (changes requested):
Unfortunately, doesn't work for me, file still shows up in my library.
This is the file
https://drive.google.com/file/d/1e_sSHD86NI6u7A-LetWW1C6Ap16w1p2d/view?usp=sharing

kelson42:
Yes, this is for the online library only. To me this is good enough. If someone wants really to load one, and use it with HTTP daemon, then it should be possible. This is why I have talked of OPDS stream in the ticket, but thiscwad not super specific indeed.

veloman-yunkan (approved):

PR #919, opened 2023-01-25
Windows download hang

### LINKED ISSUE #87: Kiwix goes into non responsive when trying to download 78GB wikipedia
Honestly this was expected. Is there a better way for Kiwix to handle downloads of such large files?
While I was doubting I did check the file size in windows explorer and indeed it is going up in size, however because of the large size, the computer is having difficulty with using the full capabilities of the network card.

Basically there should be a cleaner way for Kiwix handling such large downloads.

adamlamar:
Overall, the high level design is something like:

* Run a new QThread using the BackgroundDownloader. Any slots invoked on the BackgroundDownloader are run on this new thread (not the UI thread)
* The ContentManager sends signals from the UI thread to the BackgroundDownloader thread.
* When needed, the BackgroundDownloader sends signals back to the ContentManager to confirm operations, such as starting or canceling a download
* The BackgroundDownloader::updateStatus() is invoked once per second by a QTimer, and updates the internal m_status map with information about the download
* The BackgroundDownloader:: getDownloadStatus() method can be called from the ContentManager on the UI thread to get the status of any particular download
* Since BackgroundDownloader can be called from two different threads, reading from the m_status map is protected by a ReadWrite lock

Fixes #87

### DIFF
--- a/src/backgrounddownloader.cpp
+++ b/src/backgrounddownloader.cpp
@@ -46,9 +46,7 @@ void BackgroundDownloader::startDownload(const QString& bookId, const QString& u
     try {
         std::pair<std::string, std::string> downloadDir("dir", downloadPath.toStdString());
         const std::vector<std::pair<std::string, std::string>> options = { downloadDir };
-        m_rwlock.lockForWrite();
         download = mp_downloader->startDownload(uri.toStdString(), options);
-        m_rwlock.unlock();
         std::cerr << "got download with id " << download->getDid() << '\n';
         emit(confirmStartDownload(bookId, QString::fromStdString(download->getDid())));
     } catch (std::exception& e) {
@@ -61,11 +59,11 @@ void BackgroundDownloader::startDownload(const QString& bookId, const QString& u
 void BackgroundDownloader::completeDownload(const QString& did)
 {
     kiwix::Download* download = mp_downloader->getDownload(did.toStdString());
+    download->cancelDownload();
 
     m_rwlock.lockForWrite();
     QMap<std::string, std::string> map = m_status.value(did.toStdString());
     m_status.remove(did.toStdString());
-    download->cancelDownload();
     m_rwlock.unlock();
 }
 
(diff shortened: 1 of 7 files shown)


src/backgrounddownloader.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+
+void BackgroundDownloader::cleanup() {
+    // TODO: delete resources
+
+    m_thread->quit();
+}
+
+// getDownloadStatus returns the status map for this downloadId. May be called from a separate thread
+QMap<std::string, std::string> BackgroundDownloader::getDownloadStatus(const std::string did)
```
adamlamar, line 35: `getDownloadStatus()` is called by `ContentManager:: updateDownloadInfos()`, which is running on the UI thread. `m_rwlock` protects the map read.

src/backgrounddownloader.cpp
```diff
@@ -0,0 +1,152 @@
...
+        status.insert("downloadSpeed", std::to_string(download->getDownloadSpeed()));
+        status.insert("verifiedLength", std::to_string(download->getVerifiedLength()));
+        status.insert("status", convertStatusResult(download->getStatus()));
+        m_status.insert(did, status);
+        m_rwlock.unlock();
+    }
+    int mapSize = 0;
+    for (auto it = m_status.begin(); it != m_status.end(); ++it) {
+        mapSize++;
+    }
```
adamlamar, line 129: This iterator loop was just debugging a problem with an unexpected map size being reported. Its equivalent to `m_status.size()`. 

I found the download map was bigger than expected because `mp_downloader->getDownloadIds()` never forgets downloads. Even though the map entry is removed during cancel, it gets added back again because the old download ID is always returned.

src/backgrounddownloader.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <QString>
+#include <QTimer>
+#include <QDebug>
+
+BackgroundDownloader::BackgroundDownloader(kiwix::Downloader* downloader)
+    :  mp_downloader(downloader)
+{
+    QThread* newThread = new QThread(); // no parent
+    m_thread.reset(newThread);
+    moveToThread(m_thread.get());
```
adamlamar, line 13: This parentless thread and call to `moveToThread()` is key to ensuring that `BackgroundDownloader` slots are invoked on their own thread that is separate from the UI thread.

kelson42:
Hi @adamlamar. Have you given up on this PR? After the implementation of https://github.com/kiwix/kiwix-desktop/pull/921, nothing really blocks the completion here, do I'm right?

kelson42:
@mgautierfr Its important to finally complete this PR.

adamlamar:
Yep! Closing

PR #955, opened 2023-07-07
Disable random article in settings view

### LINKED ISSUE #954: Random icon must be greyed if settings open
image

juuz0:
Fixes #954

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -435,21 +435,22 @@ void KiwixApp::createAction()
 }
 
 void KiwixApp::postInit() {
-    connect(getTabWidget(), &TabBar::libraryPageDisplayed,
-            this, &KiwixApp::disableItemsOnLibraryPage);
+    connect(getTabWidget(), &TabBar::tabDisplayed,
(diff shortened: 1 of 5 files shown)


veloman-yunkan (commented):
A more proper fix seems to be the following:

1. change the type of the parameter of `KiwixApp::disableItemsOnLibraryPage(bool libraryDisplayed)` and `TabBar::libraryPageDisplayed(bool displayed)`
from `bool` to a new enum type which tells the kind of the current tab (content, library, settings). Inside that function enable only actions that make sense for the current tab kind.
2. Naturally after that change `KiwixApp::disableItemsOnLibraryPage()` and `TabBar::libraryPageDisplayed(bool displayed)` have to be renamed.

mgautierfr:
Agree with @veloman-yunkan, those functions have been created when there was no setting page. Now we have more than 2 kinds of tab, we should create more generic functions to handle "action configuration" when we change tabs.

kelson42:
@juuz0 Ready for a new review?

juuz0:
@kelson42 yes

veloman-yunkan (approved):

PR #960, opened 2023-07-11
OPDS based languages and categories

### LINKED ISSUE #934: Don't use a static list of language/categories in the catalog filtering.
When we have added the language selector, the catalog hadn't the language list feature.
So we've set a fixed list of languages in kiwix-desktop.

When next version of libkiwix will be released, catalog will have a API to get the list of languages and categories. We should use it.

juuz0:
Fixes #934 
Requires libkiwix from: https://github.com/kiwix/libkiwix/pull/967


mgautierfr (changes requested):

src/contentmanager.cpp
```diff
@@ -47,6 +54,42 @@ QStringList ContentManager::getTranslations(const QStringList &keys)
...
+        categories.push_back("all");
+        for (auto category : categoryData) {
+            auto categoryName = QString::fromStdString(category.get("name")->string_value());
+            categories.push_back(categoryName);
+        }
+        m_categories = categories;
+        emit(categoriesLoaded(m_categories));
+        return;
+    }
+    m_remoteLibraryManager.getCategoriesFromOpds();
```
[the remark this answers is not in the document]
juuz0: Was hesitant with doing changes in libkiwix but done now!

kelson42:
@juuz0 Can you please rebase and update about the status? Are you just missing a new review?

juuz0:
@kelson42 Yes, it's missing a new review. The libkiwix side is merged and done.
@mgautierfr Please!

src/contentmanager.cpp
```diff
@@ -585,6 +587,18 @@ void ContentManager::updateRemoteLibrary(const QString& content) {
     });
 }
 
+void ContentManager::updateLanguages(const QString& content) {
+    m_remoteLibrary = kiwix::Library();
```
mgautierfr, line 632: This re initialize the`m_remoteLibrary`. Why we need this ?
If we update categories after languages, we will lost the known languages.

src/opdsrequestmanager.cpp
```diff
@@ -49,19 +48,62 @@ void OpdsRequestManager::doUpdate(const QString& currentLanguage, const QString&
     qInfo() << "Downloading" << url.toString(QUrl::FullyEncoded);
 
     QNetworkRequest request(url);
-    if (mp_reply) {
-        mp_reply->abort();
-    }
-    mp_reply = m_networkManager.get(request);
-    connect(mp_reply, &QNetworkReply::finished, this, &OpdsRequestManager::receiveContent);
+    auto mp_reply = m_networkManager.get(request);
```
mgautierfr, line 50: Could we set the builded url (before this line) with `opdsResponseFromPath` ?

src/contentmanager.cpp
```diff
@@ -70,7 +70,7 @@ ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader,
     connect(treeView, &QTreeView::doubleClicked, this, &ContentManager::openBookWithIndex);
     connect(&m_remoteLibraryManager, &OpdsRequestManager::languagesReceived, this, &ContentManager::updateLanguages);
     connect(&m_remoteLibraryManager, &OpdsRequestManager::categoriesReceived, this, &ContentManager::updateCategories);
-    setLanguages();
+    setCategories();
```
mgautierfr, line 72: This comment is for the commit message, not this line.

Commit message say : `If local library is displayed, we get the categories from LibraryDumper`
It is false as the categories are taken from Library itself (and it is a good thing)

mgautierfr (changes requested):

src/contentmanager.cpp
juuz0: This is what happens when I forget things to remove while editing commits in a rebase.
Thanks and fixed.

src/opdsrequestmanager.cpp
juuz0: Im divided on whats the better way here
1. Have 2 functions:
`opdsResponseFromPath(path)` (which calls the next one with an empty query)
`opdsResponseFromPath(path, query)`

2. Have a default empty query parameter
`opdsResponseFromPath(path, query = "")`

src/kiwixapp.h
```diff
@@ -86,6 +86,7 @@ class KiwixApp : public QtSingleApplication
     QString getText(const QString &key) { return m_translation.getText(key); };
     void setMonitorDir(const QString &dir);
     bool isCurrentArticleBookmarked();
+    kiwix::UpdatableNameMapper* getNameMapper() { return &m_nameMapper; }
```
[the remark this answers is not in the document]
juuz0: None. Forgot to remove while changing methods

src/contentmanager.cpp
juuz0: Fixed in history

juuz0:
Rebased fixups

PR #963, opened 2023-07-16
Deleting files move them to trash

### LINKED ISSUE #479: Deleting a ZIM file should move it to the trash can

juuz0:
Fixes #479


kelson42:
@juuz0 Superseeds #531?

kelson42:
> @juuz0 Superseeds #531?

@juuz0 ???

resources/i18n/en.json
```diff
@@ -144,5 +144,8 @@
     "download-storage-error": "Storage Error",
     "download-storage-error-text": "The system doesn't have enough storage available.",
     "download-unavailable": "Download Unavailable",
-    "download-unavailable-text": "This download is unavailable."
+    "download-unavailable-text": "This download is unavailable.",
+    "move-files-to-trash": "Move deleted files to trash",
+    "move-files-to-trash-text": "<br><br><i> This action will move the file to trash. </i>",
+    "perma-delete-files-text": "<br><br><i> This action will permanently delete the file. </i>"
```
mgautierfr, line 150: We should put the html tags out of the translatable text.

src/settingsmanager.cpp
```diff
@@ -89,11 +89,18 @@ void SettingsManager::setMonitorDir(QString monitorDir)
     emit(monitorDirChanged(monitorDir));
 }
 
+void SettingsManager::toggleTrashCheck()
```
mgautierfr, line 92: `toggleTrashCheck` is really implementation oriented.
Functions should be named about the "functionnal" they do, not how they are implemented.

A better name should be `toggleMoveToTrash`

src/settingsmanager.h
```diff
@@ -27,13 +27,15 @@ class SettingsManager : public QObject
     qreal getZoomFactor() const { return m_zoomFactor; }
     QString getDownloadDir() const { return m_downloadDir; }
     QString getMonitorDir() const { return m_monitorDir; }
+    bool getTrashCheck() const { return m_moveToTrash; }
```
mgautierfr, line 30: Same here, a better name is `getMoveToTrash`

src/settingsmanager.h
```diff
@@ -51,6 +53,7 @@ public slots:
     qreal m_zoomFactor;
     QString m_downloadDir;
     QString m_monitorDir;
+    bool m_moveToTrash;
```
mgautierfr, line 56: It would be good to also create a signal for `m_moveToTrash`.

mgautierfr (changes requested):

juuz0:
> Superseeds https://github.com/kiwix/kiwix-desktop/pull/531?

@kelson42 Yes.

src/kiwixapp.cpp
```diff
@@ -105,10 +105,14 @@ void KiwixApp::init()
...
-        m_watcher.addPath(monitorDir);
-        m_library.asyncLoadMonitorDir(monitorDir);
+    QString downloadDir = m_settingsManager.getDownloadDir();
+    auto dirList = QSet<QString>({monitorDir, downloadDir});
+    for (auto dir : dirList) {
+        if (dir != "") {
+            m_library.setMonitorDirZims(dir, m_library.getLibraryZimsFromDir(dir));
+            m_watcher.addPath(dir);
+            m_library.asyncLoadMonitorDir(dir);
+        }
```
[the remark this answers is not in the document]
juuz0: I'm on the assumption that adding the same book using manipulator will not add it twice here. Is it possible to duplicate?

src/settingsview.cpp
```diff
@@ -21,6 +21,7 @@ SettingsView::SettingsView(QWidget *parent)
     connect(KiwixApp::instance()->getSettingsManager(), &SettingsManager::downloadDirChanged, this, &SettingsView::onDownloadDirChanged);
     connect(KiwixApp::instance()->getSettingsManager(), &SettingsManager::monitorDirChanged, this, &SettingsView::onMonitorDirChanged);
     connect(KiwixApp::instance()->getSettingsManager(), &SettingsManager::zoomChanged, this, &SettingsView::onZoomChanged);
+    connect(ui->trashCheck, &QCheckBox::clicked, KiwixApp::instance()->getSettingsManager(), &SettingsManager::toggleTrashCheck);
```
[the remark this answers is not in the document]
juuz0: Made naming and everything more consistent.

src/library.cpp
```diff
@@ -139,7 +139,7 @@ void Library::loadMonitorDir(QString monitorDir)
     QMutexLocker locker(&mutex);
```
mgautierfr, line 139: This is useless if `mutex` is not shared across `locker`s.
As mutex is here a "simple" variable, each running `loadMonitorDir` has its own mutex and nothing is protected.

src/kiwixapp.cpp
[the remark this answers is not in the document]
mgautierfr: I think we should be mostly ok:
- QFileSystemWatcher handles duplicate correctly
- `libkiwix::Library` handles duplicate (`Manager` doesn't but I think it is ok)
- `Library::m_monitorDirZims` splits different directory in different list (and handles duplicate also as it is now a QMap)

However, the problem is race condition. If we have two `loadMonitorDir` running in the same time, `Library::m_monitorDirZims` is not protected against multi threaded access. (`libkiwix::Library` is).
And we launch at least two `loadMonitorDir` in the same time here with the two `m_library.asyncLoadMonitorDir(dir);`
And it would be even more tricky if `monitorDir` == `downloadDir`.
Properly protecting `loadMonitorDir` from concurrent executions **should** be enough.

mgautierfr (changes requested):

juuz0:
I've a concern with this PR - this won't work on focal. `moveToTrash` was added in Qt 5.15, but focal is 5.12.
I see 2 solutions when QT version < 5.15:
1. write custom platform specific code to move to trash. I am not exactly sure how this would work, will search and see.
2. disable the option to move to trash when the version is lower than 5.15

kelson42:
> I've a concern with this PR - this won't work on focal. `moveToTrash` was added in Qt 5.15, but focal is 5.12.

@juuz0 Thank you for raising this point. Here many remarks regarding this:
* Qt 5.15 is not on Focal (our oldest supported Ubuntu version), only in Jammy 22.04
* For our own build, we could maybe (should be doable) move to Qt15.5, but this is a question for @mgautierfr 
* Moving to Qt6 is a topic anyway, see https://github.com/kiwix/kiwix-build/issues/594

I believe your approach is very pragramatic and if the (2) option is not too complicated, please implement it. On our side we should really consider to move to Qt15.5 in our compilation toolchain.

kelson42 (approved):
From a functional perspective, tested on Ubuntu 22.04, it works perfectly to me.

juuz0:
> I believe your approach is very pragramatic and if the (2) option is not too complicated, please implement it.

Done

src/library.h
```diff
@@ -51,6 +52,7 @@ public slots:
     void bookmarksChanged();
 
 private:
+    QMutex mutex;
```
mgautierfr, line 55: Would be better to give a more useful name. Something like `m_updateFromDirMutex`

src/library.cpp
```diff
@@ -135,7 +135,6 @@ QStringList Library::getLibraryZimsFromDir(QString dir) const
 
 void Library::loadMonitorDir(QString monitorDir)
```
mgautierfr, line 136: I wonder (not sure) if this should be renamed to `updateFromDir`

src/library.h
```diff
@@ -51,6 +52,7 @@ public slots:
...
     kiwix::Library m_library;
     QString m_libraryDirectory;
     QMap<QString, QStringList> m_monitorDirZims;
```
mgautierfr, line 58: I wonder (not sure) if this should be renamed to `m_knownZimsInDir`

mgautierfr (changes requested):
A renaming of the mutex to do.

I'm not sure about other renaming. I'm open to discussion here. (Or you can directly do the change if you agree with me)

src/library.cpp
juuz0: Done

src/library.h
juuz0: Done

src/library.h
juuz0: Yeah, these all naming makes sense because it's a generic function now. Not only monitorDir specific

mgautierfr (changes requested):
Few comments on the last commit.

kelson42:
@juuz0 Please fix the last details so we can merge before the WE.

src/contentmanager.cpp
```diff
@@ -498,7 +502,10 @@ QString formatText(QString text)
 void ContentManager::eraseBook(const QString& id)
 {
     auto text = gt("delete-book-text");
-    const auto moveToTrash = KiwixApp::instance()->getSettingsManager()->getMoveToTrash();
+    auto moveToTrash = KiwixApp::instance()->getSettingsManager()->getMoveToTrash();
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
+    moveToTrash = false; // we do not support move to trash functionality for qt versions below 5.15
+#endif
```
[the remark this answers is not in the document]
juuz0: Neat stuff, fixed in commit history

kelson42:
@juuz0 Thank you for the last cosmetic fix, I will merge.

Issue #957, opened 2023-07-09
Online library filter/search are not optimal


kelson42:
For the moment, a new request of type "https://library.kiwix.org:443/catalog/search?lang=amh&count=-1&tag=_category:gutenberg&notag=_sw:yes" is done each time a filter is changed in the sidebar. Then everything (book details, no pagination) is downloaded from https://library.kiwix.org. The "search" (currently library top bar) is then applied _a posteriori_ on the local result list to have a match.

I'm not satified because that generates a lot of data transfered (slow and expensive) between local computer and library.kiwix.org.

I woud like to propose:
* Free text search is handled like any other filter and generates a new API request if changing
* We don't download more book details than what we can display on the screen (but we need to know how many results we have in total). We should use result pagination.
* Libkiwix should cache properly things
  * We should have a cache of requests/response that if we have twice the same requests we identify and avoid downloading twice the results
  * Results list should be handled differently from books. Results reonses should only have the list of book id and book details should be requested (and cached) separatly. That way we can cache book details independently from the API request/response (and save bandwidth).

kelson42:
@mgautierfr @veloman-yunkan @juuz0 Does that make sense? Is the libkiwix ready for that?

juuz0:
> We don't download more book details than what we can display on the screen (but we need to know how many results we have in total). We should use result pagination.

One interesting problem here is to handle the sorting of books.
Current kiwix-desktop applies sort on all of the list. But obviously, if we use pagination, it's only going to apply on the list available to it and give false positives (eg: sorting in descending order by size will not give the book with biggest size in *Global* catalog).
libkiwix should probably support a parameter to give out sorted data, if we need this.

kelson42:
> libkiwix should probably support a parameter to give out sorted data, if we need this.

Yes, we have multiple tickets regarding sorting of results. But we could handle this with https://github.com/kiwix/libkiwix/issues/702. Please put the requirements you have in a comment so you can secure Kiwix-Desktop features can continue to work.

juuz0:
does the cache here have to be implemented in libkiwix?
QT has QNetworkDiskCache which can help in this

kelson42:
@mgautierfr We urgently need your feedback here

mgautierfr:
> Free text search is handled like any other filter and generates a new API request if changing

[...]

PR #965, opened 2023-07-23
Improved filters panel

### LINKED ISSUE #958: Library filters components have to be improved
Today the current solution to display a select list (used for languages, categories, content types) has the following weaknesses:
* It's not immediatly visible what is the current choice(s)
* Impossible to make multiple choices
* If the list is very long (like for the languages or categories), very combersome (no free text search)

### LINKED ISSUE #973: Big space at bottom of library category list seems not appropriate
image

Can we somehow adjust the list height dynamically based on amount of categories available?

### LINKED ISSUE #974: Remember library filter values over restart
For:
* Categories
* Languages
* Content type

juuz0:
Fixes #958
Fixes #973 as well
Fixes #974 

This PR provides a new filter panel which:
 - allows multiple selections
 - allows to search on categories
 - clearly shows current selected choice
 - distinction between selected and unselected items
 
 
Multiple selections work for languages (both local and remote) and categories (requires https://github.com/kiwix/libkiwix/pull/974, remote filtering will obviously require that PR to be merged)


juuz0:
@kelson42 I implemented what was needed as described in the ticket, going along with the design guidelines. Would like you to play around with the filters and let me know any changes! Little sneak peek:


https://github.com/kiwix/kiwix-desktop/assets/26814569/85a7ee55-7d52-4b06-8c77-b2380c57eb61

kelson42:
@juuz0 Thank you very much! Will give you a detailed feedback, but I thunk the ultimate solution is more around a custom combobox.

kelson42:
@juuz0 https://github.com/ThisIsClark/Qt-MultiSelectComboBox is the nearest thing I have found in Qt… but its pretty ugly.

juuz0:
> We should be able to see the list of language or categories... so we need a clear way to go through the list

Can you expand more on this please, not sure I get it fully

> I think we should get something like https://harvesthq.github.io/chosen/#multiple-select

This looks good imo, I'll try creating this

> [github.com/ThisIsClark/Qt-MultiSelectComboBox](https://github.com/ThisIsClark/Qt-MultiSelectComboBox) is the nearest thing I have found in Qt… but its pretty ugly.

Yeah definitely similar on principle, although not good looking. Thanks for the reference I'll try replicating the first desing :+1:

kelson42:
> > We should be able to see the list of language or categories... so we need a clear way to go through the list
> 
> Can you expand more on this please, not sure I get it fully

If you have a free text searchbox with no list, you can theoritically get any item of the list... but you can not scroll the whole list.

> > I think we should get something like https://harvesthq.github.io/chosen/#multiple-select

Just to be clear I meant this:
image

> This looks good imo, I'll try creating this

Would be reall awesome to get something like this!!!

kelson42:
@juuz0 What is the status here? This is a very important PR which is open since 2 weeks already.

juuz0:
@kelson42 ready for another review, I have tried replicating that design :>

juuz0:
Did some refactoring

kelson42:
@juuz0 From the way it works, we are indeed near, but then we have design details to reach the usability we have with HTML element given as example. For the moment here are the most important things to fix:
* Please use a https://doc.qt.io/qt-6/qlistview.html, here we have some kind of free text list IMO
* We have to get ride of the search box, and get it working like in the example.

mgautierfr:
I have a segmentation fault on my side with this branch.
No problem on main branch.

juuz0:
> Please use a https://doc.qt.io/qt-6/qlistview.html or similar, here we have some kind of free text list IMO

It's a listwidget so the same thing, but I had hidden the scrollbar so maybe that's why it looked like free text? Scrollbar is back now. Let me know if some other change is needed

> We have to get ride of the search box, and get it working like in the example.

Gone

@kelson42 We go again, new review needed :> I think I'm almost there

image

> I have a segmentation fault on my side with this branch (when starting the application)

I won't say the code is ready for review :P (*shy*). Jokes aside, what OS was this on? Seems to work fine on my Ubuntu. But I'll recheck.

mgautierfr:
> I won't say the code is ready for review :P (shy). Jokes aside, what OS was this on? Seems to work fine on my Ubuntu. But I'll recheck.

Yes. But better to know the sooner. I'm on Fedora 38

kelson42 (changes requested):
@juuz0 Still a lot of differences to https://harvesthq.github.io/chosen/#multiple-select:
* Please remove "All", this is simply the case without any filter
* Really minimal radius in the corners of the boxes
* List appearing only if you search (click)
* Borders in the list
* Proper standard listitem selection (and custom like it curently with only bold)
* Remove of placeholder text when focused/clicked, not when starting to input keyboard

Once this is done we can make a new review round.

juuz0:
@kelson42 new review needed!
I followed all the above points, only point I had a confusion was:
> Proper standard listitem selection (and custom like it curently with only bold)

But check and let me know please!

kelson42 (changes requested):
@juuz0 Indeed, things improve rapidly, keep going please. Here are the next points:
* Corner radius are still too big, make them like in the original for both outer and inner boxes
* Padding/inner margins are far too big, should be only +/- 2 pixels
* When one new item select, the input text should be empty (but vertical bar of search blinking waiting next input)
* Vertical scrollbar does not work (when I click on it the whole list disappear)
* Listbox appearing should be over the other filters (and not moving.translating up and down with the listbox appearing and disappearing)
* Stick two boxes (the selection/search and the listitem) together, no space between each others
* Remove the one pixel left white padding in the listitem on a select item

kelson42 (changes requested):
Need additional cosmetic changes to get it working like the model one

juuz0:
@kelson42 all done, please review again. Code's a mess right now but after design approval, I'll quickly clean it

> Padding/inner margins are far too big, should be only +/- 2 pixels

I've put everything to zero, still there's a bit of space which I can't figure out. If that's acceptable, good. Otherwise, I'll keep checking

image

kelson42 (changes requested):
Next round of improvements:
* Vertical scrollbar still does not work like intended because clicking on top/down arrows closes the overlay. Please just hide/remove this scrollbar, I believe this would be for the better.
* For selected items, the closing logo is actually just a `X` character, please reuse the one given in the model
* The high margins/paddings is a kind of visual and space optimisation problem, I hardly believe you can not do smaller smaller. Maybe you can configure colored backgrounds to better identify if this is padding of the container or margin of the content and therefore better treat the problem?
* Replace "Search ..." placeholder by "Filter languages..." and "Filter categories..."
* The grey for the two placeholders fonts is too dark, take the same as in the example which is lighter
* Change the listitem height, so we never see an item cut vertically in the middle.
* Align all items of the whole sidebar properly on the left (most elements should be more on the left).
* Typing ESC should close the current list (if one has the focus and is open)

It's a lot of small improvements but things are really becoming better and I'm super glad we are already that far.

juuz0:
@kelson42 Done all.
I have put the new filters to Content Type too (latest commit). Hope it works looks/works well or we can remove and revert to old one. Looked weird to have a different type of filter.

This now fixes #974 too and is also ready for code review

kelson42 (changes requested):
At this stage we definitly are polishing the details:
* Make the whole searchbox clickable, currently the focus can be get only by clicking on "Filter language"
* when results are given, we should be able to navigate up/down in the selection with the up/down keyboard arrows, and select one with enter
* I'm glad you have achieve to reduce the space between the box and the content. If I'm not wrong its one pixel now. 2 pixels, like in the model would be better IMO.
* Reduce the space between the three filters so the user see it belongs together
* You don't have fixed the listbox height, and I still see items cut

Once these last details are fixed, I think @mgautierfr can start with code review.

kelson42:
@juuz0 any news?

juuz0:
Done. Sorry for the delay, college got in the last 2 days

kelson42 (changes requested):
Getting better and better but still a new round (sorry):
* Something has changed around the triggering of the online search (online api requests), in comparison to before, it keeps searching online whereas before it was only searching one at start! Without a proper cache strategy this si a problem because everything start to be slow, see #957
* Always the first element of the list should be selected per default, here it seems to be random and to keep track beetween reapparances. Each time the content of the list changes, we need to put the selection on the firsts item (otherwise I have a hard time to navigate with the arrows)
* When I select an item then I get a new search, but if I click on the bottom arrow, the listbox does not appear, only when I click on the input text with the mouse, then the fulll list is displayed
* Please put the freet text "search files" between "local files" and "Languages" filter
* Clicking on the label should not give the focus on the inputbox/listbox. Only clicking on the inputox box should give the focus
* If I click twice on the inputbox then the listbox is redisplayed and flash. It should not, if the listbox is already displayed, nothing should happens.
* Please display the placeholder "Filter ..." only if there is no selection.

I hope that afterthat I won't have new changes to request. I have a few other cosmetics details but not that sure, so will open an other ticket.

kelson42:
@juuz0 any news here?

kelson42:
It keeps getting better but I keep getting things to improve:
* If I put categories to "Wikinews" and "Phets" for online content, I get no result at all, but I should get all wikinews and all phets book. Should be a OR operator between 
* If I select two items, the the box starts to be two lines height big. It should not, it's only when one additional item is selected that we should display on two lines (or more) if not possible to display on one line
* The placeholder does not reappear propelry if I remove all selection of a box. It should appear immediatly as replacment of the last selected item.
* I achieved to get this (should not be possible) Only one item can be select at a time. The handling of selected item is still broken
image 
image

@juuz0 Please test these filter accurately, we have already made a few iterations and what I keep reporting can be found by you at this stage. So please polish everything that it works fine, so we can start with code review.

kelson42 (changes requested):
Better but still slightly buggy

juuz0:
All of this took so much time. Little details were a mess but I tried fixing.

> If I put categories to "Wikinews" and "Phets" for online content, I get no result at all, but I should get all wikinews and all phets book. Should be a OR operator between

Im sure this is because the remote kiwix library is not using the latest libkiwix (or the new changes have not been release yet).

> If I select two items, the the box starts to be two lines height big. It should not, it's only when one additional item is selected that we should display on two lines (or more) if not possible to display on one line

Searcher Space is now based on number of characters entered

> The placeholder does not reappear propelry if I remove all selection of a box. It should appear immediatly as replacment of the last selected item.

Fixed

> I achieved to get this (should not be possible) Only one item can be select at a time. The handling of selected item is still broken

Fixed

Additionally if all items were selected and down arrow was used to navigate, it would go in infinite loop. Fixed

If we chose a language from online library (consider there was no book in local library with that language) and then switched to local library, this would not make sense because that "language" is not possible - 0 books associated in local library. Automatically removed now.

kelson42 (changes requested):
Only a few seconds of testing and here is what I get:
image

Still very buggy:
* No selection
* Widget heigh far to big
* Unable to navigate via keybaord

Please test and fix and test. These problems have been reported first around two weeks ago and are super easy to reproduce. It should have bee fixed meanwhile!

kelson42 (commented):
Made a quick review this morning and seems to work fine. I will do a most in depth last check later but @mgautierfr please go ahead with code review.

src/opdsrequestmanager.cpp
```diff
@@ -23,20 +23,10 @@ void OpdsRequestManager::doUpdate(const QString& currentLanguage, const QString&
...
     }
 
-    // Add "special negative" filter for "other" category (if necessary)
-    if (categoryFilter == "other") {
-        for (auto& category: KiwixApp::instance()->getContentManager()->getCategories()) {
-            if (category != "other" && category != "all") {
-                excludeTags += "_category:"+category;
-            }
-        }
-    }
```
mgautierfr, line 37: Aren't we losing the "other" category ?

src/choiceitem.cpp
```diff
@@ -9,6 +9,10 @@ ChoiceItem::ChoiceItem(QString key, QString value, QWidget *parent) :
     m_value(value)
 {
     ui->setupUi(this);
+    QFile file(QString::fromUtf8(":/css/choiceBox.css"));
+    file.open(QFile::ReadOnly);
+    QString styleSheet = QString(file.readAll());
+    this->setStyleSheet(styleSheet);
```
mgautierfr, line 16: This four lines seems to be copied a lot of time, I wonder if it is time to move them in a helper function.

src/contentmanager.cpp
```diff
@@ -617,26 +618,30 @@ QStringList ContentManager::getDownloadIds()
...
+    KiwixApp::instance()->getSettingsManager()->setLanguage(languageList);
     emit(currentLangChanged());
     emit(filterParamsChanged());
 }
 
-void ContentManager::setCurrentCategoryFilter(QString category)
+void ContentManager::setCurrentCategoryFilter(QStringList categoryList)
 {
-    if (m_categoryFilter == category)
+    if (m_categoryFilter == categoryList.join(","))
```
mgautierfr, line 640: We should be prepared for list containing the same categories but in different orders.

src/contentmanager.cpp
```diff
@@ -617,26 +618,30 @@ QStringList ContentManager::getDownloadIds()
...
+    for (auto &language : languageList) {
+        if (language.length() == 2) {
+          try {
+            language = QString::fromStdString(
+                         kiwix::converta2toa3(language.toStdString()));
+          } catch (std::out_of_range&) {}
+        }
     }
-    if (m_currentLanguage == language)
-        return;
```
mgautierfr, line 630: We lost the early exist if the languages is not actually changed.

src/kiwixchoicebox.cpp
```diff
@@ -171,28 +171,17 @@ void KiwixChoiceBox::setSelections(QStringList selections, QString defaultSelect
...
 {
     choiceSelector->clear();
     for (const auto &selection: selections)
     {
         auto item = new KListWidgetItem(beautifyString(selection.second));
         item->setData(Qt::UserRole, selection.first);
         choiceSelector->addItem(item);
-        if (selection.second ==  defaultSelection)
-        {
-            m_defaultItem = item;
```
mgautierfr: `m_defaultItem` is not changed anymore.
Either we remove it from the class or we have to handle it somehow.

resources/i18n/en.json
```diff
@@ -156,6 +156,10 @@
     "move-files-to-trash-text": "This action will move the file to trash.",
     "perma-delete-files-text": "This action will permanently delete the file.",
     "clear-filter": "Clear the currently set filters",
-    "search-language": "Search language",
-    "search-category": "Search category"
+    "choose-language": "Filter language",
+    "choose-category": "Filter category",
+    "choose-content-type": "Filter content type",
```
mgautierfr, line 161: As this is used as placeholder, I would name them `language-searcher-placeholder`

src/kiwixlistwidget.cpp
```diff
@@ -0,0 +1,113 @@
...
+    : QListWidget(parent), currRow(0), m_visibleItems(count())
+{
+    connect(this, &KiwixListWidget::currRowChanged, this, &KiwixListWidget::handleCurrRowChange);
+    setMouseTracking(true);
+}
+
+void KiwixListWidget::showEvent(QShowEvent *event)
+{
+    QListWidget::showEvent(event);
+}
```
mgautierfr, line 19: Is this really necessary ?

src/kiwixlistwidget.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+    if (selectedItems().size() == count())
+        return;
+    KListWidgetItem *currItem = dynamic_cast<KListWidgetItem*>(item(currRow));
+    int oldRow = currRow;
+    do {
+        currRow--;
+        if (currRow < 0) currRow = count() - 1;
+        currItem = dynamic_cast<KListWidgetItem*>(item(currRow));
+    } while (currItem->isSelected() || currItem->isHidden());
```
mgautierfr, line 82: Why we need a loop here ?
In which context we want to move more than one step above ?

src/contentmanagerside.ui
```diff
@@ -109,6 +106,9 @@
          </property>
         </widget>
        </item>
+       <item>
```
mgautierfr, line 109: This comment is about the commit message.
`Moved as per suggestion` is definitively not a good commit message.
It is nice to discard you about the choice made but it will not help us to understand the choice in few months.
The reason why you made this change is not "It has been suggested to me".
The suggestion come for a reason (and you have agree with it, else you would not do the change), it is better to directly write the reason.

src/kiwixchoicebox.cpp
```diff
@@ -112,18 +105,40 @@ KiwixChoiceBox::~KiwixChoiceBox()
     delete ui;
 }
 
-void KiwixChoiceBox::mouseReleaseEvent(QMouseEvent *event)
+void KiwixChoiceBox::mousePressEvent(QMouseEvent *event)
```
mgautierfr, line 116: Almost (all?) toolkit generate a click event on `release` mouse button. Why changing this ?

src/contentmanagerside.cpp
```diff
@@ -80,18 +80,12 @@ void ContentManagerSide::setContentManager(ContentManager *contentManager)
     mp_contentManager = contentManager;
     connect(mp_languages, &KiwixChoiceBox::choiceUpdated,
             this, [=](QStringList values) {
-                if (values[0] == "all") {
-                    values = QStringList();
-                }
```
mgautierfr: All the things about the `all` or `*` is important. It would be better to have a specific commit instead of hidding it behind a "Also made a few structural changes making everything consistent."

src/kiwixchoicebox.cpp
```diff
@@ -238,20 +240,25 @@ void KiwixChoiceBox::setSelections(QStringList selections, QStringList defaultSe
...
+    clearSelections();
     choiceSelector->clear();
     for (const auto &selection: selections)
     {
         auto item = new KListWidgetItem(beautifyString(selection.second));
         item->setData(Qt::UserRole, selection.first);
         choiceSelector->addItem(item);
         if (defaultSelection.contains(selection.first)) {
-            addSelection(item);
+            defaultSelection.removeOne(selection.first);
```
mgautierfr, line 251: I put a comment here because it is the only line I don't know the exact purpose but it is for the whole commit.

The commit is about removing item widget if corresponding choice is not valid (switch to local library with less choices). But how do you check the choice is valid ? You seem to never check the choice against the actual library displayed.

mgautierfr (changes requested):

src/kiwixchoicebox.cpp
```diff
@@ -0,0 +1,221 @@
...
+        choiceSelector->setVisible(false);
+        ui->currentChoices->setStyleSheet("#currentChoices{border: 1px solid #ccc;}");
+    });
+
+    connect(searcher, &KiwixLineEdit::focusedIn, [=]() {
+        ui->currentChoices->setStyleSheet("#currentChoices{border: 2px solid #4e63ad;}");
+        adjustSize();
+        choiceSelector->setVisible(true);
+        choiceSelector->raise();
+        searcher->setPlaceholderText("");
```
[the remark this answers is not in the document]
juuz0: Yes we need to, when there are some items selected and the search is empty, it should not show up

src/kiwixchoicebox.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (m_choiceItems.contains(key))
+        return false;
+    auto chItem = new ChoiceItem(key, value);
+    connect(chItem, &ChoiceItem::closeButtonClicked, [=](QString text) {
+        removeSelection(text);
+    });
+    m_choiceItems.insert(key, chItem);
+    currentChoicesLayout->removeWidget(searcher);
+    currentChoicesLayout->addWidget(chItem);
+    currentChoicesLayout->addWidget(searcher);
```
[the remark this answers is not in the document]
juuz0: Searcher needs to be the last widget always, so i remove it and add again after adding a new widget. I could've use `insertWidget` too, was that better?

[Widget] [Widget] [Searcher]

*new widget is added*

[Widget] [Widget] [New Widget] [Searcher]

src/kiwixchoicebox.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+void KiwixChoiceBox::adjustSize()
+{
+    QWidget::adjustSize();
+    choiceSelector->adjustSize();
+    choiceSelector->setGeometry(this->x() + ui->currentChoices->x(),
+                                this->y() + ui->currentChoices->y() + ui->currentChoices->height(),
+                                choiceSelector->width(),
+                                choiceSelector->count() * KListWidgetItem::getItemHeight() + 2);
```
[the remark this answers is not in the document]
juuz0: Yes, it's for a little border/margin. Will add a comment

src/kiwixchoicebox.cpp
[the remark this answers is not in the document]
mgautierfr: `insertWidget` is more consistent (at least it express what you intend. We don't want to remove the searcher, so let's not remove it)

mgautierfr:
The category filter use a `AND` operator. We should use a `OR` (as for language)

The query is `https://library.kiwix.org:443/catalog/search?lang=eng,fra&count=-1&category=iFixit,other" when we select English and French languages and iFixit and Other categories.
For me it is a libkiwix issue (to probably fix in https://github.com/kiwix/kiwix-tools/issues/440) but while it is not fixed, allowing user to select several categories is a bug (zim files have only one category, so a and on two categories will always return zero book)

juuz0:
> The category filter use a AND operator. We should use a OR (as for language)

Are you sure? I did recheck it was an OR operator when i added it here https://github.com/kiwix/libkiwix/pull/974

I think it gives the illusion of being an AND operator because the current library.kiwix.org is not using nightly libkiwix. 
the same query works on the dev library: https://dev.library.kiwix.org/catalog/search?lang=eng,fra&count=-1&category=iFixit,other

mgautierfr:
Indeed, my bad.

src/flowlayout.cpp
```diff
@@ -0,0 +1,221 @@
...
+**
+** "Redistribution and use in source and binary forms, with or without
+** modification, are permitted provided that the following conditions are
+** met:
+**   * Redistributions of source code must retain the above copyright
+**     notice, this list of conditions and the following disclaimer.
+**   * Redistributions in binary form must reproduce the above copyright
+**     notice, this list of conditions and the following disclaimer in
+**     the documentation and/or other materials provided with the
+**     distribution.
```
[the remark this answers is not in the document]
kelson42: I'm concerned about that....

src/kiwixchoicebox.cpp
```diff
@@ -0,0 +1,221 @@
...
+        adjustSize();
+        choiceSelector->setVisible(true);
+        choiceSelector->raise();
+        searcher->setPlaceholderText("");
+    });
+
+    ui->clearButton->setCursor(Qt::PointingHandCursor);
+    connect(ui->clearButton, &QPushButton::clicked, [=]() {
+       clearSelections();
+       emit(choiceUpdated(getCurrentSelected()));
```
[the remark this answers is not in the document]
juuz0: I thought `clearSelections()` as just a utlity to remove everything (also used in `setSelections`) so went against putting a signal. Haven't added it yet but i'll do it if its needed/better structure

src/kiwixchoicebox.cpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    choiceSelector->clear();
+    for (const auto &selection: selections)
+    {
+        auto item = new KListWidgetItem(beautifyString(selection.second));
+        item->setData(Qt::UserRole, selection.first);
+        choiceSelector->addItem(item);
+        if (selection.second ==  defaultSelection)
+        {
+            m_defaultItem = item;
```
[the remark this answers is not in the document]
juuz0: In future commits, I did away from using m_defaultItem so this was removed

src/kiwixchoicebox.h
```diff
@@ -0,0 +1,59 @@
...
+signals:
+    void choiceUpdated(QStringList);
+
+private:
+    Ui::kiwixchoicebox *ui;
+    QLabel *choiceLabel;
+    QLineEdit *choiceSearch;
+    QListWidget *choiceSelector;
+    QListWidgetItem *m_defaultItem;
+    QMap<QString, ChoiceItem*> m_choiceItems;
```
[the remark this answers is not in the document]
juuz0: Honestly, right. thanks. It felt messy and duplicate - my initial thought was keeping a map so I can check selected items and remove/delete them easily.
I've gone and completely removed `m_choiceItems` and using just the `choiceSelector` and `ui->currentChoices` instead - using object names. Hope this looks better?

src/contentmanager.cpp
juuz0: list is sorted before setting values now. So if there are same values, the order should be same and it equality check will work

src/kiwixlistwidget.cpp
juuz0: Wasn't, was a consequence of overlooking after a rebase

src/contentmanagerside.ui
juuz0: Done.

src/contentmanagerside.cpp
juuz0: Done

src/kiwixchoicebox.cpp
juuz0: Yep, I went big brain on this.

`defaultSelection` will contain the current values set (in settings).
What we know is the elements in `selections` should contain all of elements in `defaultSelection`. so whenever we get that an element exists in both, we remove that element from `defaultSelections`.

In the end, if there's any element left in `defaultSelection`,  it means that isnt available in `selections` (current list of languages/categories) - it must've come from online library.

At this point, sending the signal to update choices with new values will do 2 things:
1. refresh library with our choices
2. set the correct filter in settings (that was common to `defaultSelection` and `selections`)

src/kiwixlistwidget.cpp
juuz0: When the searcher is used with some term - only those elements are shown which match that term (others are not removed but hidden) so while navigating with keyboard to get to the actual element i need to skip the hidden ones

src/flowlayout.cpp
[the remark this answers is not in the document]
juuz0: It's a pretty good example layout from QT - better than reinventing the wheel. I understand if adding this is a problem though but it's BSD licensed so maybe it's permissive enough?

src/opdsrequestmanager.cpp
juuz0: Is "other" not a real category?

mgautierfr: Indeed. We had to implement "other" on kiwix-desktop side has we were using a fixed list of categories. But now "other" is a real category.

src/kiwixchoicebox.cpp
juuz0: When the `lineEdit` is currently focused,
if the outer widget is *pressed*, `lineEdit` loses focus and hides the options.
If I use `mouseRelease`, it will create a flicker effect: Pressing causes the `lineEdit` to lose focus and hide the options - Release causes the options to show up again.

showing the options on a `mousePress` doesn't allow this

juuz0:
@mgautierfr another review stage please :)

src/kiwixchoicebox.cpp
```diff
@@ -97,7 +97,7 @@ KiwixChoiceBox::KiwixChoiceBox(QWidget *parent) :
     });
 
     connect(this, &KiwixChoiceBox::choiceUpdated, [=]() {
-        if (!m_choiceItems.size())
```
mgautierfr: `m_choiceItems` is never used now. We should remove it from the class.

src/kiwixchoicebox.cpp
```diff
@@ -186,11 +182,8 @@ bool KiwixChoiceBox::addSelection(QListWidgetItem *item, bool updateRequired)
 
 bool KiwixChoiceBox::removeSelection(QListWidgetItem *item)
 {
-    auto selection = item->text();
-    if (!m_choiceItems.contains(selection)) return false;
```
mgautierfr: Shouldn't we handle the case when `ui->currentChoices` doesn't contain `item->text()` ?

src/kiwixchoicebox.cpp
```diff
@@ -204,10 +197,9 @@ void KiwixChoiceBox::clearSelections()
 {
     for (auto &item : choiceSelector->selectedItems()) {
         item->setSelected(false);
-        auto chItem  = m_choiceItems[item->text()];
-        currentChoicesLayout->removeWidget(chItem);
-        chItem->deleteLater();
-        m_choiceItems.remove(item->text());
+        auto chItem  = ui->currentChoices->findChild<ChoiceItem*>(item->text());
+        ui->currentChoices->layout()->removeWidget(chItem);
+        delete chItem;
```
mgautierfr, line 199: `chItem->deleteLater()` ?

src/kiwixchoicebox.h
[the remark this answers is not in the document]
mgautierfr: It is

src/kiwixchoicebox.cpp
mgautierfr: Please add this explanation as a comment in the source code.

src/kiwixlistwidget.cpp
mgautierfr: Ok

src/kiwixchoicebox.cpp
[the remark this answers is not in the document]
mgautierfr: Ok

src/kiwixchoicebox.cpp
mgautierfr: > Yep, I went big brain on this.

Going big brain is a tricky way :)
If your reviewer don't understand what you do, you can be sure that you will not be able to understand it if few months. (And if you cannot, no one will do)

Why not going the basic way ?

---

In fact, I even wonder if the commit is useful.
If I have selected `wikipedia` category when browsing a remote library, I make sens to still have the filtering when a go to the local library, even if I don't have local wikipedia books.
And if we remove it, when I go back to the remote library, my filters will be changed automatically (and so I loose theme)

From a UI perspective, I would say that we clear all filters or none when we move from a library to the other.

mgautierfr (changes requested):
Few small changes to do on the code side.

But one open question about the feature of removing filters when we change viewed library.

src/kiwixchoicebox.cpp
juuz0: I had started with a `deleteLater` early on but seems like it doesn't delete just then. Later on, adding selection creates duplicates if the old aren't deleted, hence the `delete` usage.
It's also not inside a slot and sort of "independent", so I think it's safe to use delete.

src/kiwixchoicebox.cpp
juuz0: Thank you, I agree about the tricky way :)
I went with keeping the old list of values when changing catalog. https://github.com/kiwix/kiwix-desktop/pull/965/commits/df9873feb5dc8eaf9c45c10dc16c895a71e57292

I think another way could be that we keep the settings as a map of values like
{
English: eng
French: fra
}

mgautierfr (approved):
LGTM !!

I have few graphical issues but I will open new issues on github.
~~This PR is long enough, it is time to merge it.~~
Those issues are not related to this PR, so we don't care.
We only have my last comment left.

mgautierfr:
Hum, I have a real issue after all with the last change about keeping the filter.

- Go to remote library and select a language filter for a language you don't have in local.
- Go to local library, the filter is keep and displayed (we are good here).
- Quit kiwix-desktop and relaunch it.
- The language filter is applied but not displayed in the filter (we are bad)
- Go to remote library, the filter is displayed
- Go back to local library, the filter is displayed.

I would say we can live with this (as we will fix soon with another issue) but @kelson42 may think it is a important bug to fix before merging.

kelson42:
definitly, real filter should be displayed filter. We have a last thing to fix here.

src/contentmanager.cpp
```diff
@@ -619,8 +619,12 @@ QStringList ContentManager::getDownloadIds()
     return list;
 }
 
-void ContentManager::setCurrentLanguage(QStringList languageList)
+void ContentManager::setCurrentLanguage(QList<QPair<QString, QString>> langPairList)
```
mgautierfr, line 622: This `QList<QPair<QString, QString>>` is a bit ugly (and everywhere).
We should rename it to something more meaningful. I suggest would `FilterList` (until I realize we have already `SelectionList` which does the same 😄 )

But we can keep a different name to help us remember that it may be different kind of things:
- FilterList : A list of (potential) filters.
- SelectionList: A list of selected filters.

mgautierfr (changes requested):
Just a small change about the naming and we are good.

src/contentmanager.cpp
juuz0: Done :>

mgautierfr (approved):

mgautierfr:
We are good. Time to merge 🎉 
Congrats on this difficult PR.

PR #969, opened 2023-07-31
Update CI to use the last version of docker image and dependencies.

mgautierfr:


kelson42:
@mgautierfr Why do we need to come back to bionic, considering that focal is the new oldest version we support?

mgautierfr:
I've mixed both version. And there is no need for this PR.

PR #970, opened 2023-08-02
Open zim location in file manager

### LINKED ISSUE #98: Add "Open folder" command to right-click menu
Windows/beta2: it would be nice to be able to figure out where previously downloaded and stored zim files are located. In my particular use case there's been several downloaded over time in different folders, and I have no idea where since Kiwix only lists these files in its library without telling me where they are.

### LINKED ISSUE #971: App crash on mouse right click in local library
Specificaly in the white zone at the bottom at of the screenshot (without any book)
image

Clearly a regression (at least in `openFolder` git branch)

juuz0:
Fixes #98 
Fixes #971

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -21,6 +21,7 @@
 #include "kiwixconfirmbox.h"
 #include <QtConcurrent/QtConcurrentRun>
 #include "contentmanagerheader.h"
+#include <QDesktopServices>
 
 ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader, QObject *parent)
     : QObject(parent),
@@ -99,6 +100,7 @@ void ContentManager::onCustomContextMenu(const QPoint &point)
     QAction menuPauseBook(gt("pause-download"), this);
     QAction menuResumeBook(gt("resume-download"), this);
     QAction menuCancelBook(gt("cancel-download"), this);
+    QAction menuOpenFolder(gt("open-folder"), this);
 
     if (bookNode->isDownloading()) {
         if (bookNode->getDownloadInfo().paused) {
@@ -110,8 +112,24 @@ void ContentManager::onCustomContextMenu(const QPoint &point)
     } else {
         try {
             const auto book = KiwixApp::instance()->getLibrary()->getBookById(id);
+            auto bookPath = QString::fromStdString(book.getPath());
             contextMenu.addAction(&menuOpenBook);
             contextMenu.addAction(&menuDeleteBook);
+            contextMenu.addAction(&menuOpenFolder);
+            connect(&menuOpenFolder, &QAction::triggered, [=]() {
+                QFileInfo fileInfo(bookPath);
+                QDir bookDir = fileInfo.absoluteDir();
+                bool dirOpen = bookDir.exists() && bookDir.isReadable() && QDesktopServices::openUrl(bookDir.absolutePath());
+                if (!dirOpen) {
+                    QString failedText = gt("couldnt-open-location-text");
+                    failedText = failedText.replace("{{FOLDER}}", "<b>" + bookDir.absolutePath() + "</b>");
+                    KiwixConfirmBox *dialog = new KiwixConfirmBox(gt("couldnt-open-location"), failedText, true, mp_view);
+                    dialog->show();
+                    connect(dialog, &KiwixConfirmBox::okClicked, [=]() {
+                        dialog->deleteLater();
+                    });
+                }
+            });
         } catch (...) {
             contextMenu.addAction(&menuDownloadBook);
         }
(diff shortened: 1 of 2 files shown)


kelson42:
@juuz0 Thx, will review later today!

src/contentmanager.cpp
```diff
@@ -132,6 +137,12 @@ void ContentManager::onCustomContextMenu(const QPoint &point)
     connect(&menuResumeBook, &QAction::triggered, [=]() {
         resumeBook(id, index);
     });
+    connect(&menuOpenFolder, &QAction::triggered, [=]() {
+        if (bookPath != "") {
+            QFileInfo fileInfo(bookPath);
+            QDesktopServices::openUrl(fileInfo.absoluteDir().absolutePath());
+        }
+    });
```
mgautierfr, line 145: It is better to setup the action in the "try/catch" instead of creating a "global" `bookPath` and setting if `if` in the action.

kelson42:
I will check and merge, but clearly the edge case to test is th handling of exception if the folder is not redable or not available.

kelson42 (changes requested):
Seems to work, but we should have a proper error message if not possible to open a folder. Something like "Kiwix is not able to open folder xyz`.

juuz0:
> Something like "Kiwix is not able to open folder xyz`.

Done. Added checks if directory exists/isReadable

src/contentmanager.cpp
```diff
@@ -108,12 +112,29 @@ void ContentManager::onCustomContextMenu(const QPoint &point)
...
+            auto bookPath = QString::fromStdString(book.getPath());
             contextMenu.addAction(&menuOpenBook);
             contextMenu.addAction(&menuDeleteBook);
-        }
-        else
+            contextMenu.addAction(&menuOpenFolder);
+            connect(&menuOpenFolder, &QAction::triggered, [=]() {
+                QFileInfo fileInfo(bookPath);
+                QDir bookDir = fileInfo.absoluteDir();
+                bool dirOpen = bookDir.exists() && bookDir.isReadable() && QDesktopServices::openUrl(bookDir.absolutePath());
```
kelson42, line 124: @juuz0 Here is quite comment code pattern and let me take a few minutes to make an in-depth analysis.

You have two different things you might be interested to know:

1. * Does the directory can be "open" (boolean answer)
2. * Maybe, if it fails: Why it fails to open

My request was only about (1) and to do that you should only rely on `QDesktopServices::openUrl` IMO.

Then, you might want to do (2) to help the user with a meaningful message (why the folder can't actually be open), but here you don't as your message is always the same.

Would you be able to implement it that way, or do the multiple checks on line 124 are here to compensate some kind of weakness of the `QDesktopServices::openUrl` (for example it open even if the directory is not readable?)

juuz0: > do the multiple checks on line 124 are here to compensate..........

@kelson42 
Yes, this is an issue with [`QDesktopServices::openUrl`](https://doc.qt.io/qt-6/qdesktopservices.html#openUrl) 
From the article:
> Warning: A return value of true indicates that the application has successfully requested the operating system to open the URL in an external application. The external application may still fail to launch or fail to open the requested URL. This result will not be reported back to the application.

Should put a better message yeah, should these be enough?

1. Kiwix is not able to open the location because the folder doesn't exist
2. Kiwix is not able to open the location because the folder is not readable
3. Kiwix is not able to open the location [Generic]

juuz0: > Could we let the file explorer display a message to the user (it is better placed to know why it has fail to open the directory and so gives a useful message).

This works for cases where directory is not readable.
But atleast on Ubuntu, if the directory doesn't exist - file explorer wont launch but show `gio: No such file or directory.` in the terminal. 
Most people won't have kiwix-desktop running through a terminal so atleast for the case where `directory doesn't exist`, its better to show the dialog in kiwix-desktop?

mgautierfr: Is it possible that we have a zim file displayed in the library view when the containing directory doesn't exist ?

mgautierfr: I doubt it.

ContentManager use a filter with local and valid (https://github.com/kiwix/kiwix-desktop/blob/main/src/contentmanager.cpp#L692-L694).
And valid means that we can access the file : (I spare you all the code, but at this end it is : https://github.com/kiwix/libkiwix/blob/main/src/tools/pathTools.cpp#L314-L321)

This could be possible if the user remove the directory after loading the library. We should be prepare for this (not crashing) but I'm not sure we have to do more.
And if we have to do more, we should probably doing it by detecting this sooner and "greying" the book in the view and not display a "open in directory".

kelson42: > @kelson42 
> Yes, this is an issue with [`QDesktopServices::openUrl`](https://doc.qt.io/qt-6/qdesktopservices.html#openUrl) 

Then fine to me. Current approach is probably the best compromise. Most important is not to fail silently.

kelson42 (approved):

PR #981, opened 2023-08-08
Fix CloseTab related weakness

### LINKED ISSUE #968: Shortcuts `ctrl+f4` and `ctrl+w` should close current tab
The shortcut `ctrl + f4` only works immediately after a tab has been clicked. What I expect is that that shortcut always closes the current tab regardless of what has been clicked or what is in focus like in any other browser e.g. chrome or firefox. Furthermore, any other browser support `ctrl + w` along with `ctrl + f4` to close the current tab.

Additional suggestion: Add `ctrl + w` to close the current tab.

Currently, I tested this issue on kiwix-desktop 2.3.1 on both windows 7 and 10.

juuz0:
Fixes #968

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -296,6 +296,8 @@ void KiwixApp::setMonitorDir(const QString &dir) {
 
 #define CREATE_ACTION(ID, TEXT) \
     mpa_actions[ID] = new QAction(TEXT)
+#define CREATE_ACTION_ICON(ID, ICON, TEXT) \
+    mpa_actions[ID] = new QAction(QIcon(":/icons/" ICON ".svg"), TEXT);
 #define SET_SHORTCUT(ID, TEXT, SHORTCUT) \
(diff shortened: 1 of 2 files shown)


kelson42:
@juuz0 IMO, it shpuld mot be "in place", but "in addition".

juuz0:
>  IMO, it should mot be "in place", but "in addition".

 Do you mean the shortcuts?
Sadly if I use `QKeySequence::close` in addition to these two, QT consider that an ambiguity so i explicitly mentioned both shortcuts.

kelson42 (approved):
Seems to work properly now and principle LGTM. I would still put a comment explaining about the usage of `CREATE_ACTION_ICON_SHORTCUTS` why we don't use anymore `QKeySequence::Close`.

mgautierfr (approved):

mgautierfr:
>  why we don't use anymore QKeySequence::Close.

This is explain in the commit message.

PR #983, opened 2023-08-11
Grammar update for "About Kiwix" box

psypherium:


mgautierfr:
Thanks for you PR @psypherium 

I agree with your proposed changes. However, can you squash your commits to have only one commit in the PR ?
Rebasing (not merging) your `master` branch on top of our `main` branch should do the work.

psypherium:
I still don't know much about git, sorry.

I can do that but it'll take a few minutes, I'll have to set up github desktop and do some stuff there, there isn't an option for any of that in the web interface.

Would the "squash and merge" or "rebase and merge" github.com options not give the desired result here? Or does this repo not allow those functions?

mgautierfr:
"Squash en merge" will indeed squash all your commits into one, but it will create only one commit in master.
"Rebase and merge" is not available (probably because the branch is in your repository)

We want to keep a specific merge commit (to track which commits are part of which PR) with a "clean" git history.
(You can see the result here : https://github.com/kiwix/kiwix-desktop/network . The main branch is only composed of merge commits. And each branch is "small" and never "crossing" each other)

kelson42:
I have fixed the git revision log in #986, closing this one.

PR #989, opened 2023-08-17
On Startup: switch to online files if there are no local files

### LINKED ISSUE #978: If not local book, switch to online library (default behaviour)
Per default, we show the local library. Seems appropriate. But if there is nothing in the local library, then the user has to figure out what to do next.

This is particulary blattent at the first start.

Therefore I propose to show per default online library in place of local library if nothing is in it.

juuz0:
Fixes #978

### DIFF
--- a/src/contentmanager.h
+++ b/src/contentmanager.h
@@ -16,6 +16,7 @@ class ContentManager : public QObject
     Q_PROPERTY(QStringList bookIds READ getBookIds NOTIFY booksChanged)
     Q_PROPERTY(QStringList downloadIds READ getDownloadIds NOTIFY downloadsChanged)
     Q_PROPERTY(QString currentLanguage MEMBER m_currentLanguage WRITE setCurrentLanguage NOTIFY currentLangChanged)
+    Q_PROPERTY(bool isLocal MEMBER m_local READ isLocal WRITE setLocal NOTIFY localChanged)
 
 public:
     typedef QList<QPair<QString, QString>> LanguageList;
@@ -66,6 +67,7 @@ signals:
     void pendingRequest(const bool);
     void categoriesLoaded(QStringList);
     void languagesLoaded(LanguageList);
+    void localChanged(const bool);
 
 public slots:
     QStringList getTranslations(const QStringList &keys);
--- a/src/contentmanagerside.cpp
+++ b/src/contentmanagerside.cpp
@@ -12,17 +12,19 @@ ContentManagerSide::ContentManagerSide(QWidget *parent) :
     mp_ui(new Ui::contentmanagerside)
 {
     mp_ui->setupUi(this);
-    connect(mp_ui->allFileButton, &QRadioButton::toggled,
-            this, [=](bool checked) { this->mp_contentManager->setLocal(!checked); });
-    connect(mp_ui->localFileButton, &QRadioButton::toggled,
-            this, [=](bool checked) { this->mp_contentManager->setLocal(checked); });
+    mp_ui->buttonGroup->setId(mp_ui->allFileButton, CatalogButtonId::ALL);
+    mp_ui->buttonGroup->setId(mp_ui->localFileButton, CatalogButtonId::LOCAL);
+    connect(mp_ui->buttonGroup, QOverload<QAbstractButton *>::of(&QButtonGroup::buttonClicked), [=](QAbstractButton *btn) {
+        const auto id = mp_ui->buttonGroup->id(btn);
+        mp_contentManager->setLocal(id == CatalogButtonId::LOCAL);
+    });
+
     connect(mp_ui->allFileButton, &QRadioButton::toggled,
             this, [=](bool checked) { mp_ui->allFileButton->setStyleSheet(
                     checked ? "*{font-weight: bold}" : "");});
     connect(mp_ui->localFileButton, &QRadioButton::toggled,
             this, [=](bool checked) { mp_ui->localFileButton->setStyleSheet(
                     checked ?"*{font-weight: bold}" : "");});
-    mp_ui->localFileButton->setStyleSheet("*{font-weight: bold}");
 
     mp_ui->allFileButton->setText(gt("online-files"));
     mp_ui->localFileButton ->setText(gt("local-files"));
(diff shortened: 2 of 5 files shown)


juuz0:
I wonder if we should show a notice box "No local ZIM files detected, would you like to switch to remote catalog?" on startup before switching

kelson42:
> I wonder if we should show a notice box "No local ZIM files detected, would you like to switch to remote catalog?" on startup before switching

I don't think so, it's just the default behaviour. Actually it should not even be a switch, it should go online in place of local.

kelson42 (approved):
From a user perspective, seems to do the job

src/contentmanager.cpp
```diff
@@ -61,6 +61,12 @@ ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader,
     connect(this, &ContentManager::filterParamsChanged, this, &ContentManager::updateLibrary);
     connect(this, &ContentManager::booksChanged, this, [=]() {
         const auto nBookList = getBooksList();
+        std::call_once(zeroBooksInLocalFlag, [=]() {
+            if (nBookList.size() == 0) {
+                setLocal(false);
+                emit(catalogChanged(isLocal()));
+            }
+        });
```
mgautierfr, line 69: It seems a bit odd to have a call_once in a action being triggered a lot of time.
Setting the catalog local or remote is part of initialization and it should be put here.
Probably just after we load the local library at init.

src/contentmanagerside.cpp
```diff
@@ -123,6 +123,13 @@ void ContentManagerSide::setContentManager(ContentManager *contentManager)
                 auto category = item->data(Qt::UserRole).toString();
                 mp_contentManager->setCurrentCategoryFilter(category);
     });
+    connect(this->mp_contentManager, &ContentManager::catalogChanged, [=](bool isLocal) {
+        if (isLocal) {
+            mp_ui->localFileButton->setChecked(true);
+        } else {
+            mp_ui->allFileButton->setChecked(true);
+        }
+    });
```
mgautierfr, line 132: This is a bit odd also.
Buttons should follow the "value of isLocal". It should not be to contentManagerSide to change their state.
Even more, `setChecked`  will trigger [`toggled` event](https://doc.qt.io/qt-6/qabstractbutton.html#toggled) which will in turn change `isLocal` (https://github.com/kiwix/kiwix-desktop/blob/main/src/contentmanagerside.cpp#L15-L18)

Maybe `isLocal` should be a property (with correcty signal/slot associated) and both contentmanager and buttons should react to them.
[QButtonGroup](https://doc.qt.io/qt-6/qbuttongroup.html) may help here for the implementation.

mgautierfr (changes requested):

kelson42:
@juuz0 Any news here?

src/contentmanager.cpp
juuz0: Done.

src/contentmanagerside.cpp
juuz0: I am not exactly sure how I can use "value of isLocal" without a signal and setting state. I've used buttongroup now but hardly changes anything.

src/contentmanager.h
```diff
@@ -66,6 +67,7 @@ class ContentManager : public QObject
     void pendingRequest(const bool);
     void categoriesLoaded(QStringList);
     void languagesLoaded(LanguageList);
+    void catalogChanged(const bool);
```
mgautierfr, line 70: Maybe better to call it `localChanged` as it is about `isLocal`

src/contentmanagerside.cpp
```diff
@@ -105,6 +107,10 @@ ContentManagerSide::~ContentManagerSide()
 void ContentManagerSide::setContentManager(ContentManager *contentManager)
 {
     mp_contentManager = contentManager;
+    const auto isLocal = mp_contentManager->isLocal();
+    const auto checkedButton = mp_ui->buttonGroup->button(isLocal);
```
mgautierfr, line 111: Same here :
`const auto checkedButton = mp_ui->buttonGroup->button((int)isLocal);`

(But I'm not sure it works with a enum, at least maybe not as simple)

src/kiwixapp.cpp
```diff
@@ -64,6 +64,7 @@ void KiwixApp::init()
         gt("error-downloader-launch-message") + "<br><br>" + e.what());
     }
     mp_manager = new ContentManager(&m_library, mp_downloader);
+    mp_manager->setLocal(m_library.getBookIds().size());
```
mgautierfr, line 67: Same here : `mp_manager->setLocal(!m_library.getBookIds().isEmpty());`

src/contentmanagerside.cpp
```diff
@@ -12,17 +12,19 @@ ContentManagerSide::ContentManagerSide(QWidget *parent) :
     mp_ui(new Ui::contentmanagerside)
 {
     mp_ui->setupUi(this);
-    connect(mp_ui->allFileButton, &QRadioButton::toggled,
-            this, [=](bool checked) { this->mp_contentManager->setLocal(!checked); });
-    connect(mp_ui->localFileButton, &QRadioButton::toggled,
-            this, [=](bool checked) { this->mp_contentManager->setLocal(checked); });
+    mp_ui->buttonGroup->setId(mp_ui->allFileButton, 0);
+    mp_ui->buttonGroup->setId(mp_ui->localFileButton, 1);
```
mgautierfr, line 16: Can't we set this in the ui definition  ?

juuz0: Sadly, no.

src/contentmanagerside.cpp
```diff
@@ -12,17 +12,19 @@ ContentManagerSide::ContentManagerSide(QWidget *parent) :
...
+    connect(mp_ui->buttonGroup, QOverload<QAbstractButton *>::of(&QButtonGroup::buttonClicked), [=](QAbstractButton *btn) {
+        const auto id = mp_ui->buttonGroup->id(btn);
+        mp_contentManager->setLocal(id);
```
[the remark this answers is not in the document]
juuz0: I have used the normal `enums` instead of `enum class` to work implicitly with setId's int param value here (requires static casting otherwise)

juuz0:
@mgautierfr done!

PR #992, opened 2023-08-23
Adapt to new api introduced in kiwix/libkiwix#991

mgautierfr:
This is the following of https://github.com/kiwix/libkiwix/pull/991

`kiwix::Library` must be handle through a `shared_ptr` and cannot be created directly on the stack.


mgautierfr:
This PR is needed as it fixes the CI which is broken since several days. I'm merging.
Please open a issue if you disagree with the code :)

PR #1024, opened 2023-12-14
A couple of bugfixes in Download management

### LINKED ISSUE #1021: Download progress status disappears when filtering is applied
1. Start a download large enough, so that you have time to perform the next steps before it is complete
2. (optional) Pause the download
3. Apply filtering (so that the file being downloaded is included in the results)
4. See the download progress info and/or pause/resume button disappear. Clearing/resetting filtering doesn't restore the ability to see and manage the progress of incomplete downloads.

### LINKED ISSUE #1023: Download progress indicator is incorrect for files larger than 2 GiB
See title

veloman-yunkan:
Fixes #1021
Fixes #1023

This PR also contains some attempts at cleaning up the code, however the conclusion is that a major overhaul is required.

Overall it looks like there are many more bugs related to Download management, and the proper way to eliminate them all is to significantly simplify the code architecture through a more direct usage of `Kiwix::Download`.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -137,7 +137,7 @@
     "monitor-directory-tooltip":"All ZIM files in this directory will be automatically added to the library.",
     "next-tab":"Move to next tab",
     "previous-tab":"Move to previous tab",
-    "cancel-download": "Cancel Download",
+    "cancel-download": "Cancel download",
     "cancel-download-text": "Are you sure you want to cancel the download of <b>{{ZIM}}</b>?",
     "delete-book": "Delete book",
     "delete-book-text": "Are you sure you want to delete <b>{{ZIM}}</b>?",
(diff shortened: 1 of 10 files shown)


mgautierfr:
Nice changes !

> Overall it looks like there are many more bugs related to Download management, and the proper way to eliminate them all is to significantly simplify the code architecture through a more direct usage of Kiwix::Download.

I strongly agree. I would go even further. I think we could probably drop the whole `ContentManager`.

At the time, The library view was implemented in html/js and it was discussing with the c++ part through a `QWebChannel`. The `ContentManager` was the C++ endpoint to pass value in the channel. But now #946 is merged, we don't need it anymore. `Node` can directly call the library to get the information. Even more, we can have the `Library` itself implement `QAbstractItemModel` and directly have the view displaying the `Library`. And if we go with the idea of the implementation of a `RemoteLibrary` class (https://github.com/kiwix/kiwix-desktop/issues/957#issuecomment-1717362443) we could also make it implement `QAbstractItemModel` and simply display the remote library too.

PR #1027, opened 2023-12-19
Use kiwix-build's github action to download dependencies.

mgautierfr:


kelson42:
@mgautierfr This commit seems to modify more than jus the workflow. Looks strange at a first sight.

mgautierfr:
> @mgautierfr This commit seems to modify more than jus the workflow. Looks strange at a first sight.

Indeed, it seems I have committed a part of my ongoing works. This is fixed now.

Issue #1022, opened 2023-12-11
Cancelling a download breaks the option to start it again


veloman-yunkan:
Steps to reproduce

1. Start a download
2. Pause it
3. Cancel it
4. Start it again - the download progress indicator appears at the value last recorded (at step 2) but the download is actually NOT started.

mgautierfr:
Fixed by kiwix/libkiwix#1066

PR #1035, opened 2024-02-06
Made it easier to display info and/or confirmation boxes

veloman-yunkan:
During refactoring-based-study of kiwix-desktop code in the context of the work on fixing download management (issues #87 and #1022) I made `KiwixConfirmBox` easier and more fool-proof to use and decided to extract those changes into a separate small PR.

Note that `showConfirmBox()` lacks a parameter for the "no" action because of the absence of the need for it in the current code. However, would such a need arise in the future, adding support for it should be trivial.

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -504,6 +504,26 @@ QString formatText(QString text)
     return finalText;
 }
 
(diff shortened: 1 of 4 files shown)


mgautierfr (approved):
LGTM. I like this kind of PR.

PR #1036, opened 2024-02-07
More reliable loading of remote thumbnails

veloman-yunkan:
In the wake of #1035 comes the next small PR.

It solves two problems:

1. Loading of thumbnails if a debug server (support for which was introduced in the first commit of #1024) is used
2. Changing the library view (applying filtering, switching from online to local) while the loading of thumbnails is in progress. I don't have a test case proving that it can lead to some undesirable behaviour. I tried the simplest scenario but it worked nicely. It's  possible that no such scenario exists since the potential bug may be compensated by extra code in other places. However these changes make the implementation more robust.

### DIFF
--- a/src/thumbnaildownloader.h
+++ b/src/thumbnaildownloader.h
@@ -21,7 +21,7 @@ public:
     ~ThumbnailDownloader();
 
     void addDownload(QString url, ThumbnailId index);
-    void startDownload();
+    void startNextDownload();
     void clearQueue() { m_downloadQueue.clear(); }
 
 private:
--- a/src/thumbnaildownloader.h
+++ b/src/thumbnaildownloader.h
@@ -17,7 +17,7 @@ public:
     typedef QPair<ThumbnailId, QString> ThumbnailInfo;
 
 public:
-    ThumbnailDownloader(QObject *parent = 0);
+    ThumbnailDownloader();
     ~ThumbnailDownloader();
 
     void addDownload(QString url, ThumbnailId index);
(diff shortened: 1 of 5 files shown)


src/thumbnaildownloader.h
```diff
@@ -12,25 +12,29 @@ class ThumbnailDownloader : public QObject
...
+
 public:
     ThumbnailDownloader(QObject *parent = 0);
     ~ThumbnailDownloader();
 
-    void addDownload(QString url, QModelIndex index);
+    void addDownload(QString url, ThumbnailId index);
     void startDownload();
-    void downloadOnePair(QPair<QModelIndex, QString> urlPair);
+    void downloadOnePair(ThumbnailInfo thumbnailInfo);
```
mgautierfr: Should we also rename this method also ?

src/thumbnaildownloader.h
```diff
@@ -22,14 +22,16 @@ class ThumbnailDownloader : public QObject
 
     void addDownload(QString url, ThumbnailId index);
     void startDownload();
-    void downloadOnePair(ThumbnailInfo thumbnailInfo);
-    void clearQueue() { m_urlPairList.clear(); }
+    void clearQueue() { m_downloadQueue.clear(); }
+
+private:
+    void downloadThumbnail(ThumbnailInfo thumbnailInfo);
```
mgautierfr, line 26: This is done here..

mgautierfr (approved):

PR #1044, opened 2024-02-25
Specify max dependence versions

### LINKED ISSUE #1020: Old Kiwix desktop versions in ppa repository, blocker for inclusion in Labdoo / Ubuntu (report from Mastodon user)
Report from a Mastodon user:

> Hi, Labdoo / Ubuntu was updated from kiwix.2.1 to kiwix.2.2. Kiwix.2.2 requires libkiwix11 as a dependency, but during the update the version libkiwix10 13.00-jammy was installed and the whole thing doesn't want to work together. Actually this should not happen because we install kiwix from a ppa. But you are not in it. Do you have the bug on your screen? Is a patch planned? Thank you.
>
> PS Labdoo uses 22.04 LTS. Kiwix.2.2 needs libkiwix11, but 22.04 LTS latest patch still comes with libkiwix10 13.00-jammy.

This user doesn't have a GitHub account, so couldn't make a report themselves. Is this fixable?

kelson42:
Fixes #1020

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -6,7 +6,8 @@ Build-Depends: debhelper-compat (= 13),
  pkg-config,
  qtbase5-dev,
  qtwebengine5-dev,
- libkiwix-dev (>= 11.0.0~)
+ libkiwix-dev (>= 13.0.0), libkiwix-dev (<< 14.0.0),
+ libzim-dev (>= 9.0.0), libzim-dev (<< 10.0.0),
 Standards-Version: 4.5.0
 Homepage: https://www.kiwix.org/
 Rules-Requires-Root: no
@@ -16,6 +17,8 @@ Architecture: any
 Depends: ${shlibs:Depends},
  ${misc:Depends}
 Description: offline Wikipedia reader
+ Kiwix is an offline reader and manager for online content
+ like Wikipedia, Project Gutenberg, or TED Talks.
  Kiwix allows you to read and search through offline content
  as they were online. Similar to a browser, Kiwix works with
  the highly compressed ZIM file format.
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -183,12 +183,12 @@ unix {
   INSTALLS += mime_file
 }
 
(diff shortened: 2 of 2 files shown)


debian/control
```diff
@@ -6,8 +6,8 @@ Build-Depends: debhelper-compat (= 13),
  pkg-config,
  qtbase5-dev,
  qtwebengine5-dev,
- libkiwix-dev (>= 11.0.0~)
-Standards-Version: 4.5.0
```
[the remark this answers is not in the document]
kelson42: yes, will reintroduce

veloman-yunkan (approved):

debian/control
```diff
@@ -6,7 +6,8 @@ Build-Depends: debhelper-compat (= 13),
  pkg-config,
  qtbase5-dev,
  qtwebengine5-dev,
- libkiwix-dev (>= 11.0.0~)
+ libkiwix-dev (>= 13.0.0), libkiwix-dev (<< 14.0.0)
+ libzim-dev (>= 9.0.0), libzim-dev (<< 10.0.0)
```
mgautierfr, line 10: Debian packaging seems to not understand `libkiwix-dev (<< 14.0.0)`. (See https://github.com/kiwix/kiwix-desktop/actions/runs/8053475026/job/21995971665)

Maybe this should be a simple `<` ?

mgautierfr (changes requested):

debian/control
veloman-yunkan: I checked the [documentation](https://www.debian.org/doc/debian-policy/ch-relationships.html#syntax-of-relationship-fields) during my review. `<<` seems to be the only valid syntax:

> The relations allowed are `<<`, `<=`, `=`, `>=` and `>>` for strictly earlier, earlier or equal, exactly equal, later or equal and strictly later, respectively.

veloman-yunkan: I think the problem is with the missing comma after `libkiwix-dev (<< 14.0.0)`

kelson42: Fixed!

PR #1041, opened 2024-02-22
Some cleanup of download management related code in ContentManager 

veloman-yunkan:
This cleanup PR was extracted from #1038. It doesn't fix any ticket, but makes #87 a closer target.

PR #1045, opened 2024-02-27
More cleanup of download management related code in ContentManager

veloman-yunkan:
This PR is the next batch of changes extracted from #1038 and a follow-up to #1041.


mgautierfr (approved):
LGTM

PR #1046, opened 2024-02-28
Initial Qt6 support

### LINKED ISSUE #365: Fullscreen is broken on Wayland.
If using the wayland compositor, the fullscreen functionality hide the window. Switching back to normal display the window correctly.

It is a wayland/QT bug (https://bugreports.qt.io/browse/QTBUG-58574), there is nothing we can do to fix the issue.

Maybe we could detect that we are on Wayland and deactivate the fullscreen functionality to at least avoid the user to "lose" its Window.

### LINKED ISSUE #912: Tab graphic bug
Unnecessary tab separator is displayed when opening a new tab:

And when you close tab it disappears:

I also think that opening a blank tab with no content is itself a bug. At least I did not find how to interact with it (search doesn't work on it).

Version: 2.3.1
Windows 10 Pro 22H2 (build 19045.2364)

### LINKED ISSUE #1033: Port to qt6
Needed for : https://github.com/kiwix/kiwix-build/issues/594 and #1029

adamlamar:
Fixes #1033
Fixes #365
Fixes #912 

This PR implements compile support for Qt6. There are still some bugs and limitations, but the app compiles and runs. The changes mostly fall into these categories:

* Disabling deprecated APIs in 6.0+ in the Qt5 build
* Adding a few Q_OBJECT references per the `clazy` tool checks
* Some minor changes to string handling that usually caused ambiguous method references, or because a constructor was removed
* Handling of new methods or introduction of enum types (e.g. enums converted from int usually don't implement `operator+`)

A few more things to fix before this is ready to merge:

- [x] The "more" menu (the `...` icon) needs to be built
- [x] The context menu for the webview no longer has access to `createStandardContextMenu()` - need to find the replacement
- [x] The `print` method no longer exists on a webview page - need to write up an approach to write a temporary PDF to disk, then print the PDF
- [x] Write up a README to document instructions for building with Qt6 on ubuntu 22.04
- [x] Test the new keybind sequences on a non-VM


kelson42:
@adamlamar Thank you very much for this draft.
@mgautierfr @mgautierfr Could younplease have already a quick look and give feedback about what has been done so far?

adamlamar:
Some more updates on the context menu:

- I couldn't find much on the Qt6 docs, but AFAICT the `createStandardContextMenu` method is simply not available anymore. I decided to implement a custom QMenu and make it more consistent with typical browser behaviors
- When you right-click on the background, you get the standard Back/Forward actions. The new menu uses the same icons as the toolbar, instead of the previously used system icons
- When you right-click on an internal link, you get `Open in new tab`. This is also what happens if you middle click on a link (and this is what browsers typically do)
- When you right-click on an external link, you get `Open in new browser window`. This is what happened before
- There are no longer menu actions to `Copy link` or `Save link`, since I believe these were not intended (see screenshot below)

Old:
Screenshot 2024-02-28 at 5 35 20 PM
Screenshot 2024-02-28 at 5 35 09 PM

Here's a video with the new approach:

[new-context-menu.webm](https://github.com/kiwix/kiwix-desktop/assets/13948507/0e0626ba-3822-49f2-bc59-10b145c8c0be)

adamlamar:
@mgautierfr I've completed the tasks above and AFAICT everything is working on both Qt5 and Qt6. Should be ready for a full review. Let me know if you have any questions.

kelson42:
To me it fixes #365 (with qt6.2 on Ubuntu), @mgautierfr could you confirm?

src/webview.cpp
```diff
@@ -206,26 +206,69 @@ void WebView::wheelEvent(QWheelEvent *event) {
...
+
+    auto forwardAction = new QAction(gt("forward"));
+    forwardAction->setEnabled(app->getAction(KiwixApp::HistoryForwardAction)->isEnabled());
+    forwardAction->setIcon(app->getAction(KiwixApp::HistoryForwardAction)->icon());
+    menu->addAction(forwardAction);
+    connect(menu, &QObject::destroyed, forwardAction, &QObject::deleteLater);
+    connect(forwardAction, &QAction::triggered, this, [=](bool checked) {
+        Q_UNUSED(checked);
+        KiwixApp::instance()->getTabWidget()->triggerWebPageAction(QWebEnginePage::Forward);
+    });
```
veloman-yunkan, line 241: Can't you use `app->getAction(KiwixApp::History[Back|Forward]Action)` directly instead of creating new temporary `QAction` objects? I think that it should be possible since the handler code doesn't depend on the context (local variables).

src/kiwixapp.cpp
```diff
@@ -346,25 +346,25 @@ void KiwixApp::setMonitorDir(const QString &dir) {
 
 void KiwixApp::createAction()
 {
-    CREATE_ACTION_ICON_SHORTCUT(KiwixServeAction, "share", gt("local-kiwix-server"), QKeySequence(Qt::CTRL, Qt::Key_I));
+    CREATE_ACTION_ICON_SHORTCUT(KiwixServeAction, "share", gt("local-kiwix-server"), QKeySequence(Qt::CTRL | Qt::Key_I));
```
veloman-yunkan: I think that this entire commit better be merged with the earlier one that replaced `+` with `,`.

src/fullscreenwindow.cpp
```diff
@@ -1,4 +1,4 @@
-#include <QMenu> // avoids "`QMenu` doesn't name a type" error in Qt6
+class QMenu;
```
veloman-yunkan, line 1: Please merge/squash this commit into the earlier one that introduced the changes that are now being fixed.

veloman-yunkan (changes requested):
I only reviewed the code (didn't try to build and test `kiwix-desktop` with qt6).

adamlamar:
@kelson42 I noticed those same warnings, but I believe they were preexisting. I actually have a branch where I fixed all the warnings and enabled `-Werror`, but its based on this branch so might be easier to wait until this is merged. I also noticed that library selection issue on first start - let me see if I can reproduce and figure out what's going on.

@veloman-yunkan Thanks for the comments. I will work on improving the code portability in those places where there is some duplication. As for the commits, I typically squash before merge anyway. I can do that sooner so that the extra cruft isn't around.

src/webview.cpp
adamlamar: I was mostly following the previous approach where the `QAction` objects were allocated with `new` and then deleted using `connect`. My understanding is that `menu->addAction` expects the caller to maintain the lifecycle of the object for the duration of the menu's existence. I would guess `HistoryForwardBack` would pretty much always outlive the menu, but I'm not sure. Maybe there are edge cases during shutdown.

Reusing back and forward directly would create an implicit dependency that could cause issues if this ever changed, which might create bugs. And, `HistoryBackAction` is slightly different than `backAction` because it includes a history of items when you long-press (perhaps other differences).

For the savings of a malloc/free and a few bytes of memory, sharing doesn't seem worth the cost IMO.

adamlamar:
@kelson42 After looking into the Language filter issue I discovered that it was not constructing the QVariant correctly, even in Qt5. If I understand the intent, the default on first startup is to show you a filtered set of zim files that match your current system language+locale. For example mine is `English|en`. When printed to the console, the correct language list looks like this:

```
langList (QVariant(QString, "English|en"))
```

Two things:
* In Qt5, the language list was always empty, despite the default being set, which made the filter always empty (maybe this feature was not working as intended)
* In Qt6,  every character became is its own item in an array of `QChar`:

```
langList QList(QVariant(QChar, 'E'), QVariant(QChar, 'n'), QVariant(QChar, 'g'), QVariant(QChar, 'l'), QVariant(QChar, 'i'), QVariant(QChar, 's'), QVariant(QChar, 'h'), QVariant(QChar, '|'), QVariant(QChar, 'e'), QVariant(QChar, 'n'))
```

After my fixes, the language filter is shown on first start in both:
Screenshot 2024-03-01 at 3 29 40 PM

@veloman-yunkan This should be ready for another review.

src/webview.cpp
veloman-yunkan: > For the savings of a malloc/free and a few bytes of memory, sharing doesn't seem worth the cost IMO.

I was more concerned about the amount of code (which should always be maintained at a minimum as long as readability is not hurt).

> Reusing back and forward directly would create an implicit dependency that could cause issues if this ever changed, which might create bugs.

There is a dependency in any case and context menus are ephemeral to an extent that I can't imagine a situation causing issues with the implicit dependency without also affecting the explicit one. 

> And, `HistoryBackAction` is slightly different than `backAction` because it includes a history of items when you long-press (perhaps other differences).

That's the strongest point in your argument and it would be enough.

veloman-yunkan (changes requested):
Having all your changes squashed into a single commit hurts more than it helps. Seeing issues described and resolved one (or a few) at a time is useful.

Please address the new comments as fixup commits so that I don't have to look through the full list of changes during the next (hopefully, last) iteration.

src/kiwixapp.cpp
```diff
@@ -209,12 +215,23 @@ void KiwixApp::printPage()
...
         });
+#else
+        webview->print(printer);
+        connect(webview, &QWebEngineView::printFinished, this, [=](bool success) {
+            if (!success) {
+                showMessage(gt("print-page-error"), gt("error-title"), QMessageBox::Critical);
+            }
+            delete printer;
+        });
+#endif
```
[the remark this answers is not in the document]
adamlamar: Sure, fixed

src/settingsmanager.cpp
```diff
@@ -147,8 +148,18 @@ void SettingsManager::initSettings()
...
+
+    QList<QString> defaultLangList; // Qt5 QList doesn't support supplying a constructor list, so use append() for Qt5+Qt6 compat
+    defaultLangList.append(defaultLang);
+
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+    QVariant defaultLangVariant(defaultLangList); // Qt5 requires explicit conversion from QList to QVariant
+    m_langList = m_settings.value("language", defaultLangVariant).toList();
+#else
+    m_langList = m_settings.value("language", defaultLangList).toList();
+#endif
```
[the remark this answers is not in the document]
adamlamar: Yes, it can be done without conditional compilation. Since there is otherwise no context about why the objects are initialized in a specific way to maintain Qt5&Qt6 compatibility, I added a comment which hopefully is clear.

subprojects/QtSingleApplication/src/qtlockedfile.h
```diff
@@ -66,6 +66,9 @@ namespace QtLP_Private {
 
 class QT_QTLOCKEDFILE_EXPORT QtLockedFile : public QFile
 {
+    //Q_OBJECT
+    // TODO: Uncomment Q_OBJECT. Setting Q_OBJECT here causes this error:
+    // undefined reference to `vtable for QtLP_Private::QtLockedFile'
```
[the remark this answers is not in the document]
adamlamar: Yes, this works in Qt6 without Q_OBJECT as far as I know. I removed the comment since it doesn't seem to have any effect.

veloman-yunkan (approved):
LGTM

kelson42:
@adamlamar Thank you very much for this PR. This is a very nice step forward. We might a few adjustments to do post PR, and might then contact you then.

Issue #893, opened 2022-10-07
Kiwix-desktop is not resilient to manual change in the library.xml


mgautierfr:
The root cause is that `kiwix::Library::getArchiveById` (https://github.com/kiwix/libkiwix/blob/07ff4eab4397bc4fb208191374c1e7a6e9b1def4/src/library.cpp#L238-L252) return a nullptr if the requested book is existing in the library.xml but the archive path is invalid. (It throws a `std::out_of_range` if the bookId is not present at all in the library.xml).

But kiwix-desktop catch the `std::out_of_range` but never check if the returned archive (without exception) is not null.
This is done in several places in kiwix-desktop:
- https://github.com/kiwix/kiwix-desktop/blob/f1aafc1d9544677cdde5c3687552f2ac81ee5b3c/src/readinglistbar.cpp#L37-L40
- https://github.com/kiwix/kiwix-desktop/blob/c3eace12afdc23dfd0d839275d6e0c2d2f7de2a1/src/suggestionlistworker.cpp#L23
- https://github.com/kiwix/kiwix-desktop/blob/f1aafc1d9544677cdde5c3687552f2ac81ee5b3c/src/urlschemehandler.cpp#L44-L49
- https://github.com/kiwix/kiwix-desktop/blob/f1aafc1d9544677cdde5c3687552f2ac81ee5b3c/src/webview.cpp#L172-L176
- https://github.com/kiwix/kiwix-desktop/blob/e8478670755a1940bc8782c58bf550f69beb219e/src/kiwixapp.cpp#L241

Having a book in the library but not a valid path can be simply done by removing by hand a zim file referenced by the library.xml

In this case, the behavior is the following (at least):
- The book is not displayed in the local file. But the "All files" section do not propose to download the book but to open it. Of course, trying to open it leads to a crash as we try to use a nullptr.
- If a article in the book was bookmarked. kiwix-desktop crash at start as we try to populate the readinglist with the registered bookmark.

It **seems** that the simplest thing would be to make `kiwix::Library::getArchiveById` throw a `std::out_of_range` if the path is invalid. But I'm not sure this is the best thing to do:
- One may want to differentiate a book not in the library from a invalid path in the library, so we may throw different kind of exception
- I'm not sure that throwing a `std::out_of_range` would fix the strange behavior of not displaying the book in the local files but not being able to download the book again.
[...]

PR #1048, opened 2024-03-07
Fix #186: Introduced Tab Restoration on Restart

### LINKED ISSUE #186: Remember open tabs over restart
A small suggestion for a feature: how about opening the last opened zim file automatically? I think this would help many people that are using only one zim file.

ShaopengLin:
Changes:
- ~Pop-up to prompt users whether they want to restore the tabs. If not, start a new session.~ Added a checkbox in settings for users to toggle whether they want tab restoration or not.
- Tab order is preserved.
- Tab closures are recorded.
- Tab navigation URL is preserved.
- Blank/Settings tabs are recorded but not opened again on startup.

Fix-up Changes:
- Does not open zim files that are deleted from the File System.
- Pop-ups replaced with toggle option in settings.
- Writes to a separate file than the program settings registry.
- Simplifies session commit strategy.

Fix #186

### DIFF
--- a/resources/i18n/qqq.json
+++ b/resources/i18n/qqq.json
@@ -45,5 +45,6 @@
 	"monitor-dir-dialog-msg": "\"Monitor\" means \"watch\" in this context. The monitor directory is monitored/watched for new ZIM files.",
 	"monitor-clear-dir-dialog-title": "\"Monitor\" means \"watch\" in this context. The monitor directory is monitored/watched for new ZIM files.",
 	"monitor-clear-dir-dialog-msg": "\"Monitor\" means \"watch\" in this context. The monitor directory is monitored/watched for new ZIM files.",
-	"open-book": "\"Open\" is a imperative, not an adjective."
+	"open-book": "\"Open\" is a imperative, not an adjective.",
+	"open-previous-tabs-at-startup": "The tabs that were open when the user closed the application is opened again when the application is restarted."
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -189,15 +189,18 @@ void KiwixApp::restoreTabs()
 
     /* Restart a new session to prevent duplicate records in openURL */
     saveListOfOpenTabs();
-    for (const auto &zimUrl : tabsToOpen)
+    if (m_settingsManager.getReopenTab())
     {
-      try
+      for (const auto &zimUrl : tabsToOpen)
       {
+        try
+        {
           /* Throws exception if zim file cannot be found */
           m_library.getArchive(QUrl(zimUrl).host().split('.')[0]);
           openUrl(QUrl(zimUrl));
+        }
+        catch (std::exception &e) { /* Blank */ }
       }
-      catch (std::exception &e) { /* Blank */ }
     }
 }
 
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -32,7 +32,8 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
       mp_manager(nullptr),
       mp_mainWindow(nullptr),
       mp_nameMapper(std::make_shared<kiwix::UpdatableNameMapper>(m_library.getKiwixLibrary(), false)),
-      m_server(m_library.getKiwixLibrary(), mp_nameMapper)
+      m_server(m_library.getKiwixLibrary(), mp_nameMapper),
+      mp_session(nullptr)
 {
     try {
         m_translation.setTranslation(QLocale());
@@ -113,6 +114,8 @@ void KiwixApp::init()
             m_library.asyncUpdateFromDir(dir);
         }
     }
+
+    restoreTabs();
 }
 
 KiwixApp::~KiwixApp()
@@ -174,6 +177,28 @@ QString KiwixApp::findLibraryDirectory()
   }
 
   return currentDataDir;
+}
+
+void KiwixApp::restoreTabs()
+{
+    /* Place session file in our global library path */
+    QDir dir(m_libraryDirectory);
+    mp_session = new QSettings(dir.filePath("kiwix-desktop.session"),
+                               QSettings::defaultFormat(), this);
+    QStringList tabsToOpen = mp_session->value("reopenTabList").toStringList();
+
+    /* Restart a new session to prevent duplicate records in openURL */
+    saveListOfOpenTabs();
+    for (const auto &zimUrl : tabsToOpen)
+    {
+      try
+      {
+          /* Throws exception if zim file cannot be found */
+          m_library.getArchive(QUrl(zimUrl).host().split('.')[0]);
+          openUrl(QUrl(zimUrl));
+      }
+      catch (std::exception &e) { /* Blank */ }
+    }
 }
 
 KiwixApp *KiwixApp::instance()
@@ -503,4 +528,9 @@ QString KiwixApp::parseStyleFromFile(QString filePath)
     QString styleSheet = QString(file.readAll());
     file.close();
     return styleSheet;
+}
+
+void KiwixApp::saveListOfOpenTabs()
+{
+  return mp_session->setValue("reopenTabList", getTabWidget()->getTabUrls());
 }
--- a/src/kiwixapp.h
+++ b/src/kiwixapp.h
@@ -87,6 +87,7 @@ public:
     void setMonitorDir(const QString &dir);
     bool isCurrentArticleBookmarked();
     QString parseStyleFromFile(QString filePath);
+    void saveListOfOpenTabs();
 
 public slots:
     void newTab();
@@ -116,10 +117,12 @@ private:
     kiwix::Server m_server;
     Translation m_translation;
     QFileSystemWatcher m_watcher;
+    QSettings* mp_session;
 
     QAction*     mpa_actions[MAX_ACTION];
 
     QString findLibraryDirectory();
+    void restoreTabs();
 };
 
 QString gt(const QString &key);
--- a/src/settingsmanager.h
+++ b/src/settingsmanager.h
@@ -29,6 +29,7 @@ public:
     QString getDownloadDir() const { return m_downloadDir; }
     QString getMonitorDir() const { return m_monitorDir; }
     bool getMoveToTrash() const { return m_moveToTrash; }
+    bool getReopenTab() const { return m_reopenTab; }
     FilterList getLanguageList() { return deducePair(m_langList); }
     FilterList getCategoryList() { return deducePair(m_categoryList); }
     FilterList getContentType() { return deducePair(m_contentTypeList); }
@@ -40,9 +41,11 @@ public slots:
     void setDownloadDir(QString downloadDir);
     void setMonitorDir(QString monitorDir);
     void setMoveToTrash(bool moveToTrash);
+    void setReopenTab(bool reopenTab);
     void setLanguage(FilterList langList);
     void setCategory(FilterList categoryList);
     void setContentType(FilterList contentTypeList);
+
 private:
     void initSettings();
     QList<QVariant> flattenPair(FilterList pairList);
@@ -54,6 +57,7 @@ signals:
     void downloadDirChanged(QString downloadDir);
     void monitorDirChanged(QString monitorDir);
     void moveToTrashChanged(bool moveToTrash);
+    void reopenTabChanged(bool reopenTab);
     void languageChanged(QList<QVariant> langList);
     void categoryChanged(QList<QVariant> categoryList);
     void contentTypeChanged(QList<QVariant> contentTypeList);
@@ -67,6 +71,7 @@ private:
     QString m_downloadDir;
     QString m_monitorDir;
     bool m_moveToTrash;
+    bool m_reopenTab;
     QList<QVariant> m_langList;
     QList<QVariant> m_categoryList;
     QList<QVariant> m_contentTypeList;
--- a/src/settingsview.cpp
+++ b/src/settingsview.cpp
@@ -12,6 +12,7 @@ SettingsView::SettingsView(QWidget *parent)
     ui->widget->setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/_settingsManager.css"));
     connect(ui->zoomPercentSpinBox, QOverload<int>::of(&QSpinBox::valueChanged), this, &SettingsView::setZoom);
     connect(ui->moveToTrashToggle, &QCheckBox::clicked, this, &SettingsView::setMoveToTrash);
+    connect(ui->reopenTabToggle, &QCheckBox::clicked, this, &SettingsView::setReopenTab);
     connect(ui->browseButton, &QPushButton::clicked, this, &SettingsView::browseDownloadDir);
     connect(ui->resetButton, &QPushButton::clicked, this, &SettingsView::resetDownloadDir);
     connect(ui->monitorBrowse, &QPushButton::clicked, this, &SettingsView::browseMonitorDir);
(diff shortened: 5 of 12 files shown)


ShaopengLin:
Like an option for either restoring automatucally or have a pop-up for it? For automatic restoration it would be good to have a short-small notification about it as well.

kelson42:
Yes, a basic boolean setting "Restore tabs at start".

ShaopengLin:
Just found we have a settingsmanager.cpp. Will move code over there to refactor a commit.

kelson42:
> Yes, a basic boolean setting "Restore tabs at start". I will lokk how FF and Chrome exactly work...

I confirm, we should have a "Open previous windows and tabs" checkbox in the settings. Per default unchecked.

ShaopengLin:
> > Yes, a basic boolean setting "Restore tabs at start". I will lokk how FF and Chrome exactly work...
> 
> I confirm, we should have a "Open previous tabs at startup" checkbox in the settings. Per default unchecked.

I think currently I set it as default checked, simple change. Will wait for reviewer comment and change in a batch.

kelson42:
@ShaopengLin Option not visible in the settings, 
image

ShaopengLin:
Weird. I will test it using Qt6 as well. I was still developing on the Qt5 build. Might be some differences in building UI?

kelson42:
@ShaopengLin Sorry, me doing a wrong manipulation. Seems to work fine... but setting wording is NOK. SHould be "Open previous tabs at startup"

ShaopengLin:
Is there a way for me to update i18n for the text?

kelson42:
@ShaopengLin You should put only the string in English

ShaopengLin:
I just put it like this? 
`ui->reopenTabLabel->setText(gt("open-previous-tabs-on-startup"));`

src/settingsmanager.h
```diff
@@ -67,9 +74,11 @@ public slots:
     QString m_downloadDir;
     QString m_monitorDir;
     bool m_moveToTrash;
+    bool m_reopenTab;
     QList<QVariant> m_langList;
     QList<QVariant> m_categoryList;
     QList<QVariant> m_contentTypeList;
+    QStringList m_tabList;
```
veloman-yunkan, line 81: I don't think that the list of open tabs belongs to the app settings. It should rather be considered part of the app session.

Under Linux the settings are stored in `~/.config/Kiwix/Kiwix-desktop.conf`.  Other information is written under `~/.local/share/kiwix/`. There is a `kiwix.session` file in that latter directory that seems to always be empty. I think that the list of tabs should be written into that file.

@kelson42 Does it make sense?

src/tabbar.cpp
```diff
@@ -276,6 +278,47 @@ void TabBar::closeTabsByZimId(const QString &id)
...
+}
+
+void TabBar::swapTabInSave(int from, int to)
+{
+    QStringList tabs = KiwixApp::instance()->getSettingsManager()->getTabList();
+    QString tabUrl = tabs[from];
+    tabs.removeAt(from);
+    tabs.insert(to, tabUrl);
+    KiwixApp::instance()->getSettingsManager()->setTabList(tabs);
+}
```
veloman-yunkan, line 320: Why such a complex scheme of maintaining the list of open tabs? Can't you just generate/save the list of open tabs (by iterating over them) during the termination of the application?

src/settingsmanager.h
ShaopengLin: Agrees with moving them from settings to application. I believe the QSessionManager is the one that writes the session file? On the documentation it still mentions to use QSettings to store those session information. I can probably move QSettings code to work with QSessionManager.

ShaopengLin: So I found kiwix.session is located in the library data directory. I can just use QSettings to write to this file if we confirm this file isn't used is any other placed.

ShaopengLin:
I will update this PR after #1057 and on resolution of the comments: [comment on text translation](https://github.com/kiwix/kiwix-desktop/pull/1048#issuecomment-1987321480) and [comment on session](https://github.com/kiwix/kiwix-desktop/pull/1048#discussion_r1520276617)

src/settingsmanager.h
kelson42: LGTM

src/kiwixapp.cpp
```diff
@@ -111,6 +111,16 @@ void KiwixApp::init()
...
+    
+    QStringList tabs = m_settingsManager.getTabList();
+    m_settingsManager.setTabList(QStringList{});
+    if (m_settingsManager.getReopenTab())
+    {
+      /* Restart a new session to prevent duplicate records in openURL */
+      for (const auto &zimFile : tabs)
+        if (!zimFile.isEmpty())
+          openUrl(zimFile);
+    }
```
mgautierfr, line 123: Maybe we want to check the saved tabs (url) are still valid before open them.

If user delete a zim file after quitting the application, we will try to open a zim file which is not present anymore.
`openUrl` will open a new tab and it will be a 404.
It may be what we want or we may not open a tab at all in this case...

src/settingsmanager.h
ShaopengLin: Ok, we can't do that. kiwix.session is used by aria2 to restore download sessions when the application closes. I can put the content in another file in the data directory.

src/kiwixapp.cpp
ShaopengLin: On my side, the application simply segfaults if the user deletes the zim file. Tested on a fresh main as well. This might be a new issue?

ShaopengLin: Actually, I will fix them right here, as using openBook to handle an error like this is going to be the way to go anyways.

mgautierfr: The crash is probably because of https://github.com/kiwix/kiwix-desktop/issues/893

ShaopengLin: Got it. I will remove that nullptr check then.

src/tabbar.cpp
ShaopengLin: Done

src/settingsmanager.h
ShaopengLin: Done

veloman-yunkan (changes requested):
I am a little concerned with the new calls to `Kiwix::instance()`, but I think that I'd rather address it separately in a dedicated design improvement PR.

Apart from fixing the issues outlined in the review comments, please split this PR in two commits:

1. Add unconditional reopening of tabs from previous session;
2. Introduce a setting that controls that behaviour.

src/kiwixapp.cpp
```diff
@@ -113,6 +114,28 @@ void KiwixApp::init()
             m_library.asyncUpdateFromDir(dir);
         }
     }
+
+    /* Place session file in our global library path */
+    QDir dir(m_libraryDirectory);
+    this->mp_session = new QSettings(std::move(dir).filePath("user.session"),
```
veloman-yunkan, line 120: Why is `std::move()` needed?

ShaopengLin: Oops forgot to remove the std::move. No point moving since I am not performing any copying.

ShaopengLin:
@veloman-yunkan Could you tag me when you start the refactoring work? I am a GSOC student so seeing practical design patterns being applied is great learning for me as well. I am guessing something like dependency injection or complete functions.

src/webview.cpp
```diff
@@ -166,12 +166,13 @@ QWebEngineView* WebView::createWindow(QWebEnginePage::WebWindowType type)
 
 void WebView::onUrlChanged(const QUrl& url) {
     auto zimId = getZimIdFromUrl(url);
+    auto app = KiwixApp::instance();
+    app->saveListOfOpenTabs();
```
ShaopengLin, line 170: I replaced the previous call on `webpage.cpp` as I found that was a bad choice. No need for the `acceptNavigationRequest`. onUrlChanged is indeed called on in-page URL navigations.

src/kiwixapp.cpp

src/kiwixapp.h
```diff
@@ -84,6 +84,7 @@ class KiwixApp : public QtSingleApplication
     kiwix::Server* getLocalServer() { return &m_server; }
     SettingsManager* getSettingsManager() { return &m_settingsManager; };
     QString getText(const QString &key) { return m_translation.getText(key); };
+    QSettings* getSession() { return mp_session; }
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/tabbar.cpp
```diff
@@ -163,6 +163,7 @@ ZimView* TabBar::createNewTab(bool setCurrent, bool nextToCurrentTab)
     connect(tab, &ZimView::webActionEnabledChanged,
             this, &TabBar::onWebviewHistoryActionChanged);
 
+    this->saveListOfOpenTabsToSession();
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/tabbar.cpp
```diff
@@ -268,6 +269,30 @@ void TabBar::closeTabsByZimId(const QString &id)
...
+      ->getSession()
+      ->value("reopenTabList", QStringList{})
+      .value<QStringList>();
+}
+
+void TabBar::saveListOfOpenTabsToSession()
+{
+  KiwixApp::instance()->getSession()->setValue("reopenTabList",
+                                               this->getTabUrls());
+}
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/tabbar.cpp
```diff
@@ -268,6 +269,30 @@ void TabBar::closeTabsByZimId(const QString &id)
     }
 }
 
+QStringList TabBar::getTabUrls() {
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/tabbar.cpp
```diff
@@ -268,6 +269,30 @@ void TabBar::closeTabsByZimId(const QString &id)
...
+    return idList;
+}
+
+QStringList TabBar::getListOfOpenTabsFromSession()
+{
+  return KiwixApp::instance()
+      ->getSession()
+      ->value("reopenTabList", QStringList{})
+      .value<QStringList>();
+}
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/kiwixapp.cpp
```diff
@@ -113,6 +114,28 @@ void KiwixApp::init()
             m_library.asyncUpdateFromDir(dir);
         }
     }
+
+    /* Place session file in our global library path */
+    QDir dir(m_libraryDirectory);
+    this->mp_session = new QSettings(dir.filePath("user.session"),
+                                     QSettings::defaultFormat(), this);
+    QStringList tabsToOpen =
+        this->mp_session->value("reopenTabList", QStringList{})
```
veloman-yunkan, line 123: What happens if you don't provide `QStringList{}` as the default value? Won't a null `QVariant` object be converted to an empty `QStringList`?

veloman-yunkan (changes requested):
In case the number of new comments makes you think that I am not happy with the progress you will be wrong. The fact is that we are much closer to having this PR merged and I am just paying attention to more mundane details. :wink:

src/kiwixapp.cpp
```diff
@@ -113,6 +114,28 @@ void KiwixApp::init()
...
+    for (const auto &zimUrl : tabsToOpen)
+    {
+      try
+      {
+          /* Throws exception if zim file cannot be found */
+          this->m_library.getArchive(QUrl(zimUrl).host().split('.')[0]);
+          this->openUrl(QUrl(zimUrl));
+      }
+      catch (std::exception &e) { /* Blank */ }
+    }
```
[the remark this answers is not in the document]
ShaopengLin: I agree as the rest of the code in KiwixApp is following the 'no this' approach. I'm going to stay consistent.

src/kiwixapp.cpp
ShaopengLin: True, it is converted to emtpy QStringList. I will revert to default call.

src/kiwixapp.cpp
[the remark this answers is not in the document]

src/kiwixapp.cpp
```diff
@@ -501,3 +524,9 @@ QString KiwixApp::parseStyleFromFile(QString filePath)
     file.close();
     return styleSheet;
 }
+
+void KiwixApp::saveListOfOpenTabs()
+{
+  return this->mp_session->setValue("reopenTabList",
+                                    this->getTabWidget()->getTabUrls());
```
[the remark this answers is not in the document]

src/kiwixapp.cpp
```diff
@@ -113,6 +114,28 @@ void KiwixApp::init()
             m_library.asyncUpdateFromDir(dir);
         }
     }
+
+    /* Place session file in our global library path */
+    QDir dir(m_libraryDirectory);
+    this->mp_session = new QSettings(dir.filePath("user.session"),
```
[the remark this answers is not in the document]
ShaopengLin: Made the modification just in case.

src/kiwixapp.cpp

src/tabbar.h
```diff
@@ -49,6 +49,7 @@ class TabBar : public QTabBar
     virtual QSize tabSizeHint(int index) const;
     void openFindInPageBar();
     void closeTabsByZimId(const QString &id);
+    const QStringList getTabUrls();
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/kiwixapp.cpp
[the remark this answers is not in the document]

src/settingsview.cpp
```diff
@@ -31,14 +33,18 @@ SettingsView::SettingsView(QWidget *parent)
     ui->monitorHelp->setText("<b>?</b>");
     ui->monitorHelp->setToolTip(gt("monitor-directory-tooltip"));
     ui->moveToTrashLabel->setText(gt("move-files-to-trash"));
+    ui->reopenTabLabel->setText(gt("open-previous-tabs-on-startup"));
```
veloman-yunkan, line 36: I don't remember if entries for `open-previous-tabs-on-startup` have been added in `resources/i18n/en.json` and `resources/i18n/qqq.json` in previous versions of this PR, but currently they are not there.

ShaopengLin: They are not. I thought originally these translations are added with translation wiki PRs after the merge.

Are these two files the only ones for me to change at the moment?

veloman-yunkan: Yes. Put the text in English (I thinks @kelson42's choice is "Open previous tabs at startup") in `en.json` and the explanation of that message for translators (since sometimes translations cannot be done out of context) in `qqq.json`.

ShaopengLin: Done

resources/i18n/qqq.json
```diff
@@ -45,5 +45,6 @@
 	"monitor-dir-dialog-msg": "\"Monitor\" means \"watch\" in this context. The monitor directory is monitored/watched for new ZIM files.",
 	"monitor-clear-dir-dialog-title": "\"Monitor\" means \"watch\" in this context. The monitor directory is monitored/watched for new ZIM files.",
 	"monitor-clear-dir-dialog-msg": "\"Monitor\" means \"watch\" in this context. The monitor directory is monitored/watched for new ZIM files.",
-	"open-book": "\"Open\" is a imperative, not an adjective."
+	"open-book": "\"Open\" is a imperative, not an adjective.",
+	"open-previous-tabs-on-startup": "The tabs that were open when the user closed the application is opened again when the application is restarted."
```
veloman-yunkan, line 49: The message key in `qqq.json` doesn't match the key in `en.json` (and the code).

ShaopengLin: Oops, done.

veloman-yunkan (approved):
LGTM (but note my so far unanswered [question](https://github.com/kiwix/kiwix-desktop/pull/1048#pullrequestreview-1963112899) in the previous review).

mgautierfr:
We may want to save (and restore) also the which tab is the current one.
But it may be done in another PR (open a issue in this case). Else nothing to add.

ShaopengLin:
> We may want to save (and restore) also the which tab is the current one. But it may be done in another PR (open a issue in this case). Else nothing to add.

@mgautierfr It's rather a simple change that makes this feature complete. I will just add it here. Sorry for the delayed response.

kelson42:
@ShaopengLin Thank you very much for this PR. Please don't forget to open the ticket like agreed. Merging...

PR #1051, opened 2024-03-08
Fix warnings and enable Werror

adamlamar:
Fixed all the warnings in the build and added `-Werror`. Not sure if `Werror` is wanted since it might be annoying during development - please let me know either way.

Regarding the loading/installation of the translators, there seems to be a bug. `m_qtTranslator.load()` for the Qt translations succeeds, but then installing always fails. Both loading and installing `m_appTranslator` always fail. I didn't want to break any application functionality, so I added variables to check the return values to satisfy the warnings, but the behavior should be roughly the same.

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -543,6 +543,9 @@ void ContentManager::downloadBook(const QString &id)
 
 void ContentManager::eraseBookFilesFromComputer(const QString dirPath, const QString fileName, const bool moveToTrash)
 {
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
+    Q_UNUSED(moveToTrash);
+#endif
     if (fileName == "*") {
         return;
     }
@@ -638,6 +641,8 @@ void ContentManager::resumeBook(const QString& id)
 
 void ContentManager::cancelBook(const QString& id, QModelIndex index)
 {
+    Q_UNUSED(index);
+
     auto text = gt("cancel-download-text");
     text = text.replace("{{ZIM}}", QString::fromStdString(mp_library->getBookById(id).getTitle()));
     showConfirmBox(gt("cancel-download"), text, mp_view, [=]() {
@@ -738,7 +743,7 @@ QString makeHttpUrl(QString host, int port)
 } // unnamed namespace
 
 void ContentManager::updateRemoteLibrary(const QString& content) {
-    QtConcurrent::run([=]() {
+    (void) QtConcurrent::run([=]() {
         QMutexLocker locker(&remoteLibraryLocker);
         mp_remoteLibrary = kiwix::Library::create();
         kiwix::Manager manager(mp_remoteLibrary);
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -41,17 +41,14 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
         QMessageBox::critical(nullptr, "Translation error", e.what());
         return;
     }
-#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
-    m_qtTranslator.load(QLocale(), "qt", "_",
-                        QLibraryInfo::location(QLibraryInfo::TranslationsPath));
-#else
-    m_qtTranslator.load(QLocale(), "qt", "_",
-                        QLibraryInfo::path(QLibraryInfo::TranslationsPath));
-#endif
-    installTranslator(&m_qtTranslator);
 
-    m_appTranslator.load(QLocale(), "kiwix-desktop", "_", ":/i18n/");
-    installTranslator(&m_appTranslator);
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+    QString path = QLibraryInfo::location(QLibraryInfo::TranslationsPath);
+#else
+    QString path = QLibraryInfo::path(QLibraryInfo::TranslationsPath);
+#endif
+    loadAndInstallTranslations(m_qtTranslator, "qt", path);
+    loadAndInstallTranslations(m_appTranslator, "kiwix-desktop", ":/i18n/");
 
     QFontDatabase::addApplicationFont(":/fonts/Selawik/selawkb.ttf");
     QFontDatabase::addApplicationFont(":/fonts/Selawik/selawkl.ttf");
@@ -59,6 +56,12 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
     QFontDatabase::addApplicationFont(":/fonts/Selawik/selawksl.ttf");
     QFontDatabase::addApplicationFont(":/fonts/Selawik/selawk.ttf");
     setFont(QFont("Selawik"));
+}
+
+void KiwixApp::loadAndInstallTranslations(QTranslator& translator, const QString& filename, const QString& directory) {
+    if (translator.load(QLocale(), filename, "_", directory)) {
+        installTranslator(&translator);
+    }
 }
 
 void KiwixApp::init()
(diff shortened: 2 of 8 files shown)


adamlamar:
Oops, looks like lambda's don't satisfy the `QRunnable` interface on Qt 5.12. I will mark as draft and come back to this.

adamlamar:
Build fixed, ready for review.

src/kiwixapp.cpp
```diff
@@ -40,17 +40,30 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
...
 
-    m_appTranslator.load(QLocale(), "kiwix-desktop", "_", ":/i18n/");
-    installTranslator(&m_appTranslator);
+    loaded = m_appTranslator.load(QLocale(), "kiwix-desktop", "_", ":/i18n/");
+    if (loaded) {
+        installed = installTranslator(&m_appTranslator);
+        if (!installed) {
+            // ignore for now
+        }
+    }
```
veloman-yunkan, line 66: I would deduplicate this code (and make it more readable) by introducing a small helper function `loadAndInstallTranslations(translator, filename, directory)`.

src/contentmanager.cpp
```diff
@@ -780,7 +784,9 @@ void ContentManager::updateRemoteLibrary(const QString& content) {
         manager.readOpds(content.toStdString(), catalogUrl.toStdString());
         emit(this->booksChanged());
         emit(this->pendingRequest(false));
-    });
+    };
+    // wrap the lambda in LambdaRunnable for Qt 5.12 and later compat
+    QThreadPool::globalInstance()->start(new portutils::LambdaRunnable(f));
```
veloman-yunkan, line 789: Memory leak (since `LambdaRunnable` doesn't enable [auto-deletion](https://doc.qt.io/qt-5/qrunnable.html#setAutoDelete))

src/library.cpp
```diff
@@ -170,9 +171,11 @@ void Library::updateFromDir(QString monitorDir)
 
 void Library::asyncUpdateFromDir(QString dir)
 {
-    QThreadPool::globalInstance()->start([this, dir]() {
+    std::function f = [this, dir]() {
         updateFromDir(dir);
-    });
+    };
+    // wrap the lambda in LambdaRunnable for Qt 5.12 and later compat
+    QThreadPool::globalInstance()->start(new portutils::LambdaRunnable(f));
```
veloman-yunkan, line 178: Memory leak (since `LambdaRunnable` doesn't enable [auto-deletion](https://doc.qt.io/qt-5/qrunnable.html#setAutoDelete))

src/library.cpp
```diff
@@ -170,9 +171,11 @@ void Library::updateFromDir(QString monitorDir)
 
 void Library::asyncUpdateFromDir(QString dir)
 {
-    QtConcurrent::run( [=]() {
+    std::function f = [this, dir]() {
         updateFromDir(dir);
-    });
+    };
+    // wrap the lambda in LambdaRunnable for Qt 5.12 and later compat
+    QThreadPool::globalInstance()->start(new portutils::LambdaRunnable(f));
```
veloman-yunkan, line 178: I think that it would be better to provide a higher level utility `portutils::runConcurrently()` that would directly replace `QtConcurrent::run()`.

src/contentmanager.cpp
veloman-yunkan: And use `portutils::runConcurrently()` (as suggested in the other comment) here too

veloman-yunkan (changes requested):

src/contentmanager.cpp
adamlamar: There shouldn't be a memory leak because autoDelete is [enabled by default](https://doc.qt.io/qt-5/qrunnable.html#QRunnable).

src/kiwixapp.cpp
adamlamar: Sure, I added a helper method that is slightly cleaner, but a few lines longer overall.

src/library.cpp
adamlamar: `QtConcurrent::run()` returns a future and getting the return result blocks until the concurrent function has completed. Since this PR enforces checking return values, we can't use `QtConcurrent::run()` without blocking, or having some mechanism to come back and check later.

`QThreadPool::start()` is the [recommended approach](https://doc.qt.io/qt-6/qtconcurrentrun.html#returning-values-from-the-function) for fire-and-forget style requests.

src/kiwixapp.cpp
```diff
@@ -41,28 +41,20 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
...
-    if (loaded) {
-        installed = installTranslator(&m_qtTranslator);
-        if (!installed) {
-            // ignore for now
-        }
+    bool success = false;
+    success = loadAndInstallTranslations(m_qtTranslator, "qt", path);
+    if (!success) {
+        // ignore for now
     }
```
veloman-yunkan, line 53: Why don't you move this into `loadAndInstallTranslations()` and change the latter's return type to `void`?

src/library.cpp
veloman-yunkan: I think you misunderstood me. I meant that you could package your replacement code for `QtConcurrent::run()` (based on `QThreadPool::start()`) into a helper function `portutils::runConcurrently()` and then would only have replace the calls to  `QtConcurrent::run()` with calls to `portutils::runConcurrently()`. That would hide the implementation details and make the code look cleaner/more readable.

veloman-yunkan (changes requested):

src/library.cpp
veloman-yunkan: On another note, since the purpose of this PR is to eliminate compilation warnings rather than improve portability, can't you suppress the warning in this case in a simpler way like

```c++
(void)QtConcurrent::run( [=]() { ... });
```

or

```c++
const auto unused = QtConcurrent::run( [=]() { ... });
Q_UNUSED(unused);
```

kelson42:
@adamlamar Any feedback?

src/library.cpp
adamlamar: Fair enough, its definitely easier to simply ignore return values with `(void)` and `QtConcurrent` is easier to use. I restored `QtConcurrent::run()` and just ignored the return value.

src/kiwixapp.cpp
adamlamar: My particular style (very common in Go) is that the calling function checks the return value and then decides how to handle the error (e.g. log something). Moving the success check into the called function would mean the called function has to decide what to do on failure, which I see as a bit of a general anti-pattern, especially if the goal is to maintain consistent handling during failures.

I solved this by simply ignoring the return value from `loadAndInstallTranslations()`, let me know if that works.

adamlamar:
Sorry @kelson42 this fell off my list. I pushed updates as requested.

src/kiwixapp.cpp
```diff
@@ -60,6 +57,16 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
     setFont(QFont("Selawik"));
 }
 
+bool KiwixApp::loadAndInstallTranslations(QTranslator& translator, const QString& filename, const QString& directory) {
+    bool loaded = false;
+
+    loaded = translator.load(QLocale(), filename, "_", directory);
+    if (loaded) {
```
veloman-yunkan, line 64: Why don't you get rid of the `loaded` variable?

src/kiwixapp.cpp
```diff
@@ -40,17 +40,14 @@ KiwixApp::KiwixApp(int& argc, char *argv[])
...
-    m_qtTranslator.load(QLocale(), "qt", "_",
-                        QLibraryInfo::path(QLibraryInfo::TranslationsPath));
+    QString path = QLibraryInfo::path(QLibraryInfo::TranslationsPath);
 #endif
-    installTranslator(&m_qtTranslator);
-
-    m_appTranslator.load(QLocale(), "kiwix-desktop", "_", ":/i18n/");
-    installTranslator(&m_appTranslator);
+    (void) loadAndInstallTranslations(m_qtTranslator, "qt", path);
+    (void) loadAndInstallTranslations(m_appTranslator, "kiwix-desktop", ":/i18n/");
```
veloman-yunkan, line 50: Since it is not a matter of defining an API that will be hard to change once we release it, I think that at this point it is cleaner to change the return type of `loadAndInstallTranslations()` to `void`.

veloman-yunkan (changes requested):
For the next (hopefully final) review please clean the PR history so that it doesn't contain changes that get reverted (like `LambdaRunnable`).

kelson42:
@adamlamar Thank you for the update. At some point we will need as well to rebase (and fix conflicts)

PR #1054, opened 2024-03-11
Fix #1053: Close Button No Longer Only Close Current Tab on Linux

### LINKED ISSUE #1053: [Linux] Close Buttton on Other Tabs Closes the Displayed Tab Instead
On Linux, the close button on tabs seems to only close the currently displayed tab, irrelevant of which tab the close button is located on. The problem is displayed in this video:

https://github.com/kiwix/kiwix-desktop/assets/39098000/a7b748c6-9bc4-446c-8ee5-200476ee7e81

I supposed this was the code that was handling tab closures, and it seems to be the case. This issue doesn't seem to be present on Windows, which means it might be an OS-dependent behavior. The middle button close functionality works fine, and it uses indexes of clicked tabs to close tabs. I have a method to retrieve a tab index (not merged) based on the widget which might be useful #1048? 
Screenshot from 2024-03-07 12-35-58

ShaopengLin:
The Problem:

**The `CloseTabAction` has nothing wrong. Its purpose is to close the current tab. However, to close non-current tabs, the behaviour is unwanted (No idea why it works on Windows, might be the tab switch event happens first.). We cannot use the [tabCloseRequested()](https://doc.qt.io/qt-6/qtabbar.html#tabCloseRequested) function since we replaced the close button widget with our own. Closing non-current tabs is only possible with mouse clicks**

Changes:

- Replaced `QToolButton` with its new subclass `TabCloseButton` that will overtake the mouse click event.
- Introduce `tabIndexOf` function to get tab index based on a view widget. (Will modify #1048's less flexible version to this)
- `TabCloseButton` takes a view widget and will close the tab based on where the view is located.

Fix #1053

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -69,7 +69,7 @@ TabBar::TabBar(QWidget *parent) :
                 mp_stackedWidget->insertWidget(index, view);
                 emit tabDisplayed(TabType::SettingsTab);
                 insertTab(index,QIcon(":/icons/settings.svg"), gt("settings"));
-                QToolButton *tb = new QToolButton(this);
+                TabCloseButton *tb = new TabCloseButton(this, view);
                 tb->setDefaultAction(KiwixApp::instance()->getAction(KiwixApp::CloseTabAction));
                 setTabButton(index, QTabBar::RightSide, tb);
                 setCurrentIndex(index);
@@ -162,7 +162,7 @@ ZimView* TabBar::createNewTab(bool setCurrent, bool adjacentToCurrentTab)
     }
     mp_stackedWidget->insertWidget(index, tab);
     index = insertTab(index, "");
-    QToolButton *tb = new QToolButton(this);
+    TabCloseButton *tb = new TabCloseButton(this, tab);
     tb->setDefaultAction(KiwixApp::instance()->getAction(KiwixApp::CloseTabAction));
     setTabButton(index, QTabBar::RightSide, tb);
     if (setCurrent) {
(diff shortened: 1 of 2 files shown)


veloman-yunkan (commented):
Thanks for the PR.

This is the first time I look into this part of `kiwix-desktop` code and to me it asks to be refactored. It's not surprising that taking the current design as given you chose to fix the bug with minimal effort. Unfortunately my first impression is that your fix contributes to further deterioration of the design. As the current maintainer of `kiwix-desktop` I'd like to clean-up this part of the code whereupon you can fix the bug in a much nicer way (unless it gets fixed automatically).

ShaopengLin:
> Thanks for the PR.
> 
> This is the first time I look into this part of `kiwix-desktop` code and to me it asks to be refactored. It's not surprising that taking the current design as given you chose to fix the bug with minimal effort. Unfortunately my first impression is that your fix contributes to further deterioration of the design. As the current maintainer of `kiwix-desktop` I'd like to clean-up this part of the code whereupon you can fix the bug in a much nicer way (unless it gets fixed automatically).

I believe the extra toolbutton was uncessary as we can style this instead https://stackoverflow.com/questions/1663561/how-to-style-qtabbar-close-button. This will provide us with a better tooling handling tab closure problems. The QAction to allow key closures I am not sure yet.

ShaopengLin:
@veloman-yunkan One discovery, the button actions are irrelevant other than a tooltip. The actual action is triggered by `m_fileMenu.ADD_ACTION(CloseTabAction);` in `mainmenu.cpp`. I believe what I said above should be the way to go in refactoring. Just need to add some extra tooltips to the QTabBar::CloseButton.

veloman-yunkan:
@ShaopengLin

I am not an expert Qt developer so I could not answer your questions without investigating the matter. I started doing the refactoring and ended up fixing #1053 in a different way. You are welcome to look at my PR (#1057) and comment on it (I wanted to add you as a reviewer but GitHub won't let me do so).

ShaopengLin:
Closed. Fixed already by #1057

PR #1058, opened 2024-03-14
Import file path memorization

### LINKED ISSUE #976: Remember the current directory path in the file explorer over restart
When opening a new zim file via `Open file (ctrl + o)` then the current directory path in the explorer gets reset to the path of the application directory over a restart.

I tested this on kiwix-desktop 2.3.1 on windows 7 and 10.

sgourdas:
Added functionality for the download path to be retrievable and modifiable. On the settings menu, based on the discussion on #976, there is now a new section for the import directory to be set. With this, after a restart, the import directory stays the same based on what the user has selected previously and is saved in settings

Added:
- Logic to keep last location of import remembered
- Option in settings for default import directory

This closes #976

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -225,18 +225,18 @@ void KiwixApp::openZimFile(const QString &zimfile)
 {
     QString _zimfile;
     if (zimfile.isEmpty()) {
-        _zimfile = QFileDialog::getOpenFileName(
-                    getMainWindow(),
-                    gt("open-zim"),
-                    QString(),
-                    "ZIM Files (*.zim);;Splitted ZIM Files (*.zimaa)");
-
-        if (_zimfile.isEmpty()) {
-            return;
+        QString importDir = mp_session->value("zim-import-dir").toString();
+        if (importDir.isEmpty()) { 
+            importDir = QStandardPaths::writableLocation(QStandardPaths::DownloadLocation);
+            if (importDir.isEmpty()) { importDir = QDir::currentPath(); }
         }
-        _zimfile = QDir::toNativeSeparators(_zimfile);
-    }
+        _zimfile = QFileDialog::getOpenFileName(getMainWindow(), gt("open-zim"), importDir, "ZIM Files (*.zim);;Split ZIM Files (*.zimaa)");
 
+        if (_zimfile.isEmpty()) { return; }
+        _zimfile = QDir::toNativeSeparators(_zimfile);
+        QFileInfo fileInfo(_zimfile);
+        mp_session->setValue("zim-import-dir", fileInfo.absolutePath());
+    }
     QString zimId;
     const auto &validZimFile = zimfile.isEmpty() ? _zimfile : zimfile;
     try {


kelson42:
@sgourdas Would you be able please to rebase (and fix conflict) based on `main` so we (finally) can move forward with the review?

veloman-yunkan (commented):
I don't think that we have to have an explicit entry in settings for the import dir. Rather it should be implicit information saved in the session file similar to how its done for the list of open tabs (see #1048).

sgourdas:
@veloman-yunkan I feel like in Kiwix, opening a file is a more common every day task, so it would be better if we clearly show the path in settings. I assume that is also why we have the `Open file` in the mainwindow and not in the main menu dropdown, because it is part of the main workflow. 

In any case I can understand why it could also not be thought as ideal. If you agree, I can go ahead and transfer the logic to the session file.

veloman-yunkan:
@sgourdas It's up to @kelson42 to decide

kelson42:
@sgourdas @veloman-yunkan Sorry for the late feedback. We have to stick to the issue. This does not imply any new setting. We just need to remember over restart the last path used to load manually a ZIM file... and if not set the use the "Download" directory pre-configured by the OS.

kelson42:
@sgourdas Hi, do you still plan to update/complete this PR?

sgourdas:
Hi @kelson42,

Sorry for the lack of feedback, I am out of base until Wednesday.

sgourdas:
@kelson42 I reverted the code surrounding the setting logic and implemented the memorization in the session file.

Also, fixed a small typo `Splitted` to `Split`

kelson42 (changes requested):
Thank you for the update, here my remarks:
* Please secure that if the settings "import/dir" is not set you fallback to the OS "Download" directory and otherwise fallback to the current directory
* Settings naming is not optimal IMHO, something around "default-third-party-zim-directory-path" (sorry, not aware about the settings naming nomenclature) would be clearer
* Please be careful not to add trailing spaces/tabs. You can can the following in your `.gitconfig`
```
[core]
        whitespace = trailing-space,space-before-tab,tab-in-indent
```

@sgourdas Be careful, I have rebased on git `main HEAD` your branch.

sgourdas:
@kelson42 got it. For the naming, would it be ok if we use something smaller like just `zim-import-dir`?

kelson42:
@sgourdas why not ;)

sgourdas:
@kelson42 I am not sure if the current state of the commits is ok based on the rebase with main :') Please advise me

kelson42:
@sgourdas please squash all commits and rebase on main.

kelson42:
@sgourdas Don‘t worry to mich about the Windows CI, but failing Linux CI seems to be a blocker!

sgourdas:
Sorry for the total mess @kelson42. I think we are ok now, hopefully you agree.

kelson42 (approved):
@sgourdas LGTM, Thx.
@veloman-yunkan Can you please do the code review?

veloman-yunkan (approved):
LGTM

PR #1060, opened 2024-03-15
Fix #832: Seperated Bookmark Button from Search Icon

### LINKED ISSUE #832: Bookmark icon not always available
I am using version 2.2.1 on Linux. When I click on a link in a book, an icon appears to the left of the title that allows me to bookmark it:

image

Alternatively, it is possible to search pages and then select one of the search results. Then the article is loaded, but no bookmark icon appears to the left of the title:

image

Trying to make the icon visible by pressing the escape key or clicking in the article does not work. The only solution I have found is to click on an arbitrary link in the article I want to bookmark, and then return to the previous article (Alt-Left). Only then does the icon appear.

I would expect that after clicking on a search result, the magnifying glass icon is replaced by the bookmark icon and the cursor disappears. This is not the case - instead you can even continue typing, which makes no sense because you have selected a search result.

ShaopengLin:
Bug Reason:
The Bookmark mode is not enabled if there is text in the search bar.

Edited Changes:
- Moved the Bookmark button to the right of the search bar.
- The Bookmark button is disabled on non-Zim tabs
- Users can find the Bookmark button in the Edit menu with Shortcut _Ctrl + D_, with Donation Shortcut now as _Ctrl + Shift + D_
- Refactored SearchBar to QToolbar and moved line edits and buttons within the toolbar for less coupling.

~Changes:~
~- Separated the focus check and text existence check. Having text in the search doesn't justify disabling the~ bookmark function.
~- Making the button in search mode transparent, as it doesn't do anything and should trigger search mode. Having it being a focus-out event doesn't make sense.~
~- Making the button in bookmark mode have focus since it is not supposed to trigger search mode.~

~Carried over bugfix: The title isn't displaying properly after these steps:~
~1. searching which changes the page title~
~2. removes text in the search bar~
~3. Click on the article which should be displaying the previous(or a wrong) title.~

Fix #832


ShaopengLin:
@kelson42 Are you aware if this PR or "Carried over bugfix" is related to any other existing issues? I don't seem to find any.

mgautierfr (approved):
LGTM

veloman-yunkan (commented):
I think that we rather have a UI/UX bug that has to be fixed by separating the bookmark button from the search icon.

kelson42:
@ShaopengLin We don't forget you... give us a bit of time please.

veloman-yunkan:
> I think that we rather have a UI/UX bug that has to be fixed by separating the bookmark button from the search icon.

Why don't we (always) display the bookmark button on the right of the searchbar (like in Firefox)?

kelson42:
@ShaopengLin I think your PR works well, but @veloman-yunkan is pretty pertinent. Would that be possible to move the star on the other end of the bar like in Web browser?

ShaopengLin:
I think its doable. Just to clarify do you want it inside the search bar?

Do you want a seperate PR for this? If we want it done in the same PR, I will reword the title of this PR since we are changing the UI.

kelson42:
@ShaopengLin Great. Yes, inside the searchbar (like in browser) and after discussion, we would prefer to have it in current PR.

ShaopengLin:
One thing to note, Chrome browsers all have *Ctrl + D* as the bookmark as a bookmark shortcut. This shortcut is taken already by the donation shortcut. This might be something to discuss if we deem a shortcut necessary.

ShaopengLin:
Found a bug that mouse clicks to the button is falling through to the line edit. Working on a fix...

kelson42:
> @kelson42 One thing to note, Chrome browsers all have _Ctrl + D_ as the bookmark as a bookmark shortcut. This shortcut is taken already by the donation shortcut. This might be something to discuss if we deem a shortcut necessary.

CTRL+D being a standard browser shortcut to bookmark, I suspect this has the priority and we should not change this. Considering that donation menus is not something used on a regular base (would be great ;), you can put another letter which is maybe not that often use, like `Z` for example.

kelson42 (changes requested):
@ShaopengLin I have rebased on git main HEAD! Then tested this new version.

It works IMO basically how it should. At this stage I was not able to find fundamental problems. But a few things should be improved and (partly) are visible in this screenshot:
image

First thing is that the highligted zone margin is not the same on the top and on the right. It should be tested as well for RTL languages (with `./kiwix-desktop -reverse`)
The second thing is that under certain conditions, the pointer is not OK and still show the input cursor in place of the arrow pointer. We have the same problem with the magnify glass.

ShaopengLin:
Since we might be adding a multi-zim search button as well later into the search bar, the search bar looks more like a sub-toolbar of the top widget. I am leaning toward refactoring it to a toolbar with a line edit and tool buttons. @veloman-yunkan

kelson42:
@veloman-yunkan Ready for tech review IMHO

veloman-yunkan (commented):
I looked through the changes briefly but their amount was above my natural ability to comprehend them all at once. Whenever possible, changes must be made in small incremental steps (and I believe this is how you arrived at the final code starting from the original code). Will you please kindly decompose this PR into several smaller changes? Each change must make sense on its own so that it is accurately described by its commit message.

veloman-yunkan (approved):
LGTM

PR #1061, opened 2024-03-15
Fix #967: Search Bar Key Board Short Cut Performs the same as Mouse Click

### LINKED ISSUE #967: Set focus on search bar shortcuts such as f6, ctrl + l, alt + d don't properly work
When I press the focus on search bar shortcut such as f6, ctrl + l, alt + d it focuses on the search bar as expected. Afterward I type some text on the search bar and press enter, nothing happens. What I expect is that it goes to the site I searched for. It does however work properly when I click on the search bar with the mouse. So mouse clicking on the search bar and pressing one of the shortcuts don't have the same behavior, even though they should. Additionally, after pressing one of the shortcuts, the mouse clicking doesn't work as well anymore. So pressing one of the shortcuts changes a state to where the search bar doesn't work. If then I mouse click somewhere on the page to remove focus on the search bar followed by a mouse click on the search bar, the state changes to where the search bar works again.. Also, the search bar doesn't work immediately after focus by mouse click. Afterward typing some text it does work again. That issue might be related to this one with the shortcuts.

Seems like the pull request https://github.com/kiwix/kiwix-desktop/pull/436 didn't implement the shortcuts to focus on the search bar correctly, but I'm not sure. I guess there is something weird with the implementation of the `SearchBar` class.

Currently, I'm running kiwix-desktop 2.3.1 on windows 7. I hope this issue could be reproduced on other operating systems.

ShaopengLin:
Bug reason:
Our current focus in events performs actions based on event types, but we did not pass an event type into a user-defined focus event.

Changes:
- Refactored previous redundant code into the actions menu code.
- Moved the action to SearchBar
- Explicitly stated the reason for the focus in the event is from a shortcut (though the OtherReason would work as well but let's try to be explicit.)

Fix #967

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -439,8 +439,7 @@ void KiwixApp::createActions()
     HIDE_ACTION(SavePageAsAction);
     */
 
-    CREATE_ACTION_SHORTCUT(SearchArticleAction, gt("search-article"), QKeySequence(Qt::CTRL | Qt::Key_L));
-    HIDE_ACTION(SearchArticleAction);
+    CREATE_ACTION_SHORTCUTS(SearchArticleAction, gt("search-article"), QList<QKeySequence>({QKeySequence(Qt::Key_F6), QKeySequence(Qt::CTRL | Qt::Key_L), QKeySequence(Qt::ALT | Qt::Key_D)}));
 
     CREATE_ACTION_SHORTCUT(SearchLibraryAction, gt("search-in-library"), QKeySequence(Qt::CTRL | Qt::SHIFT | Qt::Key_R));
     HIDE_ACTION(SearchLibraryAction);
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -148,18 +148,6 @@ void MainWindow::when_libraryPageDisplayed(bool showed)
     }
 }
 
-void MainWindow::keyPressEvent(QKeyEvent *event)
-{
-    auto key = event->key();
-    auto modifier = event->modifiers();
-    if (key == Qt::Key_F6 ||
-        (key == Qt::Key_L && modifier == Qt::ControlModifier) ||
-        (key == Qt::Key_D && modifier == Qt::AltModifier)) {
-        getTopWidget()->getSearchBar().selectAll();
-        getTopWidget()->getSearchBar().setFocus();
-    }
-    return QWidget::keyPressEvent(event);
-}
 TabBar* MainWindow::getTabBar()
 {
     return mp_ui->tabBar;
--- a/src/mainwindow.h
+++ b/src/mainwindow.h
@@ -27,7 +27,6 @@ public:
     QWidget getMainView();
 
 protected:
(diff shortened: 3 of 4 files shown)


kelson42:
@ShaopengLin Thank you very much for your PR, I finally have assigned a reviewer for it.

src/kiwixapp.cpp
```diff
@@ -403,8 +403,7 @@ void KiwixApp::createActions()
     HIDE_ACTION(SavePageAsAction);
     */
 
-    CREATE_ACTION_SHORTCUT(SearchArticleAction, gt("search-article"), QKeySequence(Qt::CTRL | Qt::Key_L));
-    HIDE_ACTION(SearchArticleAction);
+    CREATE_ACTION_SHORTCUTS(SearchArticleAction, gt("search-article"), QList<QKeySequence>({QKeySequence(Qt::Key_F6), QKeySequence(Qt::CTRL | Qt::Key_L), QKeySequence(Qt::ALT | Qt::Key_D)}));
```
veloman-yunkan, line 406: Did you unhide the `SearchArticleAction` intentionally?

src/kiwixapp.cpp
```diff
@@ -403,8 +403,7 @@ void KiwixApp::createActions()
...
 
     CREATE_ACTION_SHORTCUT(SearchLibraryAction, gt("search-in-library"), QKeySequence(Qt::CTRL | Qt::SHIFT | Qt::Key_R));
     HIDE_ACTION(SearchLibraryAction);
```
veloman-yunkan, line 409: Question to @kelson42 and @mgautierfr: `SearchLibraryAction` doesn't seem to work in the current version of `kiwix-desktop`. Did it ever work, was support for it deliberately dropped or was that functionally accidentally broken?

kelson42: @veloman-yunkan I'm not able to answer at the code level. What is user visible impact?

veloman-yunkan: @kelson42 The question is not about the change. It is about the search-in-library action that is supposed to be activated via the CTRL+SHIFT+R keyboard shortcut.

src/kiwixapp.cpp
ShaopengLin: Yeah, the last commit history says actions not needed for the alpha are hidden and is 6 years ago. Since #436 wanted this feature, I believe it is ok to revive this feature?

src/kiwixapp.cpp
mgautierfr: Probably never worked. Multi-zim search is implemented in libzim but I don't recall we have implemented it in kiwix-desktop.

veloman-yunkan (approved):

src/kiwixapp.cpp
kelson42: For me this ticket is not about the library... so I wonder why we talk about it! Honestly I have not clue what this PR should fixed because I don't understand the issue in a first place. So probably not going to merge it as long as the issue itself is not "cristal clear".

PR #1064, opened 2024-03-21
Fixed dragging of scrollbar in fullscreen

### LINKED ISSUE #913: Windows-only autoscroll feature is broken in full screen
1. Open page with long content.
2. Enter full screen mode (F11).
3. Press the middle mouse button in an empty area of the page (where there are no links, just text or blank space) and drag the mouse up or down. The mouse cursor changes, but the scroll does not work. 

<s>PS: also noticed that I can't even drag the scrollbar with the mouse in full screen mode.</s> see https://github.com/kiwix/kiwix-desktop/issues/985

Version: 2.3.1
Windows 10 Pro 22H2 (build 19045.2364)

### LINKED ISSUE #984: Draging the vertical scrollbar does not work at all in fullscreen mode

sgourdas:
This PR closes #984 
I would like also to help investigate whether it closes #913 . I am not completely sure yet.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -110,14 +110,12 @@ bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
         const auto mouseEvent = static_cast<QMouseEvent*>(event);
         const int tabRegion = getTabBar()->height() + getTopWidget()->height() + 30;
         int clickY = portutils::getY(*mouseEvent);
-
         // We don't have to check for visibilty as calling hide() on a hidden widget, or show() on a non-hidden widget is a no-op
         if (clickY == 0) {
             showTabAndTop();
         } else if(clickY >= tabRegion) {
             hideTabAndTop();
         }
-        return true;
     }
     return false;
 }


kelson42:
@sgourdas Thank you for your PR, we will give a try to it and review the code.

veloman-yunkan (commented):
While testing the fix I noticed another problem - in fullscreen mode the mouse pointer doesn't automatically change its shape depending on content beneath it. If it is easy to fix let's do it in this PR. Otherwise we can merge this PR as is since it definitely fixes the issue being addressed.

sgourdas:
You are right. If you also notice, hovering on the library items doesn't indicate them (making them gray).

In general, it seems like the only event being able to get in the eventFilter is the `MouseMove`. Through experimenting, it seems like the reason for all this behavior is the `return true`. Despite this might seeming like the [correct thing to do](https://forum.qt.io/topic/132796/meaning-of-eventfilter-return-value/2), it actually stops any other movement event from being propagated further. 

Always returning false and letting the event be handled further fixes all related issues, including the scrolling.

If you agree, please inform me to squash, so we can merge.

kelson42:
> I would like also to help investigate whether it closes #913 . I am not completely sure yet.

I propose to just wait and test after merging with the Windows nightly... I doubt a bit that it would fix it.

veloman-yunkan (commented):
LGTM. Please squash before we merge.

kelson42:
> @kelson42 is there a nightly build of kiwix-desktop for Windows?

Yes, at http://download.kiwix.org/nightly/

kelson42:
> > @kelson42 is there a nightly build of kiwix-desktop for Windows?
> 
> Yes, at http://download.kiwix.org/nightly/

Actually seems broken since a week https://github.com/kiwix/kiwix-build/issues/690

kelson42:
> > > I would like also to help investigate whether it closes #913 . I am not completely sure yet.
> > 
> > I propose to just wait and test after merging with the Windows nightly... I doubt a bit that it would fix it.
> 
> @kelson42 I do not know if you have already verified it, since I see #913 closed, but based on todays nightly build I can confirm this fixed it.
> 

Surprisingly great!

PR #1065, opened 2024-03-21
Results were never shown in fulltext search

### LINKED ISSUE #988: After editing keyword in (Fulltext search) results differ from searching for that word from scratch
Version: 
kiwix-desktop_x86_64_2.3.1-2.appimage 

Steps:
Opened wikipedia all maxi.
In search box entered kirtz, as no pages were suggested selected "kirtz (Fulltext search)". Got many results after `No results were found for "kirtz (Fulltext search)"`. Changed kirtz to kurtz. Only one link was listed after `No results were found for "kurtz (Fulltext search)"`. However, if I deleted all text in search bar and then typed "kurtz" I got many pages suggested and when selected  "kurtz (Fulltext search)" then many results were listed after `No results were found for "kurtz"` (btw Kurtz resulted in about the same).

Expected result: editing text in search bar results in same result as entering search query from scratch. As of now I did not find it working as narrowing results either, so current behavior is IMO not useful and strange. 

P.S. Why the phrase "No results were found" in results? What does it mean since for some there were pages and for some there seems to be no pages?

sgourdas:
This closes #988 if there are no additional requests on removing the "(fulltext search)" from the searchbar.

### DIFF
--- a/src/urlschemehandler.cpp
+++ b/src/urlschemehandler.cpp
@@ -133,8 +133,8 @@ UrlSchemeHandler::handleSearchRequest(QWebEngineUrlRequestJob* request)
     }
     kiwix::SearchRenderer renderer(
         search->getResults(start, pageLength),
-        search->getEstimatedMatches(),
-        start);
+        start,
+        search->getEstimatedMatches());
     renderer.setSearchPattern(searchQuery);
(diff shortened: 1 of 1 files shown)


mgautierfr (approved):
This was a silly bug.

Thanks for this fix.

PR #1066, opened 2024-03-21
Changing boomark icon to star and adding tooltip

### LINKED ISSUE #605: Usability: unclear toggle button in the Search box
Found in current 'master' (0183274f)

Screenshot from 2021-02-26 01-24-02

Screenshot from 2021-02-26 01-24-09

'Glasses' icon does not have a tooltip. Also, a placeholder text in Search bar repeats the current page title.

It would be more logical (and useful) if the placeholder text in Search explains which kind of search will be performed, for example:

When the button is on, placeholder says:
> Type here to search in all the articles of the current document

When the button is off, the placeholder says:
> Type here to search in the current page only

As I don't know the use of this toggle button, the texts above are just an examples, they should be written to really describe the current search tool state.

sgourdas:
This closes #605, by changing the reading list add/remove icon from the "glasses" to the "star" icon that is used in most places.
Also added a tooltip to signify the action about to be done.

### DIFF
--- a/resources/i18n/qqq.json
+++ b/resources/i18n/qqq.json
@@ -39,6 +39,8 @@
 	"videos": "{{identical|Video}}",
 	"yes": "{{identical|yes}}",
 	"no": "{{identical|no}}",
+	"add-bookmark": "Represents the action of adding a page as a bookmark",
+	"remove-bookmark": "Represents the action of removing a page from the bookmarks",
 	"open-link-in-web-browser": "Ouvrir le lien dans le navigateur",
 	"download-dir-dialog-msg": "Please note that <code><nowiki>{{DIRECTORY}}</nowiki></code> should be placed on a newline.",
 	"monitor-dir-dialog-title": "\"Monitor\" means \"watch\" in this context. The monitor directory is monitored/watched for new ZIM files.",
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -24,9 +24,11 @@ void SearchButton::set_searchMode(bool searchMode)
     } else {
         auto kiwixApp = KiwixApp::instance();
         if (kiwixApp->isCurrentArticleBookmarked()) {
-            setIcon(QIcon(":/icons/reading-list-active.svg"));
+            setIcon(QIcon(":/icons/star.svg"));
+            setToolTip(gt("add-bookmark"));
         } else {
-            setIcon(QIcon(":/icons/reading-list.svg"));
+            setIcon(QIcon(":/icons/star-active.svg"));
+            setToolTip(gt("remove-bookmark"));
         }
         setIconSize(QSize(25, 25));
     }
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -24,13 +24,13 @@ void SearchButton::set_searchMode(bool searchMode)
     } else {
         auto kiwixApp = KiwixApp::instance();
         if (kiwixApp->isCurrentArticleBookmarked()) {
-            setIcon(QIcon(":/icons/star.svg"));
-            setToolTip(gt("add-bookmark"));
-        } else {
             setIcon(QIcon(":/icons/star-active.svg"));
             setToolTip(gt("remove-bookmark"));
+        } else {
+            setIcon(QIcon(":/icons/star.svg"));
+            setToolTip(gt("add-bookmark"));
         }
-        setIconSize(QSize(25, 25));
+        setIconSize(QSize(32, 32));
     }
 }
 
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -30,6 +30,8 @@
     "set-fullscreen":"Set fullscreen",
     "quit-fullscreen":"Quit fullScreen",
     "table-of-content":"Table of content",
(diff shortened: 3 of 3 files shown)


src/searchbar.cpp
```diff
@@ -24,9 +24,11 @@ void SearchButton::set_searchMode(bool searchMode)
     } else {
         auto kiwixApp = KiwixApp::instance();
         if (kiwixApp->isCurrentArticleBookmarked()) {
-            setIcon(QIcon(":/icons/reading-list-active.svg"));
+            setIcon(QIcon(":/icons/star.svg"));
+            setToolTip(gt("Add to the reading list"));
```
veloman-yunkan, line 28: Move this message into `resources/i18n/en.json` and update `resources/i18n/qqq.json`.

src/searchbar.cpp
```diff
@@ -24,9 +24,11 @@ void SearchButton::set_searchMode(bool searchMode)
...
         } else {
-            setIcon(QIcon(":/icons/reading-list.svg"));
+            setIcon(QIcon(":/icons/star-active.svg"));
+            setToolTip(gt("Remove from the reading list"));
```
veloman-yunkan, line 31: Same here

veloman-yunkan (changes requested):
Please rewrite your commit message so that the GitHub doesn't have to split the title with an ellipsis.

sgourdas:
@veloman-yunkan added the missed tooltip keys and also squashed commits to clean up the mess I made :)

resources/i18n/qqq.json
```diff
@@ -39,6 +39,8 @@
 	"videos": "{{identical|Video}}",
 	"yes": "{{identical|yes}}",
 	"no": "{{identical|no}}",
+	"add-bookmark": "Represents the action of adding a page as a bookmark",
+	"remove-bookmark": "Represents the action of removing a page from the bookmarks",
```
veloman-yunkan, line 43: Keys in `qqq.json` must match those in `en.json`. Actually these key names are better than the ones in `en.json`.

veloman-yunkan (changes requested):
Regarding the commit message - it should consist of a summary (or title) and an optional long description (or additional details). The latter is needed if the information in the summary is not enough. https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html is a more detailed guide to writing commit messages. The only thing in it that I don't fully agree with is the imperative form of the title, however it makes perfect sense if we assume that the title of the commit message is written (and thus the objective of a change is defined) before the code is actually changed. Yet I suspect that that rule rather tries to save a couple of characters which, for the 50-character length limit of the title, is quite sensible. 

Your commit message has a form of just the full description. You can easily convert it to meet the community expectations by:

1. Removing the `-` from the first line (thus making it a title rather than just an item in the unnumbered list of changes)
2. Adding an `Also:` line to the body of the message.

Thus it will read:

```
Changed icon of toggling page to reading list (glasses) to a star

Also:

- Added tooltip for add/remove on reading list
```

src/searchbar.cpp
```diff
@@ -24,9 +24,11 @@ void SearchButton::set_searchMode(bool searchMode)
     } else {
         auto kiwixApp = KiwixApp::instance();
         if (kiwixApp->isCurrentArticleBookmarked()) {
-            setIcon(QIcon(":/icons/reading-list-active.svg"));
+            setIcon(QIcon(":/icons/star.svg"));
+            setToolTip(gt("add-reading-list"));
         } else {
-            setIcon(QIcon(":/icons/reading-list.svg"));
+            setIcon(QIcon(":/icons/star-active.svg"));
+            setToolTip(gt("remove-reading-list"));
```
veloman-yunkan, line 31: You forgot to update the message keys here.

veloman-yunkan (changes requested):
Please also squash your commits when you fix the new comment.

veloman-yunkan (commented):
There are still two commits in your PR. Other than that it looks good.

sgourdas:
This was embarrassing... Going to study a bit on git... :)

At least I now understand what and where to look. LGTMBTWRE

veloman-yunkan (approved):
Great!

Let this PR be merged. But if you are fast enough you can fix/revert the commit message (to get rid of the ugly ellipsis because of the too long title)

kelson42 (changes requested):
I think this PR suffers from many problem/weanesses (even if part of the problem might already have been there first):
* The pointer is not appropriate when over the icon, should be an arrow
* The icon is too small, should be significantly bigger to be easy to click
* To me the icon is also not centered properly, it seem to be aligned on the text base position on the right
* Default value is wrong, should not be "full" as per default pages are not bookmarked (I suspect both icons full/empty have been inverted)... tooltip also wrong like the icon.

sgourdas:
Regarding these issues:
- For me, when hovering over the icon the pointer is an arrow, same as other icons
- I agree. 32x32 looks right to me. How do you see it?
- Initially it seemed off to me as well. After making it bigger I think it looks normal, but would like a second opinion on if we should center it based on the searchbar height
- Fixed

kelson42:
@sgourdas Kind of a miracle - as you don't have fixed all the points - but now everything looks good to me. Thank you!

PR #1068, opened 2024-03-21
On undefined fullscreen key sequence, hardcode to F11

### LINKED ISSUE #1067: REGRESSION: Fullscreen shortcut (F11) not working on WSL
I am using WSL (with Ubuntu on Windows 11) and I noticed that F11 is not working. As shown below, even while navigating the menu, a shortcut for the action is not available (pressing the button works).

Screenshot_3

Printing the value of  `QKeySequence::FullScreen` it gives an empty string. So I assume that because it is launching from WSL into the Windows environment, it cannot handle deciding what platform it is in, so it ends up undefined.

I will make a fix for the issue, since it might be possible to happen in other instances as well and it also helps during development.

sgourdas:
This closes #1067, by adding a condition to check if the fullscreen key sequence is undefined and in that case hardcode it with F11

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -413,8 +413,13 @@ void KiwixApp::createActions()
     mpa_actions[FindInPageAction]->setShortcuts({QKeySequence::Find, Qt::Key_F3});
     connect(mpa_actions[FindInPageAction], &QAction::triggered,
             this, [=]() { getTabWidget()->openFindInPageBar(); });
-
-    CREATE_ACTION_ICON_SHORTCUT(ToggleFullscreenAction, "full-screen-enter", gt("set-fullscreen"),  QKeySequence::FullScreen);
+    
+    bool fullScreenKeyUndefined = (QKeySequence(QKeySequence::FullScreen).toString()).isEmpty();
+    if (fullScreenKeyUndefined) {
+        CREATE_ACTION_ICON_SHORTCUT(ToggleFullscreenAction, "full-screen-enter", gt("set-fullscreen"), Qt::Key_F11);
+    } else {
+        CREATE_ACTION_ICON_SHORTCUT(ToggleFullscreenAction, "full-screen-enter", gt("set-fullscreen"), QKeySequence::FullScreen);
+    }
     connect(mpa_actions[ToggleFullscreenAction], &QAction::toggled,
             this, [=](bool checked) {
         auto action = mpa_actions[ToggleFullscreenAction];
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -414,12 +414,10 @@ void KiwixApp::createActions()
     connect(mpa_actions[FindInPageAction], &QAction::triggered,
             this, [=]() { getTabWidget()->openFindInPageBar(); });
     
-    bool fullScreenKeyUndefined = (QKeySequence(QKeySequence::FullScreen).toString()).isEmpty();
-    if (fullScreenKeyUndefined) {
-        CREATE_ACTION_ICON_SHORTCUT(ToggleFullscreenAction, "full-screen-enter", gt("set-fullscreen"), Qt::Key_F11);
-    } else {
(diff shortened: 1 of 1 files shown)


src/kiwixapp.cpp
```diff
@@ -413,8 +413,13 @@ void KiwixApp::createActions()
...
             this, [=]() { getTabWidget()->openFindInPageBar(); });
-
-    CREATE_ACTION_ICON_SHORTCUT(ToggleFullscreenAction, "full-screen-enter", gt("set-fullscreen"),  QKeySequence::FullScreen);
+    
+    bool fullScreenKeyUndefined = (QKeySequence(QKeySequence::FullScreen).toString()).isEmpty();
+    if (fullScreenKeyUndefined) {
+        CREATE_ACTION_ICON_SHORTCUT(ToggleFullscreenAction, "full-screen-enter", gt("set-fullscreen"), Qt::Key_F11);
+    } else {
+        CREATE_ACTION_ICON_SHORTCUT(ToggleFullscreenAction, "full-screen-enter", gt("set-fullscreen"), QKeySequence::FullScreen);
+    }
```
veloman-yunkan, line 422: 1. Why don't you use [`QKeySequence::isEmpty()`](https://doc.qt.io/qt-5/qkeysequence.html#isEmpty)?
2. The difference between your two `CREATE_ACTION_ICON_SHORTCUT()` lines is only in the last argument and I'd rather have the duplication eliminated by isolating the said difference into a constant that would be passed as the last argument:

```c++
    const auto fullScreenKeySeq = QKeySequence(QKeySequence::FullScreen).isEmpty()
                                ? Qt::Key_F11
                                : QKeySequence::FullScreen;
```

veloman-yunkan (changes requested):
LGTMBCBMSB

Can you decode that abbreviation based on my feedback below? :smiley:

src/kiwixapp.cpp
sgourdas: It "Looks Good To Me But Could Be Made Simpler By Moving Said Behavior" to me as well! 😃

Maybe it is not a good idea to do PRs late at night after work 😢

Totally agree with you.

veloman-yunkan: > It "Looks Good To Me But Could Be Made Simpler By ... Moving Said Behavior" to me as well!

Almost there! Actually, I meant ""Looks Good To Me But Can Be Made Slightly Better"

sgourdas: Ah... I thought it would have to be continued below :| Not for me ig 

LGTMNLMKITM 🫡

veloman-yunkan (approved):
Thanks for the PR!

veloman-yunkan:
@sgourdas Can you please squash your commit into one?

veloman-yunkan:
> @sgourdas Can you please squash your commits into one?

@sgourdas Please ignore my previous comment since the PR has already been merged.

sgourdas:
Will keep in mind for the next one @veloman-yunkan. Thanks for the guidance.

veloman-yunkan:
@sgourdas Since you are going to take my (cancelled) request as a guidance (which it was not) I will elaborate it into one. Squashing all commits of a PR is not a rule. The PR at the time of the merge should consist of a series of changes (commits) representing the cleanest way of achieving the PR's purpose. Commits in the PR must be small to an extent that the essence of every change can be accurately summarized in the (short) title of the commit message. Intermediate changes (like addressing review comments of cosmetic nature, fixing silly bugs introduced be earlier commits, etc) must not pollute the history of the main branch and must be properly squashed into other commits.

src/kiwixapp.cpp
veloman-yunkan: > LGTMNLMKITM 🫡

I can debreviate it in one of three ways but I am not sure which one it actually stands for: 

1. Looks good to me now. Let me know if that's marvelous.
2. Lest `git` triggers mysterious nonsense, let my kittens ingest their milk.
3. Little Grandpa took my neighbour's last mathematical kidney in the meantime.

PR #1069, opened 2024-03-26
Implemented application language change behavior and setting

### LINKED ISSUE #487: Enable support for service workers for new replay system
Since kiwix-desktop use a custom scheme, to enable support for the service-worker based replay, the following flag should be enabled:
```
scheme.setFlags(QWebEngineUrlScheme::ServiceWorkersAllowed);
```
https://doc.qt.io/qt-5/qwebengineurlscheme.html

Hopefully this will make it work, though there is a chance that it may not, and regular `http://localhost` server may be needed..

Here is an example ZIM file that works with kiwix-serve in the browser, and will hopefully work with service workers enabled
[zimreplay.zim](https://github.com/kiwix/kiwix-desktop/files/4915909/zimreplay.zip) (rename .zip -> .zim as github doesn't allow .zim attachments)

Here's how it looks like with kiwix serve (replay of a twitter.com/netpreserve using a baked in warc file from: https://github.com/webrecorder/replayweb.page/blob/master/examples/netpreserve-twitter.warc)
Screen Shot 2020-07-13 at 12 58 15 PM

sgourdas:
I have added logic to implement language change from the settings menu with a dropdown menu trying to close #487

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -256,6 +256,8 @@ void ContentManager::setLanguages()
         for (auto language : languageData) {
             auto langCode = QString::fromStdString(language);
             auto selfName = QString::fromStdString(kiwix::getLanguageSelfName(language));
+            // qDebug() << langCode;
+            // qDebug() << selfName << "!";
             languages.push_back({langCode, selfName});
         }
         m_languages = languages;
--- a/src/contentmanagerside.cpp
+++ b/src/contentmanagerside.cpp
@@ -97,5 +97,9 @@ void ContentManagerSide::setCategories(QStringList categories)
 
 void ContentManagerSide::setLanguages(ContentManager::LanguageList langList)
 {
+    qDebug() << "List of languages:";
+    for (const QPair<QString, QString>& languagePair : langList) {
+        qDebug() << languagePair.first;
+    }
     mp_languages->setSelections(langList, KiwixApp::instance()->getSettingsManager()->getLanguageList());
 }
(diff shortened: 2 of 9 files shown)


kelson42:
@sgourdas Here again sorry for the lack of feedback, this is a bit more PR than usual :) Can you please rebase it on `main` and fix the conflicts?

kelson42 (changes requested):
This PR works "in principle" but suffers from many problems:
* The user is not invited to restart the app after setting a new language in the settings
* The languages in the select box are all in English. Either it should be in the language of the current UI OR each language should be written in its own language
* Many language are just "C" 
image

sgourdas:
I have refactored the logic to dynamically import language codes from the available files in the `i18n` folder. The following issues have been observed:
- `io.json` and `nb.json` do not have a name field, necessary for getting the native name of the language. Regardless I have a check in place to skip files with this issue, but I think we might need to make sure that only substantially complete languages are in this directory, to be used.
- A lot of languages are incomplete like `yo.json` or `bcl.json` where even the `name` field of the json file is incorrect and misleading. One check I have in place to mitigate these issues is if `QLocale(code).language()` returns `C` I skip this language.
- Languages like Chinese are unable to be rendered with our font and I am not sure how to proceed with this. 
image
Some examples are `বাংলা"` (bn.json), `हिन्दी` (hi.json), `英語` (ja.json), `한국어` (ko.json). If you want the complete list please ask me.

PR #1070, opened 2024-03-27
Fix #697 : Clicking on a PDF link opens directly in a PDF reader

### LINKED ISSUE #697: PDF file dialog box should allow to open PDF file directly with PDF reader
More of a feature request than an issue, Currently the behavior of the Desktop app differs to the UWP and the browser extensions. In the UWA app, PDF files are automatically saved to a folder in the Downloads area, and opened automatically. 

Kiwix deskop opens a dialog, and asks the user where they would like to save the PDF. I would imagine that in the vast majority of situations, a user clicking on a PDF in Kiwix would rather it just open than be asked where to save it to. 

I would request that the files either be stored in a temporary location and automatically opened, or behave like the UWP apps and the browser extensions; storing the file in the downloads location and then opening the file in the shell associated application.

vinurk:
Fixes #697 

Checks for the mimetype to verify that its a PDF, and downloads the PDF directly and opens it in a PDF reader.

Added functionality where, clicking a PDF link does the following:
1. Checks if there exists a downloaded copy of the PDF. If yes skip to step 3.
2. Downloads the PDF into the Download directory set by the user.
3. Open the PDF in a PDF reader, using `<QDesktopServices>` as suggested by @juuz0

### DIFF
--- a/src/kprofile.cpp
+++ b/src/kprofile.cpp
@@ -43,7 +43,6 @@ void KProfile::startDownload(QWebEngineDownloadRequest* download)
     QString dmimeType = download->mimeType();
     
     if (dmimeType.toStdString()=="application/pdf"){ 
-        
         auto downloadPath = getSettingsManager()->getDownloadDir();
         fileName = downloadPath + "/" + defaultFileName;
         setDownloadInfo(fileName,dmimeType);
--- a/src/kprofile.cpp
+++ b/src/kprofile.cpp
@@ -21,14 +21,68 @@ SettingsManager* getSettingsManager()
     return KiwixApp::instance()->getSettingsManager();
 }
 
-void KProfile::setDownloadInfo(const QString &fileName, const QString &mimeType)
+void KProfile::setFileOpenChoice(QString fileName, bool choice)
 {
-    m_downloadInfo = std::make_pair(fileName, mimeType);
+    m_fileOpenChoice = std::make_pair(fileName,choice);
 }
 
-KProfile::DownloadInfo KProfile::getDownloadInfo() const
+KProfile::DownloadInfo KProfile::getFileOpenChoice() const
 {
-    return m_downloadInfo;
+    return m_fileOpenChoice;
+}
+
+bool KProfile::isNonReadableFile(QString mimeType)
+{
+    if (mimeType.toStdString()=="application/pdf"){ 
+        return true;
+    }
+    return false;
+}
+
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+    QString KProfile::getFileName(QWebEngineDownloadItem* download)
+#else
+    QString KProfile::getFileName(QWebEngineDownloadRequest* download)
+#endif
+{
+    int ret;
+    bool nrFile = isNonReadableFile(download->mimeType());
+    QString defaultFileName = download->url().fileName();
+    QString fileName("");
+    
+    if(nrFile){
+        QMessageBox msgBox;
+
+        msgBox.setText(gt("download-or-open"));
+        msgBox.setStandardButtons(QMessageBox::Open | QMessageBox::Save | QMessageBox::Close);
+        msgBox.button(QMessageBox::Save )->setText(gt("download"));
+        msgBox.setDefaultButton(QMessageBox::Save);
+        ret = msgBox.exec();
+
+        if(ret == QMessageBox::Open){
+            QTemporaryFile tempFile;
+
+            if (tempFile.open()) {
+                fileName = tempFile.fileName();
+                tempFile.close();
+            }
+            setFileOpenChoice(fileName,true);
+        }
+    }
+
+    if(!nrFile || ret == QMessageBox::Save){
+        fileName = QFileDialog::getSaveFileName(KiwixApp::instance()->getMainWindow(),
+                                                            gt("save-file-as-window-title"), defaultFileName);
+        if (!fileName.isEmpty()) {
+            QString extension = "." + download->url().url().section('.', -1);
+            if (!fileName.endsWith(extension)) {
+                fileName.append(extension);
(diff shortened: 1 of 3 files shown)


kelson42:
@vinurk Thank you very much your PR, I will give a try and come back to you.

src/kprofile.cpp
```diff
@@ -13,22 +16,59 @@ KProfile::KProfile(QObject *parent) :
...
+        auto downloadPath = getSettingsManager()->getDownloadDir();
+        fileName = downloadPath + "/" + defaultFileName;
+        setDownloadInfo(fileName,dmimeType);
+        
+        QFile file(fileName);
+
+        if (file.exists()){
+            QDesktopServices::openUrl(QUrl::fromLocalFile(fileName));
+            return;
+        }
```
veloman-yunkan, line 55: What if different PDF resources (with different URLs) have the same file-name? Then after downloading and opening one of them, when trying to download/open the other one you will actually be presented with the wrong file!

veloman-yunkan (commented):
I don't have a strong opinion on how this functionality should work but I spotted a potential problem with your implementation.

src/kprofile.cpp
vinurk: I didn't think about that, maybe we need to save the file with a unique name like a timestamp maybe?

veloman-yunkan: I am against timestamps. Better mimic the behaviour of existing familiar apps (web-browsers, in this case).

vinurk:
@mgautierfr 
Initially when the user  clicks on "open" it downloads in a temporary directory, and opens it.

My question is: how should we handle subsequent "open" clicks for the same PDF?

-  Should we redownload the file (creating duplicates)?
- Or, can we detect a previously downloaded version and open that instead?

If detection is preferred, how can we achieve this considering:

- Unique filenames are required on download.
- We need to track the link between the unique filename and the original PDF URL.

Could you give any suggestion on how to go about it.

vinurk:
@mgautierfr
I got the part where the user clicks the "download" more than one times.

> When opening the file (so downloading in tmp dir), it just add a `.N` to file to have a unique file name.

So if i understand correctly, if a user clicks on "open" on a file for example demo.pdf, first time it downloads demo.pdf (in tmp dir), then demo1.pdf, demo2.pdf, ... demoN.pdf so on..., even if the same pdf maybe present as demoN-1.pdf ?

kelson42:
@vinurk Thank you for your PR and your effort to understand the requirements. It should have been clearer indeed.

Let me try to do that:
* First of all this is not really about PDF, this is about anything Kiwix can not display (could be an MP3 file for example). As such the code should probably better be generic.
* Once recognised that Kiwix can not deal itself with this mime-type, then the user should ideally be given two choices: either "save the file locally to a place of its choice" OR "open it with the application associated to this mime-type in the OS".
* To save the file locally, Kiwix should open the usual filepicker
* To open the file with a third party app, it should ideally work like in a browser although I believe there is no reall standard way of doing this but we could try to (1) save the file to OS temporary directly (2) if the file already exists, I would append something like `(1)`, `(2}` ( or `.N` like said by @mgautierfr ) etc until the file does not exists just before the the extension part.

Hope @veloman-yunkan and @mgautierfr this respects your respective opinions.

vinurk:
@kelson42 @mgautierfr @veloman-yunkan 
Thank you all for your suggestions, I think I'll try to implement it now

vinurk:
@kelson42 @mgautierfr @veloman-yunkan

sorry for taking long

- I have added commits only for recognizing pdf files, (maybe another ticket for identify other non-readable files ?), I have added a function `isNonReadableFile()` in which other files can be added later on.

- Now when user clicks on the pdf link, it asks whether to download or open.
- When the user clicks on open, it uses the `QTemporaryFile` to create a temporary file, which guarantees a unique filename, then it opens with the pdf reader.
- If the user clicks on save or a non-readable file by kiwix which is not recognized by `isNonReadableFile()` for eg: .epub file. it opens the filepicker.

For now please ignore the wrong order of buttons on the message box. (The one which asks whether to open or download).

I have used `msgBox.setStandardButtons(QMessageBox::Open | QMessageBox::Save | QMessageBox::Close);` to add the buttons, but whatever order i add them, the button with a RejectRole is always placed second. 

Is there anyway to use the standardButtons and still get it in the correct order, or we need to make a cusotm message box. Any suggestions would be helpful.

kelson42:
@vinurk Your PR seems to do the job. Congrats, but you need:
* to have a better way to detect if the file can be rendered by the browser or not... to make the whole stuff generic
* Have a better design of the dialog box (see how the other are made)

Is that doable?

kelson42:
> * Actually is there really a need to detect non readable files? Since only non-readable files would lead to this section of the code, while all other files would simply be opened in the web view.

OK, I don't get it, but if any content which is not viewable in the WebView triggers that dialog, this is fine to me.

kelson42:
@vinurk I have rebased the PR to keep the sync with main

kelson42:
@juuz0 Vomurk does not respond and you have already experiences with customizing the appearance of dialog box, is that a PR you could complete?

vinurk:
@kelson42 Sorry for not replying, I was really busy, It would be best if anyone else can take up this PR, as I don't know when I will be able to continue. AFAIR dialog is the only part pending.

kelson42:
@juuz0 That would be awesome, we are finaly almost ready to release 2.4.0 and this is one of the very few tickets left to implement.

kelson42:
@juuz0 Any news?

juuz0:
@kelson42 https://www.theguardian.com/world/article/2024/jun/01/sensor-error-means-new-delhi-heatwave-record-overstated-by-3c


(news)

PR #1079, opened 2024-04-10
Library list pinning behavior

### LINKED ISSUE #1078: UI/UX issues with the reading list
Currently the reading list is implemented as a sidebar of fixed width in a content tab. When switching to another tab the reading list disappears (and is not restored when returning to the same tab). The sidebar nature of the reading list suggests that selecting a bookmark should load it in the current tab, yet a new tab is opened instead.

The reading list cannot be opened (directly) while on the library tab (the workaround is to open a new empty tab, whereupon the reading list button is activated, but, again, the empty tab is not used to display the target of the bookmark).

sgourdas:
This refactors the behavior of tab changes, to make sure the reading list stays pinned until the user closes it (should be hidden in library at all times)

Fixes #1078

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -32,7 +32,7 @@ MainWindow::MainWindow(QWidget *parent) :
     connect(app->getAction(KiwixApp::ToggleFullscreenAction), &QAction::triggered,
             this, &MainWindow::toggleFullScreen);
     connect(app->getAction(KiwixApp::ToggleReadingListAction), &QAction::toggled,
-            this, &MainWindow::when_ReadingList_toggled);
+            this, &MainWindow::readingListToggled);
     connect(app->getAction(KiwixApp::AboutAction), &QAction::triggered,
             mp_about, &QDialog::show);
     connect(app->getAction(KiwixApp::DonateAction), &QAction::triggered,
@@ -57,12 +57,11 @@ MainWindow::MainWindow(QWidget *parent) :
             mp_ui->mainToolBar, &TopWidget::updateBackForwardButtons);
     connect(mp_ui->tabBar, &TabBar::tabDisplayed,
             this, [=](TabType tabType) {
-                when_libraryPageDisplayed(tabType == TabType::LibraryTab);
+                tabChanged(tabType);
             });
 
     connect(mp_ui->tabBar, &TabBar::currentTitleChanged,
             &(mp_ui->mainToolBar->getSearchBar()), &SearchBar::currentTitleChanged);
-
     // This signal emited more often than the history really updated
     // but for now we have no better signal for it.
     connect(mp_ui->tabBar, &TabBar::currentTitleChanged,
@@ -120,8 +119,7 @@ bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
     return false;
 }
 
-
-void MainWindow::when_ReadingList_toggled(bool state)
+void MainWindow::readingListToggled(bool state)
 {
     if (state) {
         mp_ui->sideBar->setCurrentWidget(mp_ui->readinglistbar);
@@ -132,19 +130,16 @@ void MainWindow::when_ReadingList_toggled(bool state)
     }
 }
 
-void MainWindow::when_libraryPageDisplayed(bool showed)
+void MainWindow::tabChanged(TabType tabType) 
 {
-    auto app = KiwixApp::instance();
-
-    // When library sidebar appeared, or hidden, reading list is always hidden.
-    app->getAction(KiwixApp::ToggleReadingListAction)->setChecked(false);
-
-    if (showed) {
+    QAction *readingList = KiwixApp::instance()->getAction(KiwixApp::ToggleReadingListAction);
+    if (tabType == TabType::SettingsTab) { 
+        mp_ui->sideBar->hide();    
+    } else if(tabType == TabType::LibraryTab) { 
         mp_ui->sideBar->setCurrentWidget(mp_ui->contentmanagerside);
         mp_ui->sideBar->show();
-    }
-    else {
-        mp_ui->sideBar->hide();
+    } else { 
+        readingListToggled(readingList->isChecked());
     }
 }
 
--- a/src/mainwindow.h
+++ b/src/mainwindow.h
@@ -31,8 +31,8 @@ protected:
 
 private slots:
     void toggleFullScreen();
-    void when_ReadingList_toggled(bool state);
-    void when_libraryPageDisplayed(bool showed);
+    void tabChanged(TabBar::TabType);
+    void readingListToggled(bool state);
     void hideTabAndTop();
     void showTabAndTop();
 
--- a/src/readinglistbar.cpp
+++ b/src/readinglistbar.cpp
@@ -13,8 +13,19 @@ ReadingListBar::ReadingListBar(QWidget *parent) :
     ui->setupUi(this);
     connect(KiwixApp::instance()->getLibrary(), &Library::bookmarksChanged,
             this, &ReadingListBar::setupList);
-    connect(ui->listWidget, &QListWidget::itemActivated,
-            this, &ReadingListBar::on_itemActivated);
+    connect(ui->listWidget, &QListWidget::itemClicked,
+            this, &ReadingListBar::onItemClicked);
+    connect(ui->listWidget, &QListWidget::itemDoubleClicked,
+            this, &ReadingListBar::onItemDoubleClicked);
+    connect(ui->listWidget, &QListWidget::itemPressed, 
+            this, [this](QListWidgetItem* item) {
+                onItemPressed(item, QGuiApplication::mouseButtons());
+            });
+    connect(ui->listWidget, &QListWidget::itemActivated, 
+            this, [this](QListWidgetItem* item) {
+                onItemActivated(item, QGuiApplication::mouseButtons());
+            });
+
     setupList();
 
     ui->label->setText(gt("reading-list-title"));
@@ -24,7 +35,6 @@ ReadingListBar::~ReadingListBar()
 {
     delete ui;
 }
-
 
 void ReadingListBar::setupList()
 {
@@ -60,7 +70,41 @@ void ReadingListBar::setupList()
     }
 }
 
-void ReadingListBar::on_itemActivated(QListWidgetItem* item)
+// Receives single clicks
+void ReadingListBar::onItemClicked(QListWidgetItem* item)
+{
+    clickKind = 1;
+    QTimer::singleShot(QApplication::doubleClickInterval(), [=]() { // Give time for a double click be handled
+        if (clickKind == 1) {
+            openUrl(item, false);
+        }
+    });
+}
+
+// Receives double clicks
+void ReadingListBar::onItemDoubleClicked(QListWidgetItem* item)
+{ 
+    clickKind = 2;
+    openUrl(item, true);
+}
+
+// Receives single and middle click
+void ReadingListBar::onItemPressed(QListWidgetItem* item, Qt::MouseButtons buttons)
+{
+    if (buttons & Qt::MiddleButton) {
+        openUrl(item, true);
+    }
+}
+
+// Receives left clicks and activation key
+void ReadingListBar::onItemActivated(QListWidgetItem* item, Qt::MouseButtons buttons)
+{
+    if (!buttons) { // clicks are handled elsewhere, handle only the activation key case 
+        openUrl(item, true);
+    }
+}
+
+void ReadingListBar::openUrl(QListWidgetItem* item, bool newTab)
 {
     int index = ui->listWidget->row(item);
     auto bookmark = KiwixApp::instance()->getLibrary()->getBookmarks(true).at(index);
@@ -68,5 +112,5 @@ void ReadingListBar::on_itemActivated(QListWidgetItem* item)
     url.setScheme("zim");
     url.setHost(QString::fromStdString(bookmark.getBookId())+".zim");
     url.setPath(QString::fromStdString(bookmark.getUrl()));
-    KiwixApp::instance()->openUrl(url);
+    KiwixApp::instance()->openUrl(url, newTab);
 }
--- a/src/readinglistbar.h
+++ b/src/readinglistbar.h
@@ -18,9 +18,14 @@ public:
 
 public slots:
     void setupList();
-    void on_itemActivated(QListWidgetItem *item);
+    void onItemClicked(QListWidgetItem* item);
+    void onItemDoubleClicked(QListWidgetItem *item);
+    void onItemPressed(QListWidgetItem* item, Qt::MouseButtons buttons);
+    void onItemActivated(QListWidgetItem *item, Qt::MouseButtons buttons);
 private:
     Ui::readinglistbar *ui;
+    int clickKind;
+    void openUrl(QListWidgetItem* item, bool newTab);
 };
 
 #endif // READINGLISTBAR_H


kelson42:
@sgourdas Thank you for you PR, we will review it.

kelson42:
Beside [my comment](https://github.com/kiwix/kiwix-desktop/issues/1078#issuecomment-2054066835), it looks good from a user perspective.

sgourdas:
This was a bit more complicated than what I would want it, for code clarity, but now the behavior should be the one you mentioned.

To distinguish between a double click and not handle it as two clicks as well, I had to put a check in place so single clicks check if a double click happened at the near time they were emitted.

src/readinglistbar.cpp
```diff
@@ -60,13 +65,43 @@ void ReadingListBar::setupList()
...
+
+void ReadingListBar::on_itemClicked(QListWidgetItem* item, Qt::MouseButtons buttons) // Handles single and middle click
+{   
+    if(lastDoubleClickTime.isValid()) {
+        QDateTime thisClickTime = QDateTime::currentDateTime();
+        // Give time for a double click be handled
+        QEventLoop evtLoop;
+        int doubleClickInterval = QApplication::doubleClickInterval();
+        QTimer::singleShot(doubleClickInterval, &evtLoop, &QEventLoop::quit);
+        evtLoop.exec();
```
veloman-yunkan, line 82: Isn't this merely a more convoluted way of doing `QThread::msleep(QApplication::doubleClickInterval())`;

veloman-yunkan (changes requested):
Testing revealed the following issues:
- During the very first double click the bookmark is opened both in the current tab and in a new tab.
- Selecting a bookmark with keyboard (Tab keys) and activating it with the "Enter" key opens the bookmark in a new tab. I guess that was intended.

Overall, as you noted, the code lacks clarity. I think that having three functions `onItemClicked()`, `onItemSingleClicked()` and `onItemDoubleClicked()` will make the design more comprehensible. `onItemClicked()` will be responsible solely for distinguishing single-clicks from double-clicks and calling the corresponding self-documenting handler.

Regarding the history of this PR - please make sure that it consists of only two commits with accurate titles and useful descriptions.

src/mainwindow.cpp
```diff
@@ -132,20 +144,12 @@ void MainWindow::when_ReadingList_toggled(bool state)
...
-
-    if (showed) {
-        mp_ui->sideBar->setCurrentWidget(mp_ui->contentmanagerside);
-        mp_ui->sideBar->show();
-    }
-    else {
-        mp_ui->sideBar->hide();
-    }
+    QAction *readingList = KiwixApp::instance()->getAction(KiwixApp::ToggleReadingListAction);
+    readingList->setChecked(false); // Reading list is always disabled in library view
```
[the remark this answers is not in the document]
veloman-yunkan: I just found another small inconsistency - though the reading list keeps being displayed beside a settings tab it cannot be opened or closed while on a settings tab.

src/mainwindow.h
```diff
@@ -31,15 +31,17 @@ class MainWindow : public QMainWindow
...
+    void readingListToggled(bool state);
+    void libraryPageDisplayed();
     void hideTabAndTop();
     void showTabAndTop();
 
 private:
     Ui::MainWindow *mp_ui;
     About     *mp_about;
     LocalKiwixServer *mp_localKiwixServer;
+    bool readingListRemainActive = false;
```
veloman-yunkan, line 44: Do we really need this hidden state variable? Can't the state of the reading list be fully maintained in the `ToggleReadingListAction` action (so that it stays checked even when switching to tabs alongside which the reading list isn't supposed to be displayed).

src/mainwindow.cpp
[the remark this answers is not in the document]
sgourdas: > I don't get why the reading list is incompatible with the library view. In a newly started `kiwix-desktop` app, why can't I see the list of bookmarks without having to open a content ~(or settings)~ tab?

On top of my [comment in the issue](https://github.com/kiwix/kiwix-desktop/issues/1078#issuecomment-2059321426), after implementing the single click behavior in commit [445284a](https://github.com/kiwix/kiwix-desktop/pull/1079/commits/445284a264074caf3d3ebf64180fd76fe45d554d), this would not act correctly with the library view (or settings).

> I just found another small inconsistency - though the reading list keeps being displayed beside a settings tab it cannot be opened or closed while on a settings tab.

Can confirm. For me, the reading list should be hidden when viewing the settings as well, but would like feedback on it.

src/mainwindow.h
veloman-yunkan: I understand that but, IMHO, the checked state of the reading list button can serve the same purpose and this new variable violates the DRY principle. Whether the reading list should be displayed on the library and/or settings tab is a different matter. If the users understand that the reading list is visible only on content tabs, they can easily accept the fact that the reading list button is pressed but there is no reading list.

src/mainwindow.cpp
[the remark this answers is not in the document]
veloman-yunkan: > On top of my [comment in the issue](https://github.com/kiwix/kiwix-desktop/issues/1078#issuecomment-2059321426), after implementing the single click behavior (open in same tab) in commit [445284a](https://github.com/kiwix/kiwix-desktop/pull/1079/commits/445284a264074caf3d3ebf64180fd76fe45d554d), this would not act correctly with the library view (or settings).

I think that it will work. `TabBar::openUrl()` opens a new tab when called while the current tab is not a content tab. Thus on a library or settings tab a single-click will open a new tab, which is logical.

src/mainwindow.h
sgourdas: I agree and I can see where the confusion is coming from because I was thinking the same when I started developing this. The existence of this variable derives from the fact that the reading list shouldn't be displayed on the library. Let me explain.

Specifically, when `readingListRemainActive` is `true` it denotes that the library is hidden but should remain active on the first possible moment. This is why the value of it when navigating *inside* the library becomes true.

 The reason we cannot be reliant on the checked state of the reading list button is becausem when entering the library, that is checked to false. So, if we don't have `readingListRemainActive` and a user activates the reading list and enters the library the reading list checked value is going to be set to false. After navigating away from the library we have no way to know the if the reading list was active to keep it open.

I hope this clears things up a bit and is not confusing.

sgourdas:
> During the very first double click the bookmark is opened both in the current tab and in a new tab.

I cannot replicate this. For me it only opens a new tab with the double clicked bookmark.

> Selecting a bookmark with keyboard (Tab keys) and activating it with the "Enter" key opens the bookmark in a new tab. I guess that was intended.

Yes this is intended and is supported by the previous behavior.

> Overall, as you noted, the code lacks clarity. I think that having three functions onItemClicked(), onItemSingleClicked() and onItemDoubleClicked() will make the design more comprehensible. onItemClicked() will be responsible solely for distinguishing single-clicks from double-clicks and calling the corresponding self-documenting handler.

`on_itemActivated` handles the `QListWidget::itemActivated` event, whereas `on_itemClicked` handles the `QListWidget::itemPressed` event. Further splitting the functions I think will not really provide more clarity since the code for each function will be 1-2 lines. IMO better comments (and maybe variable/function naming) is the option here.

src/readinglistbar.cpp
sgourdas: Thanks for bringing this up because it is something I wanted to discuss. Essentially, the difference is that the singleShot alternative allows the event loop to process other events in the event queue. On the other hand the code that you provided makes the thread block while sleeping. 

I am not sure which is more sufficient in this case, but I initially chose the first one to allow for better execution. LMK if you have any doubts.

src/mainwindow.cpp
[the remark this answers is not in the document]
sgourdas: In that case, that is perfect if we decide that is what we want to do.

src/readinglistbar.cpp
sgourdas: Additionally, after testing this, it seems to me that the application freezes for a small amount of time when double clicking on a reading list bookmark.

src/mainwindow.h
veloman-yunkan: > The reason we cannot be reliant on the checked state of the reading list button is because when entering the library, that is checked to false.

That's exactly what I am proposing to change. The reading list button must be checked/unchecked solely by the user. In other words when the reading list button is checked and the current tab changes to library or settings the reading list button stays checked (but the sidebar is update accordingly). I think that such behavior will be less confusing once the users realize that the reading list is displayed only alongside the content tabs.

veloman-yunkan:
> `on_itemActivated` handles the `QListWidget::itemActivated` event, whereas `on_itemClicked` handles the `QListWidget::itemPressed` event. Further splitting the functions I think will not really provide more clarity since the code for each function will be 1-2 lines. IMO better comments (and maybe variable/function naming) is the option here.

The value of a function is not only in its size but also in its name. A good name for an identifier may eliminate the need for a comment. Whenever possible the code must explain itself. In our case "single-click" and "double-click" are easily understandable concepts in the use-model domain and it will be good if they are mirrored in the design of the code.

src/readinglistbar.cpp
veloman-yunkan: So the answer to my question is 'no' (because of the difference in the blocking behaviour).

Now, supposing that `ReadingListBar::on_itemSingleClicked()` is nevertheless introduced you can rewrite this code in a more comprehensible way as follows:

```
QTimer::singleShot(doubleClickInterval, [=]() {
    const qint64 timePassed = lastDoubleClickTime.msecsTo(thisClickTime);
    if ( timePassed <= doubleClickInterval ) {
        on_itemSingleClicked(item, buttons);
    }
});
```

src/mainwindow.h
sgourdas: In that case, I totally agree. It seems to me also that it makes sense the most.

sgourdas:
@veloman-yunkan ok, just a last comment to clear things up, with naming and amount of functions.

We have 2 events:
- `QListWidget::itemActivated` from which we get **double** click and **enter** key
- `QListWidget::itemPressed` from which we get **single** or **middle** click (and also return if it was a single click of a middle click action)

And we will refactor the logic so they are handled:
- `onItemActivated` handler with `itemDoubleClicked()` and `itemEntered()`
- `onItemPressed` handler with `itemSingleClicked()` and `itemMiddleClicked()`

Which, after adding identification of each type of click in them, operate as the title explains.

src/readinglistbar.cpp
sgourdas: After testing, using the code below, this does not work as intended. 
```
QDateTime thisClickTime = QDateTime::currentDateTime();
int doubleClickInterval = QApplication::doubleClickInterval();
QTimer::singleShot(doubleClickInterval, [=]() {
    const qint64 timePassed = lastDoubleClickTime.msecsTo(thisClickTime);
    if (timePassed <= doubleClickInterval) { return; }
});
QEventLoop().exec();
```

I have left it commented out, in my last commit, if you want to check.

veloman-yunkan:
> @veloman-yunkan ok, just a last comment to clear things up, with naming and amount of functions.
> 
> We have 2 events:
> 
>   * [`QListWidget::itemActivated`](https://doc.qt.io/qt-6/qlistwidget.html#itemActivated) from which we get **double** click and **enter** key
> 
>   * [`QListWidget::itemPressed`](https://doc.qt.io/qt-6/qlistwidget.html#itemPressed) from which we get **single** or **middle** click (and also return if it was a single click of a middle click action)
> 
> 
> And we will refactor the logic so they are handled:
> 
>   * `onItemActivated` handler with `itemDoubleClicked()` and `itemEntered()`
> 
>   * `onItemPressed` handler with `itemSingleClicked()` and `itemMiddleClicked()`
> 
> 
> Which, after adding identification of each type of click in them, operate as the title explains.

I am not sure that separating `itemMiddleClicked()` from `itemSingleClicked()` is justified - after all a middle-click is just a kind of a single-click. But let me look at the code.

src/readinglistbar.cpp
veloman-yunkan: Your code differs from my proposal significantly (the `return` statement has no effect since it returns from the lambda which is about to complete on the next line anyway). However, mine contained a bug too. It should have been as follows (note the corrected condition in the `if` statement):

```c++
int doubleClickInterval = QApplication::doubleClickInterval();
QTimer::singleShot(doubleClickInterval, [=]() {
    const qint64 timePassed = lastDoubleClickTime.msecsTo(thisClickTime);
     if (timePassed > doubleClickInterval) {
         itemSingleClicked(item);
     }
});
```

src/readinglistbar.cpp
```diff
@@ -65,34 +66,54 @@ void ReadingListBar::setupList()
...
 }
 
-void ReadingListBar::on_itemActivated(QListWidgetItem* item) // Handles double click and enter key
+void ReadingListBar::onItemActivated(QListWidgetItem* item, Qt::MouseButtons buttons) // Handles double click and enter key
 {   
-    lastDoubleClickTime = QDateTime::currentDateTime();
-    openUrl(item, true);
+    if (buttons & Qt::LeftButton) { itemDoubleClicked(item); }
+    else { itemEntered(item); }
 }
```
veloman-yunkan, line 73: [`QListWidget::itemActivated()`](https://doc.qt.io/qt-5/qlistwidget.html#itemActivated) is a somewhat confusing signal. Its description reads:

> This signal is emitted when the `item` is activated. The `item` is activated when the user **clicks or double clicks on it, depending on the system configuration**.

Therefore calling `itemDoubleClicked()` from `onItemActiavated()` may be wrong. But do we have to handle any clicks via `onItemActivated()` at all? Why not ignore any item activation events trigerred by mouse clicks and instead take advantage of [`QListWidget::itemClicked()`](https://doc.qt.io/qt-5/qlistwidget.html#itemClicked) and [`QListWidget::itemDoubleClicked`](https://doc.qt.io/qt-5/qlistwidget.html#itemDoubleClicked) signals? I think that `ReadingListBar::onItemPressed()` should be renamed "back to" `ReadingListBar::onItemClicked()` and connected to the ``QListWidget::itemClicked` signal.

src/mainwindow.cpp
```diff
@@ -110,27 +110,17 @@ bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
...
-        libraryPageDisplayed();
-    } else if (readingListRemainActive) { // Leaving the library && before the reading list was active
-        readingList->setChecked(true);
-        readingListRemainActive = false;
-    } else if (!readingList->isChecked()) { // Not in library && readinglist inactive -> hide the side bar
-        mp_ui->sideBar->hide();
-    }
+    if(libraryDisplayed) { libraryPageDisplayed(); } 
+    else if (readingList->isChecked()) { mp_ui->sideBar->setCurrentWidget(mp_ui->readinglistbar); }
 }
```
veloman-yunkan, line 124: Now you can have the sidebar of the library view on content tabs:

Screenshot from 2024-04-26 12-27-20

You should hide the sidebar when switching to a content tab while the reading list button is unchecked.

veloman-yunkan (changes requested):
The set of `ReadingListBar::onItem*()` handlers and their relations with each other still looks somewhat messy to me.

src/readinglistbar.cpp
sgourdas: My bad, I misunderstood that you were intending to do something different. This works perfect.

Also, sorry for the very late feedback. Give me a bit of time to get back to you on the rest as well.

src/readinglistbar.cpp
sgourdas: > It is also activated when the user presses the activation key 

`QListWidget::itemActivated()` is there so as to identify those clicks, but also actions like the `Enter` key for opening an element. 

Using `QListWidget::itemClicked()` and `QListWidget::itemDoubleClicked()` was my first thought as well. After seeing that `QListWidget::itemActivated()` was already there tough, I thought it would be better to just add the `QListWidget::itemPressed` signal so as to have the minimum amount of handlers possible. Is this an issue in your opinion?

src/mainwindow.cpp
sgourdas: Missed this. Fixing on next commit :)

src/readinglistbar.cpp
veloman-yunkan: Amount of handlers is not a metric that we should care about. Software engineers should first of all care about the comprehensibility and maintainability of code. If in a certain situation that translates to the amount of handlers let so be it. But we should not forget how true quality of code should be measured.

sgourdas: Got it. Lets continue based on this.

- `onItemActivated` IMO has to stay to keep the action of the **activation** key, but we are now going to return when it is triggered by a click. 
- **Double clicks** will be triggered with the `QListWidget::itemDoubleClicked` and handled by `ReadingListBar::onItemDoubleClicked`.
- `ReadingListBar::onItemPressed()` needs also to be kept from my POV since it is our way for determining **middle clicks**. We can return if we detect left click.
- If the above is true the current behavior will be taken from `ReadingListBar::onItemPressed()` to use for the `ReadingListBar::onItemClicked()` and connect it to the `QListWidget::itemClicked` signal, for the **single left clicks**.

veloman-yunkan: > * `ReadingListBar::onItemPressed()` needs also to be kept from my POV since it is our way for determining **middle clicks**. We can return if we detect left click.

Why can't you detect middle clicks inside `ReadingListBar::onItemClicked()`?

sgourdas: `QListWidget::itemClicked` does not trigger from middle clicks

veloman-yunkan: The description of [`QListWidget::itemClicked`](https://doc.qt.io/qt-5/qlistwidget.html#itemClicked) differs from that of [`QListWidget::itemPressed`](https://doc.qt.io/qt-5/qlistwidget.html#itemPressed) in exactly one word (clicked → pressed). So if your claim is true, then the documentation is defective.
Whatever, please go on with your implementation and we will continue from there.

src/readinglistbar.cpp
```diff
@@ -13,8 +13,23 @@ ReadingListBar::ReadingListBar(QWidget *parent) :
...
+                onItemDoubleClicked(item, QGuiApplication::mouseButtons());
+            });
+    connect(ui->listWidget, &QListWidget::itemPressed, 
+            this, [this](QListWidgetItem* item) {
+                onItemPressed(item, QGuiApplication::mouseButtons());
+            });
+    connect(ui->listWidget, &QListWidget::itemActivated, 
+            this, [this](QListWidgetItem* item) {
+                onItemActivated(item, QGuiApplication::mouseButtons());
+            });
```
veloman-yunkan, line 31: When you remove the `MouseButtons` parameter for some of the `onItem*()` handlers, please simplify the `connect()` expression.

src/readinglistbar.cpp
```diff
@@ -60,13 +74,45 @@ void ReadingListBar::setupList()
...
+void ReadingListBar::onItemClicked(QListWidgetItem* item, Qt::MouseButtons) // Receives single clicks
+{
+    QDateTime thisClickTime = QDateTime::currentDateTime();
+    int doubleClickInterval = QApplication::doubleClickInterval();
+    QTimer::singleShot(doubleClickInterval, [=]() { // Give time for a double click be handled
+        const qint64 timePassed = lastDoubleClickTime.msecsTo(thisClickTime);
+        if (timePassed > doubleClickInterval) {
+            openUrl(item, false);
+        }
+    });
```
veloman-yunkan, line 86: Now that detection of a double click is performed by Qt, you can replace `lastDoubleClickTime` with simpler logic (e.g. `int clickKind`, set it to 1 in `onItemClicked()` and to 2  in `onItemDoubleClicked()`).

veloman-yunkan (changes requested):

src/readinglistbar.cpp
```diff
@@ -60,13 +74,45 @@ void ReadingListBar::setupList()
...
+        openUrl(item, true);
+    }
+}
+
+void ReadingListBar::onItemActivated(QListWidgetItem* item, Qt::MouseButtons buttons) // Receives double click and activation key
+{   
+    if (buttons & ~Qt::LeftButton) { 
+        openUrl(item, true);
+    }
+}
```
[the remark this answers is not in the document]
sgourdas: Yeah actually no **mouse** button is pressed at all so this makes sense... Fixed it on next iteration

src/readinglistbar.cpp
sgourdas: Good idea!

src/mainwindow.cpp
```diff
@@ -111,17 +110,25 @@ bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
...
+{
+    QAction *readingList = KiwixApp::instance()->getAction(KiwixApp::ToggleReadingListAction);
+    if(libraryDisplayed) { 
+        libraryPageDisplayed(); 
+    } else if (readingList->isChecked()) { 
+        mp_ui->sideBar->setCurrentWidget(mp_ui->readinglistbar); 
+    } else {
+        mp_ui->sideBar->hide();
+    }
+}
```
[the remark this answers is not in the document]
sgourdas: Changed accordingly to not allow the reading list to be visible on the settings tab.

src/readinglistbar.cpp
```diff
@@ -60,13 +70,44 @@ void ReadingListBar::setupList()
...
+
+void ReadingListBar::onItemActivated(QListWidgetItem* item, Qt::MouseButtons buttons) // Receives double click and activation key
+{
+    if (buttons & Qt::LeftButton) { 
+        return;
+    }
+    openUrl(item, true);
```
veloman-yunkan, line 101: Change to

```c++
    if ( !buttons ) { // clicks are handled elsewhere, handle only the activation key case 
        openUrl(item, true);
    }
```

src/readinglistbar.cpp
```diff
@@ -60,13 +70,44 @@ void ReadingListBar::setupList()
     }
 }
 
-void ReadingListBar::on_itemActivated(QListWidgetItem* item)
+void ReadingListBar::onItemClicked(QListWidgetItem* item) // Receives single clicks
```
veloman-yunkan, line 73: Move all end-of-line comments like this one above the function declaration

src/readinglistbar.cpp
```diff
@@ -60,13 +70,44 @@ void ReadingListBar::setupList()
...
+}
+
+void ReadingListBar::onItemPressed(QListWidgetItem* item, Qt::MouseButtons buttons) // Receives single and middle click
+{
+    if (buttons & Qt::MiddleButton) {
+        openUrl(item, true);
+    }
+}
+
+void ReadingListBar::onItemActivated(QListWidgetItem* item, Qt::MouseButtons buttons) // Receives double click and activation key
```
veloman-yunkan, line 96: Fix the comment (activation via mouse is platform dependent and can be either a single click or a double click)

veloman-yunkan (changes requested):

src/readinglistbar.h
```diff
@@ -18,9 +19,15 @@ class ReadingListBar : public QWidget
...
     void setupList();
-    void on_itemActivated(QListWidgetItem *item);
+    void onItemClicked(QListWidgetItem* item);
+    void onItemDoubleClicked(QListWidgetItem *item);
+    void onItemPressed(QListWidgetItem* item, Qt::MouseButtons buttons);
+    void onItemActivated(QListWidgetItem *item, Qt::MouseButtons buttons);
 private:
     Ui::readinglistbar *ui;
+    QDateTime lastDoubleClickTime = QDateTime(QDate(1970, 1, 1), QTime(0, 0, 0));
+    int clickKind;
```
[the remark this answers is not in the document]
sgourdas: Forgot to commit this part :)

src/readinglistbar.cpp
```diff
@@ -59,14 +69,44 @@ void ReadingListBar::setupList()
...
+// Receives single and middle click
+void ReadingListBar::onItemPressed(QListWidgetItem* item, Qt::MouseButtons buttons)
+{
+    if (buttons & Qt::MiddleButton) {
+        openUrl(item, true);
+    }
+}
+// Receives left clicks and activation key
+void ReadingListBar::onItemActivated(QListWidgetItem* item, Qt::MouseButtons buttons)
+{
```
veloman-yunkan, line 97: Add empty lines after the closing braces of function bodies.

src/mainwindow.cpp
```diff
@@ -120,27 +119,25 @@ bool MainWindow::eventFilter(QObject* /*object*/, QEvent* event)
...
 {
-    auto app = KiwixApp::instance();
-
-    // When library sidebar appeared, or hidden, reading list is always hidden.
-    app->getAction(KiwixApp::ToggleReadingListAction)->setChecked(false);
-
-    if (showed) {
-        mp_ui->sideBar->setCurrentWidget(mp_ui->contentmanagerside);
+    if (state) {
+        mp_ui->sideBar->setCurrentWidget(mp_ui->readinglistbar);
```
veloman-yunkan, line 140: I think that the diff will be easier to comprehend if you swap the order of these two functions.

veloman-yunkan (changes requested):
Almost there. A few final touches and we will merge.

src/readinglistbar.cpp
sgourdas: Yeah I was wondering on that

veloman-yunkan (approved):
Thanks for bearing with me! I don't remember well the original shape of this PR, but I think that you should agree that it got significantly simpler and nicer.

PR #1081, opened 2024-04-11
Handling of exceptions from kiwix::Downloader::close()

### LINKED ISSUE #694: Unhandled `kiwix::AriaError` exception if KIWIX_DATA_DIR is set to a unexisting directory.
`KIWIX_DATA_DIR` env variable can be used to force a specific directory to put all configuration/zim somewhere (instead of the default user data dir).
At exist, we tell aria2 to save the session in the data directory and if it not exists, a exception is raised.

We must catch it and handle it correctly.

veloman-yunkan:
Fixes #694

If saving the downloader state during exit fails<sup>+</sup> then, instead of crashing, the following error is reported on `stderr`:

> ERROR: Failed to save the downloader state: Failed to serialize session to '/nonexistingdir/kiwix.session'.

---
<sup>+</sup>It can happen, for example, if `KIWIX_DATA_DIR` is set to a non-existing directory which in addition cannot be created due to permissions.

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -125,7 +125,12 @@ KiwixApp::~KiwixApp()
 {
     m_server.stop();
     if (mp_downloader) {
-        mp_downloader->close();
+        try {
+            mp_downloader->close();
+        } catch (const std::exception& err) {
+            std::cerr << "ERROR: Failed to save the downloader state: "
+                      << err.what() << std::endl;
+        }
         delete mp_downloader;
     }
     if (mp_manager) {


mgautierfr (approved):

kelson42:
I have many remarks:
* We should habe a section about this ENv variable and how it works in the README
* I see no reason why we have such an ENV variable, why not kust removing it?

veloman-yunkan:
@kelson42 Your remarks are valid but have nothing to do with this PR. `kiwix::Downloader::close()` can throw for a different reason (e.g. the directory `~/.local/share/kiwix` being deleted while `kiwix-desktop` is running or the aria2 process dying).

kelson42:
@veloman-yunkan OK, will open a dedicated ticket.

PR #1085, opened 2024-04-12
Remove -Werror.

mgautierfr:
`-Werror` make compilation fails on Windows.

The exact warning making the compilation failing is not clear.
See (https://ci.appveyor.com/project/Kiwix/kiwix-build/builds/49600994 for example).

Compilation without `-Werror` is ok (https://ci.appveyor.com/project/Kiwix/kiwix-build/builds/49601697)

Fix kiwix/kiwix-build#690

### DIFF
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -18,8 +18,12 @@ greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
 TARGET = kiwix-desktop
 TEMPLATE = app
 
-QMAKE_CXXFLAGS += -std=c++17 -Werror
+QMAKE_CXXFLAGS += -std=c++17
 QMAKE_LFLAGS +=  -std=c++17
+
+!win32 {
+    QMAKE_CXXFLAGS += -Werror
+}
 
 # Also change resources/org.kiwix.desktop.appdata.xml
 DEFINES += VERSION="2.3.1"


kelson42:
@veloman-yunkan @mgautierfr Is that a good idea to relax here? Actually I have asking yesterday in an other ticket to implement `-Wall`?!

mgautierfr:
> Is that a good idea to relax here?

In absolute terms, not. But without at least an error message telling us what's wrong it would be difficult to fix the warning. In the meanwhile kiwix-desktop is not compiling on Windows and we cannot publish nightly or release.

Maybe good to keep a issue open about that but for now I think the priority is to keep kiwix-desktop building on Windows.

kiwix-desktop.pro
```diff
@@ -18,7 +18,7 @@ greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
 TARGET = kiwix-desktop
 TEMPLATE = app
 
-QMAKE_CXXFLAGS += -std=c++17 -Werror
+QMAKE_CXXFLAGS += -std=c++17
```
veloman-yunkan, line 21: Can't we disable `-Werror` only under Windows?

kelson42: We have a general problem that Windows compiler options are not the same as gcc. We have a similar situation with the cpp17 option.

kelson42:
> Maybe good to keep a issue open about that but for now I think the priority is to keep kiwix-desktop building on Windows.

I have open https://github.com/kiwix/kiwix-desktop/issues/1087

kiwix-desktop.pro
sgourdas: @kelson42 it is not clear to me what the issue is. Can we not do something like the below?
```
!win32 {
    QMAKE_CXXFLAGS += -Werror
}
```

mgautierfr: Done.

Successful build with last change is https://ci.appveyor.com/project/Kiwix/kiwix-build/builds/49615228

sgourdas (approved):
LGTM

veloman-yunkan (approved):

PR #1092, opened 2024-04-16
Detection and reporting of issues with the download directory

### LINKED ISSUE #959: Give more information about the "Not enough space storage" message.
Having more information like:
- Needed space,
- Available space,
- Destination directory

would be useful.

veloman-yunkan:
Fixes #959 (and a couple of other bugs)


src/readinglistbar.cpp
```diff
@@ -39,6 +39,9 @@ void ReadingListBar::setupList()
         } catch (std::out_of_range& e) {
             continue;
         }
+        if ( !archive ) {
+            continue;
+        }
```
mgautierfr, line 44: See https://github.com/kiwix/kiwix-desktop/issues/893

mgautierfr (approved):
LGTM.

About the issue #893, your small fix is enough (in the context of this PR only) but it please update the issue about this small code fixed.

PR #1101, opened 2024-05-04
Fixed detection of clicks on download control buttons

### LINKED ISSUE #1099: Wrong detection of click on cancel/resume download button.
See the video [Wrong_click_cancel_resume.webm](https://github.com/kiwix/kiwix-desktop/assets/86161/be1eebdf-2b0b-4893-ac71-9e38d3650f98)

It appears (from code source) that detection of if we click on cancel or resume is based on if we click on right or left side of the "parent widget". We should better compute where we are clicking.

veloman-yunkan:
Fixes #1099

Note that this PR is based on #1098. Though they are not conceptually coupled in any way, they touch the same lines of code and it will be less hassle if #1098 is merged first.

### DIFF
--- a/src/contentmanagerdelegate.cpp
+++ b/src/contentmanagerdelegate.cpp
@@ -117,6 +117,8 @@ void showDownloadProgress(QPainter *painter, QRect box, const DownloadState& dow
     int arcW = w - 90;
     int arcH = h - 40;
 
+    QRect pauseResumeButtonRect(arcX, arcY, arcW, arcH);
+
     double progress  = (double) (downloadInfo.progress) / 100;
     progress = -progress;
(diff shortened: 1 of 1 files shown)


mgautierfr (approved):
Looks good to me.
Need merge of #1098 first. 

Using the bounding rectangle is ok for me. So when you're rebasing, you can rewrite the last commit and remove the suggestion to fix that.

kelson42:
@veloman-yunkan Unfortunately we have a few conflicts and I can not rebase straight.

veloman-yunkan:
> Using the bounding rectangle is ok for me. So when you're rebasing, you can rewrite the last commit and remove the suggestion to fix that.

Done

PR #1110, opened 2024-05-14
Added ContextMenu Button for Previewing Online Books

### LINKED ISSUE #1100: Support for previewing online books
Currently in `kiwix-desktop` one cannot preview a book from the online library directly in `kiwix-desktop`. They have to visit https://library.kiwix.org using their web-browser and find the book of interest. Having the ability to preview books with a single click will improve the user experience.

ShaopengLin:
Fix #1100 

Changes:
- Added context menu button for book preview.
- Button opens the book's URL on "https://library.kiwix.org/viewer#".
- Button is available for book states Online, Downloading, and Download Paused.

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -642,6 +642,14 @@ const kiwix::Book& ContentManager::getRemoteOrLocalBook(const QString &id)
     }
 }
 
+QString ContentManager::getRemoteLibraryUrl() const
+{
+    auto host = m_remoteLibraryManager.getCatalogHost();
+    auto port = m_remoteLibraryManager.getCatalogPort();
+    return port == 443 ? "https://" + host
+                        : "http://" + host + ":" + QString::number(port);
+}
+
 std::string ContentManager::startDownload(const kiwix::Book& book)
 {
     auto downloadPath = getSettingsManager()->getDownloadDir();
@@ -887,27 +895,12 @@ void ContentManager::updateLibrary() {
     } catch (std::runtime_error&) {}
 }
 
-namespace
-{
-
-QString makeHttpUrl(QString host, int port)
-{
-    return port == 443
-         ? "https://" + host
-         : "http://" + host + ":" + QString::number(port);
-}
-
-} // unnamed namespace
-
 void ContentManager::updateRemoteLibrary(const QString& content) {
     (void) QtConcurrent::run([=]() {
         QMutexLocker locker(&remoteLibraryLocker);
         mp_remoteLibrary = kiwix::Library::create();
         kiwix::Manager manager(mp_remoteLibrary);
-        const auto catalogHost = m_remoteLibraryManager.getCatalogHost();
-        const auto catalogPort = m_remoteLibraryManager.getCatalogPort();
-        const auto catalogUrl = makeHttpUrl(catalogHost, catalogPort);
-        manager.readOpds(content.toStdString(), catalogUrl.toStdString());
+        manager.readOpds(content.toStdString(), getRemoteLibraryUrl().toStdString());
         emit(this->booksChanged());
         emit(this->pendingRequest(false));
     });
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -217,17 +217,20 @@ void ContentManager::onCustomContextMenu(const QPoint &point)
     QAction menuResumeBook(gt("resume-download"), this);
     QAction menuCancelBook(gt("cancel-download"), this);
     QAction menuOpenFolder(gt("open-folder"), this);
+    QAction menuPreviewBook(gt("preview-book-in-web-browser"), this);
 
     const auto bookState = getBookState(id);
     switch ( bookState ) {
     case BookState::DOWNLOAD_PAUSED:
         contextMenu.addAction(&menuResumeBook);
         contextMenu.addAction(&menuCancelBook);
+        contextMenu.addAction(&menuPreviewBook);
         break;
 
     case BookState::DOWNLOADING:
         contextMenu.addAction(&menuPauseBook);
         contextMenu.addAction(&menuCancelBook);
+        contextMenu.addAction(&menuPreviewBook);
         break;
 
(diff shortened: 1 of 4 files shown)


src/contentmanager.cpp
```diff
@@ -501,6 +508,19 @@ void ContentManager::openBook(const QString &id)
...
+void ContentManager::openBookPreview(const QString &id)
+{
+    const auto& book = getRemoteOrLocalBook(id);
+
+    std::string pathStr = book.getUrl();
+
+    auto pathStartIndex = pathStr.find_last_of('/') + 1;
+    pathStr = "https://library.kiwix.org/viewer#" +
+              pathStr.substr(pathStartIndex,
+                             pathStr.find(".zim", pathStartIndex) - pathStartIndex);
```
veloman-yunkan, line 520: The preview URL has the form `BASEURL/viewer#BOOKNAME` where `BOOKNAME` can be obtained via `kiwix::Book::getName()` (instead of parsing/deducing it from the result of `kiwix::Book::getUrl()`).

`BASEURL` can be [overriden while debugging](https://github.com/kiwix/kiwix-desktop/commit/c10026a3a2616245f5ea47dedf3be436f5e07bc1). Please generate `BASEURL` as [this](https://github.com/kiwix/kiwix-desktop/commit/535675afdb0fce40121ccafa06909b61ad057895#diff-b143bce41389bc342b560b90019ea25d3a82243fab36352befb2be7401cc2b64R682-R701). In order to avoid introducing code duplication, please add a private helper function `ContentManager::getRemoteLibraryUrl()`.

veloman-yunkan: My previous comment regarding the `BOOKNAME` part of the preview URL was wrong. Actually [ZIM name](https://kiwix-tools.readthedocs.io/en/latest/kiwix-serve.html#term-ZIM-name) is used (see https://kiwix-tools.readthedocs.io/en/latest/kiwix-serve.html#viewer). Note that ZIM name may be different from the ZIM filename (without the extension), however I think that ZIM filenames for content provided by Kiwix are safe to be used as ZIM names. So your code for deducing the ZIM name from the download URL should work. But please compute it in a separate expression:

```c++
const auto zimname = ...;
const auto previewUrl = getRemoteLibraryUrl() + "/viewer#" + zimname;
```

src/contentmanager.cpp
```diff
@@ -501,6 +508,19 @@ void ContentManager::openBook(const QString &id)
     }
 }
 
+void ContentManager::openBookPreview(const QString &id)
+{
+    const auto& book = getRemoteOrLocalBook(id);
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/contentmanager.cpp
ShaopengLin: Done

src/contentmanager.h
```diff
@@ -109,6 +110,7 @@ public slots:
 
 private: // functions
     QStringList getBookIds();
+    QString getRemoteLibraryUrl();
```
veloman-yunkan, line 113: Make `const`

src/contentmanager.cpp
```diff
@@ -975,6 +1001,11 @@ QStringList ContentManager::getBookIds()
     }
 }
 
+QString ContentManager::getRemoteLibraryUrl() 
+{
+    return "https://library.kiwix.org";
+}
```
veloman-yunkan, line 1007: Implement like [requested](https://github.com/kiwix/kiwix-desktop/pull/1110#discussion_r1601058698) in my previous review:

> `BASEURL` can be [overriden while debugging](https://github.com/kiwix/kiwix-desktop/commit/c10026a3a2616245f5ea47dedf3be436f5e07bc1). Please generate `BASEURL` as [this](https://github.com/kiwix/kiwix-desktop/commit/535675afdb0fce40121ccafa06909b61ad057895#diff-b143bce41389bc342b560b90019ea25d3a82243fab36352befb2be7401cc2b64R682-R701). In order to avoid introducing code duplication, please add a private helper function `ContentManager::getRemoteLibraryUrl()`.

veloman-yunkan (changes requested):

src/contentmanager.cpp
veloman-yunkan: And then use `getRemoteLibraryUrl()` in `ContentManager::updateRemoteLibrary()` too

src/contentmanager.h
ShaopengLin: Done

src/contentmanager.cpp
ShaopengLin: Done

src/contentmanager.cpp
```diff
@@ -975,6 +998,12 @@ QStringList ContentManager::getBookIds()
     }
 }
 
+QString ContentManager::getRemoteLibraryUrl() const
+{
+  return makeHttpUrl(m_remoteLibraryManager.getCatalogHost(),
+                     m_remoteLibraryManager.getCatalogPort());
+}
+
```
veloman-yunkan, line 1006: Choose a better place for this definition

veloman-yunkan (changes requested):
Good.

Please split the PR into two commits:

1. Extracting a piece of inline code into a new function `ContentManager::getRemoteLibraryUrl()` so that you can take advantage of it while implementing the desired functionality.
2. Implementing the functionality itself.

As a general rule - we care not only about the final code but also about the path of arriving at it. As a series of small commits the change is easier to comprehend. Also it allows to distinguish changes that affect user-observable behaviour from refactoring changes where only the organization of the code is concerned.

src/contentmanager.cpp
ShaopengLin: I think at the moment, makeHttpUrl should just be replaced by getRemoteLibraryUrl since it is only defined for the file scope and does the same thing anyway.

veloman-yunkan: Makes sense

veloman-yunkan (approved):

PR #1115, opened 2024-05-22
Fixed Whiteouts of Book Descriptions on Hover

### LINKED ISSUE #1112: Book description misbehaves on hover
[Screencast from 05-17-2024 02:16:50 PM.webm](https://github.com/kiwix/kiwix-desktop/assets/11933502/5b1811e3-59cb-4322-bbc9-44268253a8af)

**Update 2024-05-22**: As first noted in #1115, another bug is that clicking the description row within the range of the open/download action column results in a crash.

ShaopengLin:
Fix #1112

Changes:
- For the description row, only use one column to span across.
- DescriptionNode's data is now available on column index 0 instead of 1.
- Added word wrap to previously unwrapped text
- Prevented button actions from the last column to overflow into the description row.

### DIFF
--- a/src/contentmanagerdelegate.cpp
+++ b/src/contentmanagerdelegate.cpp
@@ -194,7 +194,7 @@ void ContentManagerDelegate::paintBookState(QPainter *p, const QStyleOptionViewI
 void ContentManagerDelegate::paint(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const
 {
     QRect r = option.rect;
-    if (index.parent().isValid()) {
+    if (isDescriptionIndex(index)) {
         // additional info
         QRect nRect = r;
         auto viewWidth = KiwixApp::instance()->getContentManager()->getView()->getView()->width();
@@ -293,7 +293,7 @@ QSize ContentManagerDelegate::sizeHint(const QStyleOptionViewItem &option, const
 {
     Q_UNUSED(option);
 
-    if (index.parent().isValid()) {
+    if (isDescriptionIndex(index)) {
         return QSize(300, 70);
     }
     return QSize(50, 70);
--- a/src/contentmanagerdelegate.cpp
+++ b/src/contentmanagerdelegate.cpp
@@ -285,10 +285,22 @@ void ContentManagerDelegate::handleLastColumnClicked(const QModelIndex& index, Q
 
 QSize ContentManagerDelegate::sizeHint(const QStyleOptionViewItem &option, const QModelIndex &index) const
 {
-    Q_UNUSED(option);
+    if (isDescriptionIndex(index))
+    {
+        const auto treeView = KiwixApp::instance()->getContentManager()->getView()->getView();
 
-    if (isDescriptionIndex(index)) {
-        return QSize(300, 70);
+        const int width = treeView->header()->length() - 2*treeView->indentation();
+        // XXX: see QTreeView::padding in resources/css/_contentManager.css
+        const int verticalPadding = 4;
+        const int horizontalPadding = 4;
+        QRect descRect(0, 0, width - 2 * horizontalPadding, 0);
+
+        /* Based on the rectangle and text, find the best fitting size. */
+        QFontMetrics fm(option.font);
+        const QString text = index.data().toString();
+        const auto format = Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap;
+        const int textHeight = fm.boundingRect(descRect, format, text).height();
+        return QSize(width, std::max(textHeight + verticalPadding, 70));
     }
     return QSize(50, 70);
 }
--- a/src/contentmanagerdelegate.cpp
+++ b/src/contentmanagerdelegate.cpp
@@ -193,15 +193,9 @@ void ContentManagerDelegate::paintBookState(QPainter *p, const QStyleOptionViewI
 
 void ContentManagerDelegate::paint(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const
 {
-    QRect r = option.rect;
-    if (isDescriptionIndex(index)) {
-        // additional info
-        QRect nRect = r;
-        auto viewWidth = KiwixApp::instance()->getContentManager()->getView()->getView()->width();
-        nRect.setWidth(viewWidth);
-        painter->drawText(nRect, Qt::AlignLeft | Qt::AlignVCenter, index.data(Qt::UserRole+1).toString());
-        return;
-    }
+    if (isDescriptionIndex(index))
+        return QStyledItemDelegate::paint(painter, option, index);
+
     QStyleOptionViewItem eOpt = option;
     if (index.column() == 1) {
         auto bFont = painter->font();
--- a/src/contentmanagermodel.cpp
+++ b/src/contentmanagermodel.cpp
@@ -60,10 +60,13 @@ QVariant ContentManagerModel::data(const QModelIndex& index, int role) const
         return QVariant();
 
     const auto col = index.column();
-    const bool isThumbnailRequest = col == 0 && role == Qt::DecorationRole;
-    const bool otherDataRequest = col != 0 && (role == Qt::DisplayRole || role == Qt::UserRole+1 );
+    const bool isThumbnailRequest =
+        !isDescriptionIndex(index) && col == 0 && role == Qt::DecorationRole;
+    const bool isDescriptionRequest =
+        isDescriptionIndex(index) && col == 0 && role == Qt::DisplayRole;
+    const bool otherDataRequest = col != 0 && role == Qt::DisplayRole;
 
-    if ( !isThumbnailRequest && !otherDataRequest )
+    if ( !isThumbnailRequest && !otherDataRequest && !isDescriptionRequest )
         return QVariant();
 
     const auto item = static_cast<Node*>(index.internalPointer());
--- a/src/contentmanagermodel.cpp
+++ b/src/contentmanagermodel.cpp
@@ -86,7 +86,7 @@ QVariant ContentManagerModel::data(const QModelIndex& index, int role) const
 Qt::ItemFlags ContentManagerModel::flags(const QModelIndex &index) const
 {
     Qt::ItemFlags defaultFlags = QAbstractItemModel::flags(index);
-    if (index.isValid() && index.parent().isValid()) {
+    if (isDescriptionIndex(index)) {
         return defaultFlags & ~Qt::ItemIsDropEnabled & ~Qt::ItemIsDragEnabled & ~Qt::ItemIsSelectable & ~Qt::ItemIsEditable & ~Qt::ItemIsUserCheckable;
     }
     return defaultFlags;
--- a/src/contentmanagerview.cpp
+++ b/src/contentmanagerview.cpp
@@ -21,6 +21,7 @@ ContentManagerView::ContentManagerView(QWidget *parent)
 
     connect(mp_ui->m_view, &QTreeView::clicked, this, &ContentManagerView::onClicked);
     connect(mp_ui->m_view, &QTreeView::expanded, this, &ContentManagerView::onExpanded);
+    connect(this, &ContentManagerView::sizeHintChanged, managerDelegate, &QStyledItemDelegate::sizeHintChanged);
 }
 
 ContentManagerView::~ContentManagerView()
@@ -55,4 +56,22 @@ void ContentManagerView::onExpanded(QModelIndex index)
 {
     if (!mp_ui->m_view->isFirstColumnSpanned(0, index))
         mp_ui->m_view->setFirstColumnSpanned(0, index, true);
+}
+
+/**
+ * @brief Notify delegate to update size hint of the visible description rows.
+ */
+void ContentManagerView::updateSizeHint()
+{
+    auto view = this->getView();
+    if (!view->isVisible())
+        return;
+
+    auto visibleIndex = view->indexAt(view->rect().topLeft());
+    while (visibleIndex.isValid())
+    {
+        if (isDescriptionIndex(visibleIndex))
+            emit sizeHintChanged(visibleIndex);
+        visibleIndex = view->indexBelow(visibleIndex);
+    }
 }
--- a/src/contentmanagerview.cpp
+++ b/src/contentmanagerview.cpp
@@ -20,6 +20,7 @@ ContentManagerView::ContentManagerView(QWidget *parent)
     mp_ui->stackedWidget->setCurrentIndex(0);
 
     connect(mp_ui->m_view, &QTreeView::clicked, this, &ContentManagerView::onClicked);
+    connect(mp_ui->m_view, &QTreeView::expanded, this, &ContentManagerView::onExpanded);
 }
 
 ContentManagerView::~ContentManagerView()
@@ -48,4 +49,10 @@ void ContentManagerView::onClicked(QModelIndex index)
     } else {
         mp_ui->m_view->expand(zeroColIndex);
     }
+}
+
+void ContentManagerView::onExpanded(QModelIndex index)
+{
+    if (!mp_ui->m_view->isFirstColumnSpanned(0, index))
+        mp_ui->m_view->setFirstColumnSpanned(0, index, true);
 }
--- a/src/contentmanagerview.cpp
+++ b/src/contentmanagerview.cpp
@@ -19,17 +19,7 @@ ContentManagerView::ContentManagerView(QWidget *parent)
     loader = new KiwixLoader(mp_ui->loading);
     mp_ui->stackedWidget->setCurrentIndex(0);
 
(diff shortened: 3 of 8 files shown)


src/contentmanagerdelegate.cpp
```diff
@@ -195,9 +195,17 @@ void ContentManagerDelegate::paint(QPainter *painter, const QStyleOptionViewItem
...
+
+        /* Align text to the x coordinate of column index 1, the Name column. */
+        nRect.setX(view->visualRect(index.parent().siblingAtColumn(1)).x());
+        nRect.setWidth(view->width());
+
+        /* Enable word wrap here. QTreeView word wrap doesn't apply on drawText. */
+        QTextOption op;
+        op.setAlignment(Qt::AlignLeft | Qt::AlignVCenter);
+        op.setWrapMode(QTextOption::WordWrap);
+        painter->drawText(nRect, index.data(Qt::UserRole+1).toString(), op);
```
veloman-yunkan, line 208: Extract this piece of code into a new method `ContextManagerDelegate::paintBookDescription()`

src/contentmanagerview.cpp
```diff
@@ -23,6 +23,9 @@ ContentManagerView::ContentManagerView(QWidget *parent)
         if (index.column() == (mp_ui->m_view->model()->columnCount() - 1))
             return;
 
+        /* The descriptions will only need one column to display. */
+        mp_ui->m_view->setFirstColumnSpanned(0, index, true);
+
```
veloman-yunkan, line 28: Please convert the handler lambda containing this change into a named slot of `ContentManagerView` (do that in a separate commit). Then modify it in a separate commit.

veloman-yunkan (changes requested):
Whiteouts of book description have not been fixed - on the contrary, the name column was included in that undesirable behaviour.

On another note, this PR fixes several issues. Please extract each fix in a commit of its own.

ShaopengLin:
> Whiteouts of book description have not been fixed - on the contrary, the name column was included in that undesirable behaviour.

@veloman-yunkan I can't seem to reproduce what you mean in both 5.15.2 and 6.2.4.

kelson42:
@ShaopengLin In the future, please develop your PR in a feature branch in our repo (not in your fork).

ShaopengLin:
I will try installing 5.12 a bit later and see if I can reproduce it. Will need to download it eventually since it seems compilation issues will continue to present themselves in the future (e.g. #1116) :).

ShaopengLin:
Cannot seem to reproduce it either in 5.12. Will just wait for a response.

veloman-yunkan:
> Whiteouts of book description have not been fixed - on the contrary, the name column was included in that undesirable behaviour.

I double checked that it was not a result of an incorrect build:

[Screencast from 05-29-2024 05:44:02 PM.webm](https://github.com/kiwix/kiwix-desktop/assets/11933502/4178f4b9-ff6b-4c58-bce8-0a987d6fe1bf)

I am using Ubuntu 22.04.

kelson42:
@ShaopengLin I have been able to reproduce this problem (with both Qt6 and Qt6) but I hard a had time. It's important to:
* Deal with the very same ZIM files like in the screencast
* Windows size should be the very same
* After downloading, go once to "online" and then back "to offline"
* Finally uncollapse and the bug should be there

ShaopengLin:
Confirmed fixed in all Qt5 and Qt6. The reason it isn't working is the press on the specific drop-down button will not count as a clicked event, thus our handler for the fix isn't called.

src/contentmanagerdelegate.cpp
```diff
@@ -164,6 +164,17 @@ void ContentManagerDelegate::paintButton(QPainter *p, const QRect &r, QString t)
...
+{
+    QRect nRect = r;
+    auto view = KiwixApp::instance()->getContentManager()->getView()->getView();
+
+    /* Align text to the x coordinate of column index 1, the Name column. */
+    nRect.setX(view->visualRect(index.parent().siblingAtColumn(1)).x());
+    nRect.setWidth(view->width());
+    p->drawText(nRect, Qt::AlignLeft | Qt::AlignVCenter, index.data(Qt::UserRole+1).toString());
+}
+
```
veloman-yunkan: This function was extracted as a result of refactoring but also contains a small change (`nRect.setX()`). Better do that in separate commits. Definitely, introducing `ContentManagerDelegate::paintBookDescription()` as a refactoring step doesn't fix the bug. Does `nRect.setX()` help to do it? Please split this commit into two or three so that the bug fix contains only contains changes that actually fix the bug.

src/contentmanagerdelegate.cpp
```diff
@@ -253,7 +253,8 @@ bool ContentManagerDelegate::editorEvent(QEvent *event, QAbstractItemModel *mode
         const auto lastColumnClicked = ((index.column() == 5) && (clickX > x && clickX < x + w)
                                                         && (clickY > y && clickY < y + h));
 
-        if (lastColumnClicked)
+        /* The description row does not constitute as part of the open button. */
+        if (lastColumnClicked && !index.parent().isValid())
             handleLastColumnClicked(index, e, option);
```
veloman-yunkan, line 258: I [fixed this same problem slightly differently](https://github.com/kiwix/kiwix-desktop/pull/1118/commits/aae0fd5de2464f5d607141f2093f2de1e20cc819) in my so far WIP PR #1118. Do you think that an early return may have undesirable effects?

veloman-yunkan (changes requested):

src/contentmanagerdelegate.cpp
```diff
@@ -172,7 +174,7 @@ void ContentManagerDelegate::paintBookDescription(QPainter *p, const QRect &r, c
     /* Align text to the x coordinate of column index 1, the Name column. */
     nRect.setX(view->visualRect(index.parent().siblingAtColumn(1)).x());
     nRect.setWidth(view->width());
-    p->drawText(nRect, Qt::AlignLeft | Qt::AlignVCenter, index.data(Qt::UserRole+1).toString());
+    p->drawText(nRect, index.data(Qt::UserRole+1).toString(), descTextOp);
```
veloman-yunkan: Long text isn't wrapped nicely:

Screenshot from 2024-06-03 13-49-41

At the least, I think that the width of the box where the text is drawn should be reduced.

But a much better solution will be if you can completely eliminate the need for custom drawing of the description via `ContentManagerDelegate` and fully rely on Qt's standard handling of tree views. Then, hopefully, the height of the description line (which is now fixed) will automatically adjust to the amount of content in it.

src/contentmanagerdelegate.cpp
ShaopengLin: Yeah, it doesn't do the fix on its own. My mistake.

src/contentmanagerdelegate.cpp
ShaopengLin: An early return would indeed fix this. I will remove my line then.

ShaopengLin: @veloman-yunkan I believe fixing this your commit makes more sense. Your can do that check anytime after the middle button click check.

src/contentmanagerdelegate.cpp
ShaopengLin: Resolved. In the new version, we don't need this refactor anymore as it is removed.

src/contentmanagerdelegate.cpp
ShaopengLin: Done

src/contentmanagerdelegate.cpp
```diff
@@ -285,10 +285,26 @@ void ContentManagerDelegate::handleLastColumnClicked(const QModelIndex& index, Q
...
+        QSize textRectSize = fm
+            .boundingRect(r,
+                          Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap,
+                          index.data().toString())
+            .size();
+        
+        if (textRectSize.height() < 70)
+            textRectSize.setHeight(70);
+        return textRectSize;
     }
```
veloman-yunkan, line 308: Nice, but it seems to slightly fall short of its goal:

Screenshot from 2024-06-06 13-49-50

src/contentmanagermodel.cpp
```diff
@@ -61,14 +61,17 @@ QVariant ContentManagerModel::data(const QModelIndex& index, int role) const
 
     const auto col = index.column();
     const bool isThumbnailRequest = col == 0 && role == Qt::DecorationRole;
-    const bool otherDataRequest = col != 0 && (role == Qt::DisplayRole || role == Qt::UserRole+1 );
+    const bool isDescriptionRequest =
+        col == 0 && (role == Qt::DecorationRole || role == Qt::DisplayRole) &&
+        index.parent().isValid();
```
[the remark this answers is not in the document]
ShaopengLin: 1. Where do you think is the best place to put isDescriptionLine(index)? I personally think descriptionnode.h is the best place as placing a view, model, or the contentmanager class into delegate somewhat doesn't make sense given a MVC architecture.
2. It was previously needed because the isThumbNail overlaps this. Now that I think about it I can use isDescriptionLine to distinguish between them instead of role.
```
const bool isThumbnailRequest = col == 0 && role == Qt::DecorationRole;
...
if ( !isThumbnailRequest && !otherDataRequest && !isDescriptionRequest)
        return QVariant();
```

src/contentmanagerdelegate.cpp
```diff
@@ -285,10 +285,27 @@ void ContentManagerDelegate::handleLastColumnClicked(const QModelIndex& index, Q
...
+        QFontMetrics fm(option.font);
+        const QString text = index.data().toString();
+        const auto format = Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap;
+        const QRect textRect = fm.boundingRect(descRect, format, text);
+        QSize textSize = textRect.size();
+
+        if (textSize.height() < 70)
+            textSize.setHeight(70);
+        return textSize;
     }
```
veloman-yunkan: This still doesn't work correctly

[Screencast from 06-07-2024 01:55:59 PM.webm](https://github.com/kiwix/kiwix-desktop/assets/11933502/7daa8b59-5fb3-4cdb-8afb-bf481276783c)

veloman-yunkan: Maybe you have to account for some padding

veloman-yunkan (changes requested):

src/contentmanagermodel.cpp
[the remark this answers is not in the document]
veloman-yunkan: > 1. Where do you think is the best place to put isDescriptionLine(index)?

Please define it in `contentmanagermodel.h` as a free function

src/contentmanagerdelegate.cpp
ShaopengLin: Hmm either the header length or the boundingRect function is being inaccurate. I will try to find an appropriate width for the height calculation.

src/descriptionnode.cpp
```diff
@@ -24,6 +25,11 @@ QString DescriptionNode::getBookId() const
     return temp->getBookId();
 }
 
+bool DescriptionNode::isDescriptionIndex(const QModelIndex &index)
+{
+  return index.parent().isValid();
+}
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/contentmanagermodel.cpp
```diff
@@ -60,15 +60,18 @@ QVariant ContentManagerModel::data(const QModelIndex& index, int role) const
...
+    const bool otherDataRequest = col != 0 && role == Qt::DisplayRole;
 
-    if ( !isThumbnailRequest && !otherDataRequest )
+    if ( !isThumbnailRequest && !otherDataRequest && !isDescriptionRequest)
         return QVariant();
 
     const auto item = static_cast<Node*>(index.internalPointer());
     QVariant r = item->data(col);
-    if ( !isThumbnailRequest )
+    if ( !isThumbnailRequest || isDescriptionRequest )
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/contentmanagerdelegate.cpp
```diff
@@ -285,10 +285,26 @@ void ContentManagerDelegate::handleLastColumnClicked(const QModelIndex& index, Q
...
+        QRect lastColRect = view->visualRect(index.parent().siblingAtColumn(5));
+        r.setWidth(lastColRect.x() + lastColRect.width() - r.x());
+
+        /* Based on the rectangle and text, find the best fitting size. */
+        QFontMetrics fm(option.font);
+        QSize textRectSize = fm
+            .boundingRect(r,
+                          Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap,
+                          index.data().toString())
+            .size();
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/contentmanagermodel.cpp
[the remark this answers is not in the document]
ShaopengLin: Done

src/contentmanagerdelegate.cpp
```diff
@@ -285,10 +285,28 @@ void ContentManagerDelegate::handleLastColumnClicked(const QModelIndex& index, Q
...
+        /* Based on the rectangle and text, find the best fitting size. */
+        QFontMetrics fm(option.font);
+        const QString text = index.data().toString();
+        const auto format = Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap;
+        const QRect textRect = fm.boundingRect(descRect, format, text);
+        QSize textSize = textRect.size();
+
+        if (textSize.height() < 70)
+            textSize.setHeight(70);
+        return textSize;
```
veloman-yunkan, line 309: This still doesn't work well.

After a little experimenting myself I have arrived at the following code that seems to do the job:

```c++
        const auto treeView = KiwixApp::instance()->getContentManager()->getView()->getView();

        const int width = treeView->header()->length() - 2*treeView->indentation();
        // XXX: see QTreeView::padding in resources/css/_contentManager.css
        const int verticalPadding = 4;
        const int horizontalPadding = 4;
        QRect descRect(0, 0, width - 2 * horizontalPadding, 0);

        /* Based on the rectangle and text, find the best fitting size. */
        QFontMetrics fm(option.font);
        const QString text = index.data().toString();
        const auto format = Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap;
        const int textHeight = fm.boundingRect(descRect, format, text).height();
        return QSize(width, std::max(textHeight + verticalPadding, 70));
```

src/contentmanagerview.cpp
```diff
@@ -56,3 +57,21 @@ void ContentManagerView::onExpanded(QModelIndex index)
...
+void ContentManagerView::updateSizeHint()
+{
+    auto view = this->getView();
+    if (!view->isVisible())
+        return;
+
+    auto visibleIndex = view->indexAt(view->rect().topLeft());
+    while (visibleIndex.isValid())
+    {
+        if (visibleIndex.parent().isValid())
```
veloman-yunkan, line 73: Utilize `isDecriptionIndex()` here

veloman-yunkan (changes requested):

src/contentmanagermodel.cpp
```diff
@@ -86,7 +86,7 @@ QVariant ContentManagerModel::data(const QModelIndex& index, int role) const
 Qt::ItemFlags ContentManagerModel::flags(const QModelIndex &index) const
 {
     Qt::ItemFlags defaultFlags = QAbstractItemModel::flags(index);
-    if (index.isValid() && index.parent().isValid()) {
+    if (isDescriptionIndex(index)) {
```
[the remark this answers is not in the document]
ShaopengLin: Yep, the parent would be invalid if the index is invalid. Consequently, they have the same effect.

src/contentmanagermodel.cpp
```diff
@@ -60,10 +60,13 @@ QVariant ContentManagerModel::data(const QModelIndex& index, int role) const
         return QVariant();
 
     const auto col = index.column();
-    const bool isThumbnailRequest = col == 0 && role == Qt::DecorationRole;
-    const bool otherDataRequest = col != 0 && (role == Qt::DisplayRole || role == Qt::UserRole+1 );
+    const bool isThumbnailRequest =
+        !index.parent().isValid() && col == 0 && role == Qt::DecorationRole;
+    const bool isDescriptionRequest =
+        index.parent().isValid() && col == 0 && role == Qt::DisplayRole;
+    const bool otherDataRequest = col != 0 && role == Qt::DisplayRole;
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/contentmanagerdelegate.cpp
ShaopengLin: Done

src/contentmanagerview.cpp
ShaopengLin: Done

veloman-yunkan (commented):
LGTM

The branch must be rebased and the a conflict resolved. Then we will merge.

PR #1120, opened 2024-06-01
Remembers Previous Current Tab Index From Last Session

### LINKED ISSUE #1073: Remember current tab from previous user session
#1048 introduced the feature to reopen tabs on start-up. However, on many browsers (Firefox, Chrome), the current tab is also being kept track of when restoring user sessions. It would be good UX to have this feature.

One thing to discuss is: that we are not currently reopening invalid Zim tabs, empty tabs, settings tabs, etc. Users could be on one of those tabs from the previous session, including the library tab. How should we define the behavior of our tab reopenings?

ShaopengLin:
Fix #1073 
Changes:
- Blank and Settings tabs are now saved and reopened on startup.
- The current tab index from the previous session is recorded and restored.

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -207,7 +207,10 @@ void KiwixApp::restoreTabs()
     {
       for (const auto &zimUrl : tabsToOpen)
       {
-        openUrl(QUrl(zimUrl));
+        if (zimUrl.isEmpty())
+          getTabWidget()->createNewTab(true, true);
+        else
+          openUrl(QUrl(zimUrl));
       }
     }
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -207,13 +207,7 @@ void KiwixApp::restoreTabs()
     {
       for (const auto &zimUrl : tabsToOpen)
       {
-        try
-        {
-          /* Throws exception if zim file cannot be found */
-          m_library.getArchive(QUrl(zimUrl).host().split('.')[0]);
-          openUrl(QUrl(zimUrl));
-        }
-        catch (std::exception &e) { /* Blank */ }
+        openUrl(QUrl(zimUrl));
       }
     }
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -215,6 +215,9 @@ void KiwixApp::restoreTabs()
           openUrl(QUrl(zimUrl));
       }
     }
+
+    /* Restore current tab index. */
+    getTabWidget()->setCurrentIndex(mp_session->value("currentTabIndex", 0).toInt());
 }
 
 KiwixApp *KiwixApp::instance()
@@ -568,4 +571,9 @@ void KiwixApp::restoreWindowState()
 {
   getMainWindow()->restoreGeometry(mp_session->value("geometry").toByteArray());
   getMainWindow()->restoreState(mp_session->value("windowState").toByteArray());
+}
+
+void KiwixApp::saveCurrentTabIndex()
+{
+  return mp_session->setValue("currentTabIndex", getTabWidget()->currentIndex());
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -207,7 +207,9 @@ void KiwixApp::restoreTabs()
     {
       for (const auto &zimUrl : tabsToOpen)
       {
-        if (zimUrl.isEmpty())
+        if (zimUrl == "SettingsTab")
+          getTabWidget()->openOrSwitchToSettingsTab();
+        else if (zimUrl.isEmpty())
           getTabWidget()->createNewTab(true, true);
         else
           openUrl(QUrl(zimUrl));
--- a/src/kiwixapp.h
+++ b/src/kiwixapp.h
@@ -91,6 +91,7 @@ public:
     void saveListOfOpenTabs();
     void saveWindowState();
     void restoreWindowState();
+    void saveCurrentTabIndex();
 
 public slots:
     void newTab();
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -338,6 +338,8 @@ void TabBar::onCurrentChanged(int index)
         // In the future, other types of tabs can be added.
         // For example, About dialog, or Kiwix Server control panel.
     }
+
+    KiwixApp::instance()->saveCurrentTabIndex();
 }
 
 void TabBar::fullScreenRequested(QWebEngineFullScreenRequest request)
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -163,6 +163,7 @@ ZimView* TabBar::createNewTab(bool setCurrent, bool nextToCurrentTab)
     connect(tab, &ZimView::webActionEnabledChanged,
             this, &TabBar::onWebviewHistoryActionChanged);
 
+    KiwixApp::instance()->saveListOfOpenTabs();
     return tab;
 }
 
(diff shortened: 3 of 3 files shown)


src/kiwixapp.cpp
```diff
@@ -208,11 +208,21 @@ void KiwixApp::restoreTabs()
...
           /* Throws exception if zim file cannot be found */
           m_library.getArchive(QUrl(zimUrl).host().split('.')[0]);
           openUrl(QUrl(zimUrl));
         }
-        catch (std::exception &e) { /* Blank */ }
+        catch (std::exception &e)
+        { 
+          /* Open an empty tab for tabs that could not be loaded. */
+          openUrl(QUrl(""));
+        }
```
veloman-yunkan, line 225: This will result in the URL of the tab being lost. This is undesirable if the ZIM files are located on removable storage that was not mounted at the time when `kiwix-desktop` was launched.

A better option may be to open the tab with an error text in it.

@kelson42 What do you think?

src/kiwixapp.cpp
```diff
@@ -208,11 +208,21 @@ void KiwixApp::restoreTabs()
       {
         try
         {
+          if (zimUrl == "SettingsTab")
+          {
+            getTabWidget()->openOrSwitchToSettingsTab();
+            continue;
+          }
```
veloman-yunkan, line 215: I think this should be outside the `try` block

veloman-yunkan (changes requested):

src/kiwixapp.cpp
kelson42: Yes it should open the right link and print a proper error message (depending which exception is triggered)

src/kiwixapp.cpp
ShaopengLin: Try block is not needed since we are no longer checking if a Zim file exist.

src/kiwixapp.cpp
ShaopengLin: Done.

ShaopengLin:
> The error displayed when trying to restore a tab for a missing ZIM file does not contain any information about the ZIM file (name/path), which will complicate dealing with that situation (should I just close the tab or try to restore the ZIM file?).
> 
> After you fix that issue please structure the PR as follows:
> 
> 1. Revert the change that addressed [Fix #186: Introduced Tab Restoration on Restart #1048 (comment)](https://github.com/kiwix/kiwix-desktop/pull/1048#discussion_r1534070566) (special handling of failure to restore a tab, with explanation why it's needed)
> 2. Improve the error
> 3. Restoration of the Settings tab
> 4. Restoration of the current tab

@veloman-yunkan The only way I have to modify the error page is via Javascript, by injecting the supporting information into the body. Are we ok with this? The setHtml method simply doesn't work due to a `gio` tool error.

ShaopengLin:
In addition, I will be storing the name and path of the book along with the id in a triple. I believe we do not have a way to retrieve `kiwix::Book` objects after they are removed through id.

ShaopengLin:
Thanks! Indeed I can send content from here.

@veloman-yunkan I will store (or pass in from the save storage) the name and path information in the Zim or WebView when we do openUrl. This way we can use urls in the request handler to identify tabs with this information.

ShaopengLin:
@veloman-yunkan The third commit c03cdb3e07195173a65df01e0ca6c39682f89891 is rather large due to the complexity of storing and coordinating the name and path. Since we do not have a way to retrieve such information when the zim Id cannot be found by the Kiwix library, I think this is what we have to live with.

veloman-yunkan:
> @veloman-yunkan The third commit [c03cdb3](https://github.com/kiwix/kiwix-desktop/commit/c03cdb3e07195173a65df01e0ca6c39682f89891) is rather large due to the complexity of storing and coordinating the name and path. Since we do not have a way to retrieve such information when the zim Id cannot be found by the Kiwix library, I think this is what we have to live with.

I am not comfortable living with such a complex implementation of a relatively simple feature :)

Assuming that the ZIM file is still in the library, you can get the same information from the library. Of course there can be scenarios when the `library.xml` file is changed circumventing `kiwix-desktop` (for example, it is edited manually), but in such cases the error may be different.

Alternatively, we can encode the additional info (book name and path) in the URI (for example, in its optional username component). But that is a hack that can be justified only if obtaining the required info via the library is not acceptable.

veloman-yunkan (changes requested):

ShaopengLin:
> > @veloman-yunkan The third commit [c03cdb3](https://github.com/kiwix/kiwix-desktop/commit/c03cdb3e07195173a65df01e0ca6c39682f89891) is rather large due to the complexity of storing and coordinating the name and path. Since we do not have a way to retrieve such information when the zim Id cannot be found by the Kiwix library, I think this is what we have to live with.
> 
> I am not comfortable living with such a complex implementation of a relatively simple feature :)
> 
> Assuming that the ZIM file is still in the library, you can get the same information from the library. Of course there can be scenarios when the `library.xml` file is changed circumventing `kiwix-desktop` (for example, it is edited manually), but in such cases the error may be different.
> 
> Alternatively, we can encode the additional info (book name and path) in the URI (for example, in its optional username component). But that is a hack that can be justified only if obtaining the required info via the library is not acceptable.

@veloman-yunkan This is what I think as well. However, I believe the monitor directory will modify the library automatically if we remove the file from that directory. I could not get the information about the book after it was deleted with getArchive or getBookById. 

This alone makes me feel we should change URI. I believe I can embed them in the query parameters or as a part of the path, but what do you think?

veloman-yunkan:
> I believe the monitor directory will modify the library automatically if we remove the file from that directory.

I think that's true only if the book is removed while `kiwix-desktop` is running. Directory monitoring shouldn't modify the library on startup. I just checked that when `library.xml` is loaded entries with invalid path are preserved.

> I could not get the information about the book after it was deleted with getArchive or getBookById.

It's natural that `getArchive()` doesn't work for books with an invalid path. However `getBookById()` should work.

ShaopengLin:
> However getBookById() should work.

Nope, just tried and it simply throws an error as well.

veloman-yunkan:
Indeed. My earlier idea of how directory monitoring worked was incorrect.

A workaround may be to prevent directory monitoring from removing books for which there are open tabs. Similarly it may makes sense to prevent it from removing bookmarked books (so that the users get an informative error message when they try to open a bookmark).

Instead of removing the book in those case, directory monitoring may just mark the book as having invalid path (`kiwix::Book::setPathValid(false)`).

ShaopengLin:
> A workaround may be to prevent directory monitoring from removing books for which there are open tabs. Similarly it may makes sense to prevent it from removing bookmarked books (so that the users get an informative error message when they try to open a bookmark).

I confirm this works. Will put a PR up tomorrow...

kelson42:
> Indeed. My earlier idea of how directory monitoring worked was incorrect.

I'm worried by this sentence and I hope this is only because I don't understand properly. There are many valid reasons why the directory monitoring can remove a file from the library and all of them are external to Kiwix. We have no control on this. The rest must follow.

Saved tabs should be either in error state (blank, proper error) or not restored.

bookmarks is out-of-scope here.

ShaopengLin:
I will modify the change to avoid manipulating the behavior of monitoring.

veloman-yunkan:
> > Indeed. My earlier idea of how directory monitoring worked was incorrect.
> 
> I'm worried by this sentence and I hope this is only because I don't understand properly. There are many valid reasons why the directory monitoring can remove a file from the library and all of them are external to Kiwix. We have no control on this. The rest must follow.
> 
> Saved tabs should be either in error state (blank, proper error) or not restored. Having information about what is missing is nice to have. IMHO, just displaying the zim:// URL should be enough. In best case open a dedicated issue t think how we could get a better error.
> 
> bookmarks is out-of-scope here.

Agree. Let's split this PR into two:

1. Changes immediately relevant to restoring the active tab (mostly one of the earlier forms of this PR, before the work on the second part started).
2. Improvement of the error page displayed in case of a missing ZIM file (this will also include the change related to the handling of bookmarks).

ShaopengLin:
@veloman-yunkan done

veloman-yunkan (approved):

PR #1123, opened 2024-06-04
Show Open/Save dialog on downloading files

### LINKED ISSUE #697: PDF file dialog box should allow to open PDF file directly with PDF reader

juuz0:
Fix #697 

Superseeds #1070

### DIFF
--- a/src/kiwixmessagebox.cpp
+++ b/src/kiwixmessagebox.cpp
@@ -1,11 +1,11 @@
-#include "kiwixconfirmbox.h"
-#include "ui_kiwixconfirmbox.h"
+#include "kiwixmessagebox.h"
+#include "ui_kiwixmessagebox.h"
 #include <QFile>
 #include "kiwixapp.h"
 
-KiwixConfirmBox::KiwixConfirmBox(QString confirmTitle, QString confirmText, bool okDialog, QWidget *parent) :
+KiwixMessageBox::KiwixMessageBox(QString confirmTitle, QString confirmText, bool okDialog, QWidget *parent) :
     QDialog(parent), m_confirmTitle(confirmTitle), m_confirmText(confirmText),
-    ui(new Ui::kiwixconfirmbox)
+    ui(new Ui::kiwixmessagebox)
 {
     ui->setupUi(this);
     setWindowFlag(Qt::FramelessWindowHint, true);
@@ -32,16 +32,16 @@ KiwixConfirmBox::KiwixConfirmBox(QString confirmTitle, QString confirmText, bool
     }
 }
 
-KiwixConfirmBox::~KiwixConfirmBox()
+KiwixMessageBox::~KiwixMessageBox()
 {
     delete ui;
 }
 
 void showInfoBox(QString title, QString text, QWidget *parent)
 {
-    KiwixConfirmBox *dialog = new KiwixConfirmBox(title, text, true, parent);
+    KiwixMessageBox *dialog = new KiwixMessageBox(title, text, true, parent);
     dialog->show();
-    QObject::connect(dialog, &KiwixConfirmBox::okClicked, [=]() {
+    QObject::connect(dialog, &KiwixMessageBox::okClicked, [=]() {
         dialog->deleteLater();
     });
 }
--- a/src/kiwixmessagebox.cpp
+++ b/src/kiwixmessagebox.cpp
@@ -64,4 +64,3 @@ KiwixMessageBox::Result showKiwixMessageBox(QString title, QString text, QWidget
     });
     return dialog->execDialog();
 }
-
--- a/src/kiwixmessagebox.cpp
+++ b/src/kiwixmessagebox.cpp
@@ -3,26 +3,36 @@
 #include <QFile>
 #include "kiwixapp.h"
 
-KiwixMessageBox::KiwixMessageBox(QString confirmTitle, QString confirmText, bool okDialog, QWidget *parent) :
+KiwixMessageBox::KiwixMessageBox(QString confirmTitle, QString confirmText, bool okDialog, QWidget *parent,
+                                 QString leftAction, QString rightAction) :
     QDialog(parent), m_confirmTitle(confirmTitle), m_confirmText(confirmText),
     ui(new Ui::kiwixmessagebox)
 {
     ui->setupUi(this);
     setWindowFlag(Qt::FramelessWindowHint, true);
-    setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/confirmBox.css"));
+    setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/messageBox.css"));
     connect(ui->yesButton, &QPushButton::clicked, [=]() {
         emit yesClicked();
+        m_result = YesClicked;
+        accept();
     });
     connect(ui->noButton, &QPushButton::clicked, [=]() {
         emit noClicked();
+        m_result = NoClicked;
+        reject();
     });
     connect(ui->okButton, &QPushButton::clicked, [=]() {
         emit okClicked();
+        m_result = OkClicked;
+    });
+    connect(ui->closeButton, &QPushButton::clicked, [=]() {
+        this->close();
+        m_result = CloseClicked;
     });
     ui->confirmText->setText(confirmText);
     ui->confirmTitle->setText(confirmTitle);
-    ui->yesButton->setText(gt("yes"));
-    ui->noButton->setText(gt("no"));
+    ui->yesButton->setText(leftAction);
+    ui->noButton->setText(rightAction);
     ui->okButton->setText(gt("ok"));
     ui->okButton->hide();
     if (okDialog) {
@@ -45,3 +55,13 @@ void showInfoBox(QString title, QString text, QWidget *parent)
         dialog->deleteLater();
     });
 }
+
+KiwixMessageBox::Result showKiwixMessageBox(QString title, QString text, QWidget *parent, QString leftTitle, QString rightTitle)
+{
+    KiwixMessageBox *dialog = new KiwixMessageBox(title, text, false, parent, leftTitle, rightTitle);
+    QObject::connect(dialog, &KiwixMessageBox::finished, [=]() {
+        dialog->deleteLater();
+    });
+    return dialog->execDialog();
+}
+
--- a/src/kprofile.cpp
+++ b/src/kprofile.cpp
@@ -1,8 +1,9 @@
 #include "kprofile.h"
 
 #include "kiwixapp.h"
+#include "kiwixmessagebox.h"
+
 #include <QFileDialog>
-#include <QMessageBox>
 #include <QWebEngineSettings>
 #include <QWebEngineScript>
 #include <QWebEngineScriptCollection>
@@ -64,7 +65,7 @@ KProfile::KProfile(QObject *parent) :
 #endif
 
     scripts()->insert(getScript(":/js/headerAnchor.js"));
-    scripts()->insert(getScript(":/qtwebchannel/qwebchannel.js", 
+    scripts()->insert(getScript(":/qtwebchannel/qwebchannel.js",
                       QWebEngineScript::DocumentCreation));
 }
 
@@ -112,9 +113,10 @@ void KProfile::startDownload(WebEngineDownloadType* download)
 
 void KProfile::downloadFinished()
 {
-    QMessageBox msgBox;
-    msgBox.setText(gt("download-finished-message"));
-    msgBox.exec();
+    showInfoBox(gt("download-finished"),
+                gt("download-finished-message"),
+                KiwixApp::instance()->getMainWindow()
+    );
 }
 
 void ExternalReqInterceptor::interceptRequest(QWebEngineUrlRequestInfo &info)
--- a/src/kprofile.cpp
+++ b/src/kprofile.cpp
@@ -7,6 +7,8 @@
 #include <QWebEngineSettings>
 #include <QWebEngineScript>
 #include <QWebEngineScriptCollection>
+#include <QDesktopServices>
+#include <QTemporaryFile>
 
 namespace
 {
@@ -98,11 +100,32 @@ QString getDownloadFilePath(WebEngineDownloadType* download)
 
 } // unnamed namespace
 
-void KProfile::startDownload(WebEngineDownloadType* download)
+void KProfile::openFile(WebEngineDownloadType* download)
+{
+    const QString defaultFileName = getDownloadFilePath(download);
+    QTemporaryFile tempFile(QDir::tempPath() + "/XXXXXX." + QFileInfo(defaultFileName).suffix());
+    tempFile.setAutoRemove(false);
+    if (tempFile.open()) {
+        QString tempFilePath = tempFile.fileName();
+        tempFile.close();
+        setDownloadFilePath(download, tempFilePath);
+        connect(download, &DownloadFinishedSignal, [tempFilePath]() {
+            if(!QDesktopServices::openUrl(QUrl::fromLocalFile(tempFilePath)))
+                showInfoBox(gt("error-title"), gt("error-opening-file"), KiwixApp::instance()->getMainWindow());
+        });
+        download->accept();
+    } else {
+        qDebug()<<"tmp file err";
+        download->cancel();
+    }
+}
+
+void KProfile::saveFile(WebEngineDownloadType* download)
 {
     const QString defaultFileName = getDownloadFilePath(download);
     const QString fileName = askForSaveFilePath(defaultFileName);
     if (fileName.isEmpty()) {
+        download->cancel();
         return;
     }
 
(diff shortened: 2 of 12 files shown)


kelson42:
@juuz0 Thx a lot for your PR. Il will test ot in the next hours. Just to be sure: you went through all the comments of the superseeded PR to ensure we are on rge right track?

juuz0:
@kelson42 Yes I believe so.

kelson42 (approved):
@juuz0 Works almost exactly like it should. Just one comment: Can you please have a better choice of directory following [this logic](https://github.com/kiwix/kiwix-desktop/pull/1058#issuecomment-2067999191). Actually it could use the same settings (to keep directory path, would be the same as "import").

src/kprofile.h
```diff
@@ -19,6 +19,7 @@ class KProfile : public QWebEngineProfile
 
 private:
     UrlSchemeHandler m_schemeHandler;
+    QWebEngineDownloadItem *currentDownloadItem;
```
veloman-yunkan, line 22: Unused

src/kprofile.h
```diff
@@ -28,7 +29,15 @@ public slots:
...
-    void downloadFinished();
+
+private slots:
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+    void saveFile(QWebEngineDownloadItem*);
+    void openFile(QWebEngineDownloadItem*);
+#else
+    void saveFile(QWebEngineDownloadRequest*);
+    void openFile(QWebEngineDownloadRequest*);
+#endif
```
veloman-yunkan, line 40: You can minimize Qt 5-vs-6 conditional compilation (including the old code) as follows:

```c++
#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
typedef WebEngineDownloadType QWebEngineDownloadItem;
#else
typedef WebEngineDownloadType QWebEngineDownloadRequest;
#endif
```

src/kprofile.cpp
```diff
@@ -19,15 +25,43 @@ KProfile::KProfile(QObject *parent) :
...
+        download->setPath(tempFilePath);
+#else
+        download->setDownloadFileName(tempFilePath);
+#endif
+        connect(download, &QWebEngineDownloadItem::finished, [tempFilePath]() {
+            if(!QDesktopServices::openUrl(QUrl::fromLocalFile(tempFilePath)))
+                showInfoBox(gt("error-title"), gt("error-opening-file-with"), KiwixApp::instance()->getMainWindow());
+        });
+        download->accept();
+    } else {
```
veloman-yunkan, line 48: Memory leak - `tempFile` is not `delete`d. You can  create the `QTemporaryFile` object on the stack (rather than on the heap) and [`disable auto-remove`](https://doc.qt.io/qt-5/qtemporaryfile.html#setAutoRemove).

src/kprofile.cpp
```diff
@@ -19,15 +25,43 @@ KProfile::KProfile(QObject *parent) :
...
+        QString tempFilePath = tempFile->fileName();
+        tempFile->close();
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
+        download->setPath(tempFilePath);
+#else
+        download->setDownloadFileName(tempFilePath);
+#endif
+        connect(download, &QWebEngineDownloadItem::finished, [tempFilePath]() {
+            if(!QDesktopServices::openUrl(QUrl::fromLocalFile(tempFilePath)))
+                showInfoBox(gt("error-title"), gt("error-opening-file-with"), KiwixApp::instance()->getMainWindow());
```
veloman-yunkan, line 45: The temporary file will be left around. Shouldn't we remove it?

veloman-yunkan (changes requested):

src/kprofile.cpp
kelson42: I don't think so, but should be written in OS temporary directory.

juuz0: QTemporaryFile saves it in the tmp directory anyway so I havent changed anything here.

src/kprofile.cpp
juuz0: Its on the stack now but I havent disabled auto-remove, why do we need that?

src/kprofile.h
juuz0: Very neat! Thank you
(Fun fact: I spent an embarrassing amount of time deciding if I should use "using" or "typedef" :>)

src/kprofile.cpp
```diff
@@ -19,15 +25,43 @@ KProfile::KProfile(QObject *parent) :
...
-                                                       gt("save-file-as-window-title"), defaultFileName);
+    QTemporaryFile *tempFile = new QTemporaryFile(QDir::tempPath() + "/XXXXXX." + QFileInfo(defaultFileName).suffix());
+    if (tempFile->open()) {
+        QString tempFilePath = tempFile->fileName();
+        tempFile->close();
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
+        download->setPath(tempFilePath);
+#else
+        download->setDownloadFileName(tempFilePath);
+#endif
```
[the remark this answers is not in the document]
juuz0: Done

juuz0:
> Can you please have a better choice of directory following https://github.com/kiwix/kiwix-desktop/pull/1058#issuecomment-2067999191. Actually it could use the same settings (to keep directory path, would be the same as "import").

@kelson42 done! please check.

And I also rebased it on main (CI seems to be failing because of that?)

juuz0:
Fixed an issue where the compilation was failing for QT6 (i was still at 5.15 so it slipped, upgraded now)

src/kprofile.cpp
veloman-yunkan: Then the file will be deleted when `tempFile` goes out of scope. Since the file is opened asynchronously (after its download is finished) that operation may fail.

src/kprofile.cpp
```diff
@@ -24,39 +23,35 @@ KProfile::KProfile(QObject *parent) :
...
-void KProfile::openFile(QWebEngineDownloadItem* download)
+void setDownloadFilePath(KProfile::WebEngineDownloadType* download, QString filePath)
+{
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
+    download->setPath(filePath);
 #else
-void KProfile::openFile(QWebEngineDownloadRequest* download)
+    download->setDownloadFileName(filePath);
 #endif
+}
```
veloman-yunkan: Hide in unnamed namespace

src/kprofile.h
veloman-yunkan: My bad. What a silly mistake.

veloman-yunkan (changes requested):

veloman-yunkan (changes requested):

src/kprofile.cpp
```diff
@@ -18,40 +23,66 @@ KProfile::KProfile(QObject *parent) :
...
+void KProfile::openFile(WebEngineDownloadType* download)
+{
+    QString defaultFileName = download->url().fileName();
+    QTemporaryFile tempFile(QDir::tempPath() + "/XXXXXX." + QFileInfo(defaultFileName).suffix());
+    tempFile.setAutoRemove(false);
+    if (tempFile.open()) {
+        QString tempFilePath = tempFile.fileName();
+        tempFile.close();
+        setDownloadFilePath(download, tempFilePath);
+        connect(download, &WebEngineDownloadType::finished, [tempFilePath]() {
```
veloman-yunkan, line 46: This still shouldn't compile under Qt6 (because the signal name is `isFinished()` there).

src/kprofile.cpp
```diff
@@ -18,40 +23,66 @@ KProfile::KProfile(QObject *parent) :
...
-#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
-    download->setPath(fileName);
-#else
-    download->setDownloadFileName(fileName);
-#endif
-#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
-    connect(download, &QWebEngineDownloadItem::finished, this, &KProfile::downloadFinished);
-#else
-    connect(download, &QWebEngineDownloadRequest::isFinished, this, &KProfile::downloadFinished);
-#endif
```
veloman-yunkan, line 46: Did you eliminate `KProfile::downloadFinished()` intentionally? Now when the file is saved no `gt("download-finished-message")` message will be shown to the user.

veloman-yunkan (changes requested):

src/kprofile.cpp
juuz0: Yes, it was intentional. I forgot to add the download finished message in saveFile().

src/kprofile.cpp
juuz0: Seems like I can't typedef a signal, have used #define here, is that ok?

src/kprofile.cpp
```diff
@@ -18,40 +23,66 @@ KProfile::KProfile(QObject *parent) :
...
+        download->cancel();
+    }
+}
+
+void KProfile::saveFile(WebEngineDownloadType* download)
 {
     QString defaultFileName = download->url().fileName();
-    QString fileName = QFileDialog::getSaveFileName(KiwixApp::instance()->getMainWindow(),
-                                                       gt("save-file-as-window-title"), defaultFileName);
+    QString fileName = QFileDialog::getSaveFileName(KiwixApp::instance()->getMainWindow(), "Save File", defaultFileName);
```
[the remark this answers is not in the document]
juuz0: Oversight. Will look into future PRs/commits more closely, apologies :)

src/kiwixmessagebox.cpp
```diff
@@ -45,3 +55,10 @@ void showInfoBox(QString title, QString text, QWidget *parent)
         dialog->deleteLater();
     });
 }
+
+KiwixMessageBox::Result showKiwixMessageBox(QString title, QString text, QWidget *parent, QString leftTitle, QString rightTitle)
+{
+    KiwixMessageBox *dialog = new KiwixMessageBox(title, text, false, parent, leftTitle, rightTitle);
+    return dialog->execDialog();
+}
```
veloman-yunkan, line 63: Is `dialog` somehow deleted automatically?

veloman-yunkan (commented):
Please rebase and fixup (which should require some manual conflict resolution) for the next (hopefully, final) iteration.

src/kiwixmessagebox.cpp
juuz0: Yes, along with the parent (when the parent - mainWindow here- is deleted).

veloman-yunkan: That's a memory leak because the memory allocated to each new (displayed and closed) dialog is not reclaimed until the end of the application.

juuz0: right right

src/kiwixmessagebox.cpp
```diff
@@ -45,3 +55,11 @@ void showInfoBox(QString title, QString text, QWidget *parent)
         dialog->deleteLater();
     });
 }
+
+KiwixMessageBox::Result showKiwixMessageBox(QString title, QString text, QWidget *parent, QString leftTitle, QString rightTitle)
+{
+    KiwixMessageBox *dialog = new KiwixMessageBox(title, text, false, parent, leftTitle, rightTitle);
+    dialog->setAttribute(Qt::WA_DeleteOnClose);
```
veloman-yunkan, line 62: This is an interesting finding. Can't we use the same trick in `showInfoBox()`? Or maybe even move it into the constructor of `KiwixMessageBox`?

juuz0: Moving into the constructor is a nice idea, I did it in the new fixup. 
`showConfirmBox()` also uses execDialog() now - It looks more natural to me that a confirmBox() is high prio and should be handled before interacting with other windows (hence the `exec()` call)

veloman-yunkan: Now that I looked into the documentation of `QDialog::exec()` and `Qt::WA_DeleteOnClose`, I have some doubts. When `QDialog::exec()` returns the dialog should have closed and hence deleted. Therefore   `KiwixMessageBox::execDialog()` will execute its last few instructions on a deleted object. Am I wrong?

veloman-yunkan: Also, I am also concerned with `QDialog::exec()` being a blocking function. I wonder how download updater code will react to the the main thread being blocked.

veloman-yunkan (commented):
My review for the new changes came in the form of two more comments ([1](https://github.com/kiwix/kiwix-desktop/pull/1123#discussion_r1643159986), [2](https://github.com/kiwix/kiwix-desktop/pull/1123#discussion_r1643162362)) on a previous thread. I am posting this response to the most recent review request to make it explicit that it has not been ignored.

kelson42:
> @juuz0 Works almost exactly like it should. Just one comment: Can you please have a better choice of directory following [this logic](https://github.com/kiwix/kiwix-desktop/pull/1058#issuecomment-2067999191). Actually it could use the same settings (to keep directory path, would be the same as "import").

@juuz0 It seems this part is not implemented... it seems to always propose to save in the current path which is really not correct.

src/kiwixmessagebox.cpp
juuz0: > Now that I looked into the documentation of QDialog::exec() and Qt::WA_DeleteOnClose ..........

Sorry for the delay, life happened!
You're right, I looked into qdialog's source code and [when deleteOnClose is enabled](https://codebrowser.dev/qt5/qtbase/src/widgets/dialogs/qdialog.cpp.html#612) it deletes it instantly. I reverted it to the old way of ours - a simple deleteLater().

> Also, I am also concerned with QDialog::exec() being a blocking function.....

I don't really get this, we aren't starting the download until open/save is clicked (unless you mean the Download Books functionality, which I agree could've been an issue and already reverted...)

juuz0:
@kelson42 It works for me, can you please recheck? (unless I misunderstood what was needed)

In the attached video, original location is `common`, I save the file, then start another save file session - save it in `tmp` and in the following saves, it remembers `tmp`  

[Screencast from 2024-07-03 15-57-18.webm](https://github.com/kiwix/kiwix-desktop/assets/26814569/933c01c9-e90a-4917-ab95-9b95bac8eb82)

veloman-yunkan:
I will review the PR again after @kelson42 approves it from user's perspective.

kelson42:
@juuz0 Retested and indeed it work fine but I have a last topic: what happens if there is no know app to open the file? I though I would have a choice of app to open the file? but instead I get a filepicker, this is not correct.

If the app to open the file is not known, then we should display the list of app and the user should choose which app he wants to use to open the file.

kelson42:
@juuz0 ? Rebasing on `main` to resolve red CI

kelson42:
@juuz0 ?

juuz0:
@kelson42 sooon! (im still trying to figure out how to get the open-with selection).
QDesktopServices calls xdg-open which thinks the file manager/picker is the correct application for that specific file. I'll try finding a workaround.

kelson42:
@juuz0 Any update? Would really like to merge this PR which is slowly getting old.

kelson42:
@juuz0 Any update? We need this PR for `2.4.0`!

juuz0:
Apologies @kelson42, I'm equipped with several things IRL - cant get the time, and yeah if someone is willing to take over it, please do! :>

My 2 cents:
The issue is xdg-open (called by QDesktopServices) which thinks the file manager/picker is the correct application for that specific file (it works ok for pdf but epub files open the manager).

One workaround is using `xdg-mime`. For mime-types which dont have a specific application set, `xdg-mime` returns nothing so if we can disable the Open button when xdg-mime query is empty, that could help.
BUT remember `xdg-mime` is not available on windows obviously - its possible QDesktopServices opens the application-picker-menu on windows though...up to testing!

kelson42:
@veloman-yunkan AFAIK this is the last piece of work which is assigned to you in the `2.4.0` milestone.

veloman-yunkan:
> While rebasing dropped commit [3ce2423](https://github.com/kiwix/kiwix-desktop/commit/3ce2423675501169f5e677ba7b72120f3e011567) since a similar change [3988067](https://github.com/kiwix/kiwix-desktop/commit/3988067dde5eec3b6af6cdbbaf7376ea3a1f0536) from #1134 made it redundant.

The said two commits use different settings for remembering the previous directory used for open/save operations:

1. 3ce2423675501169f5e677ba7b72120f3e011567  (from the previous version of this PR) uses the directory of the last "Open file..." operation. Thus in this PR choosing to save a file instead of opening
2. 3988067dde5eec3b6af6cdbbaf7376ea3a1f0536 (from #1134) introduced a separate persistent setting for the directory of the "Save As..." operation.

I think that behaviour 2. is more appropriate since the content being saved from a ZIM file is of a different kind and shouldn't affect the location for opening ZIM files.

veloman-yunkan:
> Retested and indeed it work fine but I have a last topic: what happens if there is no known app to open the file? I though I would have a choice of app to open the file? But instead I get a filepicker, this is not correct.
> 
> If the app to open the file is not known, then we should display the list of app and the user should choose which app he wants to use to open the file. If this is not possible, then we should probably just disabled the button in the dialog box so the user can only save to a file.

@kelson42 What do you mean by filepicker?

In my case (on Ubuntu 22.04) I only get an error message:

> `gio: file:///tmp/eXSEiM.o: No application is registered as handling this file`

Qt doesn't seem to provide a way of finding out the application that will be used to open the given file. In my case it didn't even detect that the attempt to open the file failed, which is a [documented](https://doc.qt.io/qt-5/qdesktopservices.html#openUrl) known issue:

> **Warning:** A return value of true indicates that the application has successfully requested the operating system to open the URL in an external application. The external application may still fail to launch or fail to open the requested URL. This result will not be reported back to the application.

kelson42:
> Qt doesn't seem to provide a way of finding out the application that will be used to open the given file. In my case it didn't even detect that the attempt to open the file failed,

Then I guess we should go forward with what we have.

kelson42:
@veloman-yunkan Any news here?

veloman-yunkan:
> Then I guess we should go forward with what we have. Please rebase and make the PR ready in general so we can merge.

@kelson42 Done

PR #1125, opened 2024-06-06
Download path setting QOL improvements

### LINKED ISSUE #1113: Few improvements around Settings directories
See what I have:
image

Here are the problems:
* The whole directory (if long) is not displayed. I propose to implement proper ellipsis display (ending with ...)
* We should have a hint, when move over the directory, displaying the whole path
* Right click on the path, should allow to copy it

sgourdas:
Fixes #1113 

Added functionality for the download path in settings:
- Tooltip when hovering over, to show the full path.
- Copy to clipboard icon, with a disappearing popup to indicate success.
- Truncation of the path in case it exceeds a limit. `...` replaces the hidden path, to allow for better readability of the directory's path.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -135,6 +135,7 @@
     "monitor-dir-dialog-title":"Are you sure you want to change the monitor directory?",
     "monitor-dir-dialog-msg":"The new monitor directory path will be:\n{{DIRECTORY}}",
     "monitor-clear-dir-dialog-title":"Are you sure you want to clear the monitor directory?",
+    "path-was-copied": "Path was copied",
     "monitor-clear-dir-dialog-msg":"This will stop checking the monitor directory for new ZIM files.",
     "monitor-directory-tooltip":"All ZIM files in this directory will be automatically added to the library.",
     "next-tab":"Move to next tab",
--- a/src/settingsview.cpp
+++ b/src/settingsview.cpp
@@ -2,8 +2,24 @@
 #include "ui_settings.h"
 #include "kiwixapp.h"
 #include <kiwix/tools.h>
+#include <QClipboard>
 #include <QMessageBox>
 #include <QFileDialog>
+#include <QToolTip>
+
+namespace 
+{
+    QString formatDownloadDir(const QString& input) {
+        const int maxLength = 40;
+        if (input.length() > maxLength) {
+            QString suffix = input.right(maxLength);
+            int directoryIndex = suffix.indexOf('/');
+            return "..." + suffix.mid(directoryIndex != -1 ? directoryIndex : 0);
+        }
+        return input;
+    }
+}
+
 SettingsView::SettingsView(QWidget *parent)
     : QWidget(parent)
     , ui(new Ui::Settings)
@@ -14,6 +30,7 @@ SettingsView::SettingsView(QWidget *parent)
     connect(ui->moveToTrashToggle, &QCheckBox::clicked, this, &SettingsView::setMoveToTrash);
     connect(ui->reopenTabToggle, &QCheckBox::clicked, this, &SettingsView::setReopenTab);
     connect(ui->browseButton, &QPushButton::clicked, this, &SettingsView::browseDownloadDir);
+    connect(ui->downloadDirPathCopy, &QPushButton::clicked, this, &SettingsView::copyDownloadPathToClipboard);
     connect(ui->resetButton, &QPushButton::clicked, this, &SettingsView::resetDownloadDir);
     connect(ui->monitorBrowse, &QPushButton::clicked, this, &SettingsView::browseMonitorDir);
     connect(ui->monitorClear, &QPushButton::clicked, this, &SettingsView::clearMonitorDir);
@@ -30,6 +47,9 @@ SettingsView::SettingsView(QWidget *parent)
     ui->browseButton->setText(gt("browse"));
     ui->monitorClear->setText(gt("clear"));
     ui->monitorBrowse->setText(gt("browse"));
+    QIcon copyIcon(":/icons/copy.svg");
+    ui->downloadDirPathCopy->setIcon(copyIcon);
+    ui->downloadDirPathCopy->setIconSize(QSize(24, 24));
     ui->monitorHelp->setText("<b>?</b>");
     ui->monitorHelp->setToolTip(gt("monitor-directory-tooltip"));
     ui->moveToTrashLabel->setText(gt("move-files-to-trash"));
@@ -47,7 +67,7 @@ void SettingsView::init(int zoomPercent, const QString &downloadDir,
                         bool reopentab)
 {
     ui->zoomPercentSpinBox->setValue(zoomPercent);
-    ui->downloadDirPath->setText(downloadDir);
+    SettingsView::onDownloadDirChanged(downloadDir);
     if (monitorDir == QString()) {
         ui->monitorClear->hide();
     }
@@ -156,7 +176,17 @@ void SettingsView::setReopenTab(bool reopen)
 
 void SettingsView::onDownloadDirChanged(const QString &dir)
 {
-    ui->downloadDirPath->setText(dir);
+    ui->downloadDirPath->setText(formatDownloadDir(dir));
+    ui->downloadDirPath->setToolTip(dir);
+}
+
+void SettingsView::copyDownloadPathToClipboard()
+{
+    QString downloadPath = KiwixApp::instance()->getSettingsManager()->getDownloadDir();
+    QApplication::clipboard()->setText(downloadPath);
+
+    QPoint globalPos = ui->downloadDirPathCopy->mapToGlobal(QPoint(0, -ui->downloadDirPathCopy->height()));
+    QToolTip::showText(globalPos, gt("path-was-copied"), ui->downloadDirPathCopy);
 }
 
 void SettingsView::onMonitorDirChanged(const QString &dir)
--- a/resources/css/_settingsManager.css
+++ b/resources/css/_settingsManager.css
@@ -43,6 +43,17 @@ QPushButton:hover {
     color: white;
 }
 
+#downloadDirPathCopy {
+    background-color: none;
+    border: 1px solid transparent;
+}
+
+#downloadDirPathCopy:hover {
+    background-color: #D9E9FF;
+    border: 1px solid #3366CC;
+    border-radius: 3px;
+}
+
 #monitorDirLabel {
     padding-right: 0;
 }
--- a/resources/i18n/qqq.json
+++ b/resources/i18n/qqq.json
@@ -174,5 +174,6 @@
 	"clear-filter": "Represents the action of clearing the filters selected for a filter type.",
 	"no-details": "A content type for Zim files representing it only has an introduction.",
 	"no-pictures": "A content type for Zim files that does not contain pictures.",
-	"no-videos": "A content type for Zim files that does not contain videos."
+	"no-videos": "A content type for Zim files that does not contain videos.",
+	"path-was-copied": "Tooltip confirming that the download path from settings was copied."
 }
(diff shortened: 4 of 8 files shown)


veloman-yunkan:
@kelson42 I will review the code after you approve it from a user's perspective

kelson42:
@sgourdas Can you please systematically fix the CodeFactor if possible?

kelson42 (changes requested):
> Ability to click on it to copy the full path of the directory to the clipboard

Right click on the path does nothing. I'm not able to select the path either... not sure here if it simply does not work or if I do something wrong.

sgourdas:
> Can you please systematically fix the CodeFactor if possible? 

Sorry for this. Changed my settings now to avoid it.

> Can you please in the future create your feature-branch in our repository and not in your fork?

Sure, will do so from the next ones.

> Right click on the path does nothing.

My bad on not mentioning it here as well, but I thought it would be better if we assign left click to the copy to clipboard function, as it seems more intuitive. What do you think?

> I'm not able to select the path either... not sure here if it simply does not work or if I do something wrong.

Regarding this I have not implemented something to support it. Should I work on it?

kelson42:
> Right click on the path does nothing.
> My bad on not mentioning it here as well, but I thought it would be better if we assign left click to the copy to clipboard function, as it seems more intuitive. What do you think?

It should be right Click, a Contest Menu should be dieplayed with "Copy"

veloman-yunkan:
> It should be right Click, a Contest Menu should be dieplayed with "Copy"

I think that the current UI/UX best practice is to provide a dedicated copy button beside the field.

kelson42:
@veloman-yunkan Agree

sgourdas:
@kelson42 so just to make sure I understood: no click to copy, just a copy button like the rest of the ui has, correct?

kelson42:
> @kelson42 so just to make sure I understood: no click to copy, just a copy button like the rest of the ui has, correct?

not even a button, just a small standard copy logo on the right, like youncan find everywhere in apps and web sites.

kelson42:
@sgourdas What is the status here?  Nothing has changed since 3 days and CInis still red!

src/settingsview.cpp
```diff
@@ -30,6 +34,7 @@ SettingsView::SettingsView(QWidget *parent)
     ui->browseButton->setText(gt("browse"));
     ui->monitorClear->setText(gt("clear"));
     ui->monitorBrowse->setText(gt("browse"));
+    ui->downloadDirPathCopy->setText("<img src=':/resources/icons/copy.svg'>");
```
veloman-yunkan, line 37: This doesn't seem to use the `copy.svg` icon included in this PR:

Screenshot from 2024-06-16 15-06-48

sgourdas: For some reason I just assumed this was render issue on my side, but of course it is not. Just used the wrong path relative to the .qrc file

src/settingsview.cpp
```diff
@@ -47,7 +52,8 @@ void SettingsView::init(int zoomPercent, const QString &downloadDir,
                         bool reopentab)
 {
     ui->zoomPercentSpinBox->setValue(zoomPercent);
-    ui->downloadDirPath->setText(downloadDir);
+    ui->downloadDirPath->setText(formatDownloadDir(downloadDir));
+    ui->downloadDirPath->setToolTip(downloadDir);
```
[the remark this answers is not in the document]
sgourdas: Definitely

veloman-yunkan (commented):
It works but I don't like that there is no kind of feedback when the copy label is clicked.

@kelson42 If you don't mind about the above concern then we can merge this PR (as one squashed commit).

veloman-yunkan:
> > It works but I don't like that there is no kind of feedback when the copy label is clicked.
> 
> Agree, but not sure what could be a good feedback. Domyou have something to propose?

A tooltip reading "Path was copied" could appear for a short time.

kelson42:
@sgourdas Can you implement this that the user knows that the copy operation has been made.

sgourdas:
Added a disappearing popup, after copying the path and additionally implemented highlight on hover to match the rest of the application.

src/settingsview.cpp
```diff
@@ -156,7 +163,31 @@ void SettingsView::setReopenTab(bool reopen)
...
+    popup->setAttribute(Qt::WA_DeleteOnClose);
+    popup->setText(message);
+    popup->setStandardButtons(QMessageBox::NoButton);
+
+    popup->move(QGuiApplication::primaryScreen()->geometry().center());
+
+    popup->show();
+
+    QTimer::singleShot(1250, popup, &QObject::deleteLater);
 }
```
veloman-yunkan, line 191: A popup window doesn't look nice. Couldn't it be implemented as a tooltip (near the click position rather than in the center of the screen)?

sgourdas: I am not sure of what you are describing. Do you have a specific class/concept in mind?

src/clickableqlabel.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+ClickableQLabel::~ClickableQLabel() {}
+
+void ClickableQLabel::mousePressEvent(QMouseEvent* event) {
+    emit clicked();
+    QLabel::mousePressEvent(event);
+}
+
+void ClickableQLabel::enterEvent(QEvent* event) {
+    setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/style.css"));
```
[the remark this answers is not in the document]
sgourdas: My bad, these were committed on a first iteration of the current version and forgotten.

src/settingsview.cpp
veloman-yunkan: https://doc.qt.io/qt-5/qtooltip.html

sgourdas: Tried something like this before but could not get it to display for some reason:
```
QPoint globalPos = ui->downloadDirPathCopy->mapToGlobal(QPoint(0, 0));
QToolTip::showText(globalPos, "test", ui->downloadDirPathCopy); 
```

veloman-yunkan: There are two issues that hinder the display of the tooltip:

1. [Make sure the tooltip is diplayed on mouse release event (rather than press event)](https://forum.qt.io/topic/139220/qtooltip-showtext-doesn-t-work-in-graphicsview-click-handler/4?_=1718886680667&lang=en-US)
2. The tooltip isn't diplayed if it would contain the mouse pointer position - you can test this by clicking on different points of the copy button/label. You won't have this problem if you call `QToolTip::showText()` with the current position of the mouse.

sgourdas: Wow, this makes perfect sense. Thank you.

src/clickableqlabel.cpp
```diff
@@ -0,0 +1,14 @@
...
+ClickableQLabel::ClickableQLabel(QWidget* parent)
+    : QLabel(parent) {
+    setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/style.css"));
+}
+
+ClickableQLabel::~ClickableQLabel() {}
+
+void ClickableQLabel::mouseReleaseEvent(QMouseEvent* event) {
+    emit clicked();
+    QLabel::mousePressEvent(event);
```
veloman-yunkan, line 13: Wrong base class is called

src/settingsview.cpp
```diff
@@ -156,7 +163,17 @@ void SettingsView::setReopenTab(bool reopen)
...
+    ui->downloadDirPath->setToolTip(dir);
+}
+
+void SettingsView::onDownloadPathClicked()
+{
+    QString downloadPath = KiwixApp::instance()->getSettingsManager()->getDownloadDir();
+    QApplication::clipboard()->setText(downloadPath);
+
+    QPoint globalPos = ui->downloadDirPathCopy->mapToGlobal(QPoint(0, -ui->downloadDirPathCopy->height()));
+    QToolTip::showText(globalPos, "Path was copied", ui->downloadDirPathCopy);
```
veloman-yunkan, line 176: The tooltip text must be a translatable resource

src/clickableqlabel.h
```diff
@@ -0,0 +1,22 @@
...
+public:
+    explicit ClickableQLabel(QWidget* parent = Q_NULLPTR);
+    ~ClickableQLabel();
+
+signals:
+    void clicked();
+
+protected:
+    void mouseReleaseEvent(QMouseEvent* event);
+};
```
veloman-yunkan, line 20: I start wondering if this class is now justified. Can't you use some kind of a Qt button instead?

src/settingsview.cpp
```diff
@@ -156,7 +163,17 @@ void SettingsView::setReopenTab(bool reopen)
 
 void SettingsView::onDownloadDirChanged(const QString &dir)
 {
-    ui->downloadDirPath->setText(dir);
+    ui->downloadDirPath->setText(formatDownloadDir(dir));
+    ui->downloadDirPath->setToolTip(dir);
+}
+
+void SettingsView::onDownloadPathClicked()
```
veloman-yunkan, line 170: Rename to `copyDownloadPathToClipboard()`

veloman-yunkan (changes requested):

src/clickableqlabel.h
sgourdas: You are right. Initial target I had was different, so we can change this.

sgourdas (commented):
Thanks for the feedback, I think it is very improved now.

Just a question, should I squash after final approve or between each iteration?

src/settingsview.cpp
```diff
@@ -184,3 +200,13 @@ void SettingsView::onReopenTabChanged(bool reopen)
...
+
+QString SettingsView::formatDownloadDir(const QString& input) {
+    const int maxLength = 40;
+    if (input.length() > maxLength) {
+        QString suffix = input.right(maxLength);
+        int directoryIndex = suffix.indexOf('/');
+        return "..." + suffix.mid(directoryIndex != -1 ? directoryIndex : 0);
+    }
+    return input;
+}
```
veloman-yunkan, line 212: This doesn't need to be a member function of `SettingsView`. Please make it a free function in an unnamed namespace (and don't forget to delete its declaration from the *.h file).

resources/i18n/en.json
```diff
@@ -135,6 +135,7 @@
     "monitor-dir-dialog-title":"Are you sure you want to change the monitor directory?",
     "monitor-dir-dialog-msg":"The new monitor directory path will be:\n{{DIRECTORY}}",
     "monitor-clear-dir-dialog-title":"Are you sure you want to clear the monitor directory?",
+    "path-copy": "Path was copied",
```
veloman-yunkan, line 138: 1. "path-copy" → "path-was-copied"
2. Update qqq.json too

veloman-yunkan (changes requested):
Almost perfect! Let's eliminate the "almost" :)

veloman-yunkan:
> Just a question, should I squash after final approve or between each iteration?

The goal should be to facilitate the review. Small PRs can be rewritten in place. For large PRs fixup commits should be used so that the reviewer can easily see only the incremental changes.

src/settingsview.cpp
sgourdas: Is it ok to keep in the same file and just remove the namespace?

sgourdas:
> The goal should be to facilitate the review. Small PRs can be rewritten in place. For large PRs fixup commits should be used so that the reviewer can easily see only the incremental changes.

I mean for a PR like this one. When you ask for changes, should I squash right after making them or just wait to do it before final approval for merging? I am asking because comments are deleted after the squash and you might need to keep track of it until we finalize the PR.

src/settingsview.cpp
veloman-yunkan: Yes

veloman-yunkan:
> > The goal should be to facilitate the review. Small PRs can be rewritten in place. For large PRs fixup commits should be used so that the reviewer can easily see only the incremental changes.
> 
> I mean for a PR like this one. When you ask for changes, should I squash right after making them or just wait to do it before final approval for merging? I am asking because comments are deleted after the squash and you might need to keep track of it until we finalize the PR.

You are welcome to squash

veloman-yunkan (approved):

PR #1126, opened 2024-06-07
Window position and size is remembered

### LINKED ISSUE #837: Window size and position should be remembered
Is it possible for Kiwix to remember my last used Window size and position?
If yes, how do I achieve this?
I'm using a portable installation.

ShaopengLin:
Fix #837 

Changes:
- Stores window size and position on close.
- Restores on startup.

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -124,6 +124,7 @@ void KiwixApp::init()
     }
 
     restoreTabs();
+    restoreWindowState();
 }
 
 KiwixApp::~KiwixApp()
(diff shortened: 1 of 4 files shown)


ShaopengLin:
@kelson42 after a bit of digging I still feel the Qt's handling of this is appropriate. If possible we can have people with other Linux Window Managers test this behavior.

kelson42:
@ShaopengLin Thank you. I tested on Ubuntu (so with Gnome Window Manager) and it does the job. I'm sure we have a few edge corners. But at this stage, I believe we should focus on the most usual use cases. If someone comes now or later with an edge case, then we will treat.
@veloman-yunkan To me, good to review.

veloman-yunkan (approved):

PR #1131, opened 2024-06-15
Eliminated emit of Library class signals from other objects

### LINKED ISSUE #540: Codebase: emiting a signal forcedly from outside is a bad practice

ShaopengLin:
Fix #540

Change:
 - Created member function in Library class to emit signals
 - Other classes simply use these if they would like to emit such signals.

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -133,6 +133,8 @@ ContentManager::ContentManager(Library* library, kiwix::Downloader* downloader, 
     setCurrentCategoryFilter(getSettingsManager()->getCategoryList());
     setCurrentContentTypeFilter(getSettingsManager()->getContentType());
     connect(mp_library, &Library::booksChanged, this, [=]() {emit(this->booksChanged());});
+    connect(this, &ContentManager::libraryBooksChanged, mp_library, &Library::booksChanged);
+    connect(this, &ContentManager::libraryBookmarksChanged, mp_library, &Library::bookmarksChanged);
     connect(this, &ContentManager::filterParamsChanged, this, &ContentManager::updateLibrary);
     connect(this, &ContentManager::booksChanged, this, [=]() {
         updateModel();
@@ -568,7 +570,7 @@ void ContentManager::downloadDisappeared(QString bookId)
     bCopy.setDownloadId("");
     mp_library->getKiwixLibrary()->addOrUpdateBook(bCopy);
     mp_library->save();
-    emit(mp_library->booksChanged());
+    emit libraryBooksChanged();
 }
 
 void ContentManager::downloadCompleted(QString bookId, QString path)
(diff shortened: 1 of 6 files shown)


ShaopengLin:
Though the original issue mentions we can connect signals to signals, here I think just using a member function does the job. For this particular case, creating different signals in different classes just for this purpose feels redundant.

veloman-yunkan (commented):
I don't think that these changes are an improvement - the new member functions are no different from the signal being emitted explicitly.

ShaopengLin:
The purpose is to emit signals within the object, which I believe member function would be doing that? For the LibraryManipulator, we cannot inherit QObject so we have to rely on a direct emission or member function of some sort.

Both creating extra signals and creating member methods are hacks, but they should do the job. @asashnov Do you have an input on whether we should stick to connecting signals as much as possible or is using member function calls fine.

veloman-yunkan:
> The purpose is to emit signals within the object

But doing it via a dedicated member function doesn't make the design better. I guess the design principle is that the objects should emit signals from within otherwise meaningful methods.

ShaopengLin:
Yeah that is good overall principle, I will refactor this.

veloman-yunkan (commented):
I can't justify merging these changes as well as continuing the work on this PR. To me it looks like applying a minute amount of Dior fragrance to an athlete after extensive workout before he takes shower and changes his clothes. I believe that the mild issue being addressed should rather be fixed as a by-product of a more serious refactoring/redesign/code-cleanup.

kelson42:
> I can't justify merging these changes as well as continuing the work on this PR. To me it looks like applying a minute amount of Dior fragrance to an athlete after extensive workout before he takes shower and changes his clothes. I believe that the mild issue being addressed should rather be fixed as a by-product of a more serious refactoring/redesign/code-cleanup.

I can agree with that, but then please try to phrase the nature of the problem (and maybe a bit of the solution) within an issue. I understand that this might be a bit challenging, and if too complicated/touchy we won't give it to a student.

veloman-yunkan:
@kelson42 Let's not waste time on bureaucracy around a minor issue. Just assign it to me and I will resolve it over time as I rewrite more of `kiwix-desktop`'s source code.

PR #1132, opened 2024-06-16
Added Custom File-Not-Found page for Tabs

### LINKED ISSUE #1073: Remember current tab from previous user session

ShaopengLin:
Fix #1073
Follow-up from #1120.

Changes:
- Custom file-not-found page when tabs have Urls that point to non-existing Zim files.
- Now displays the bookmarks of the non-existing Zim files. Previously they were marked as invalid and not added.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -168,5 +168,8 @@
     "no-pictures": "No Pictures",
     "no-videos": "No Videos",
     "open-previous-tabs-at-startup": "Open previous tabs at startup",
-    "preview-book-in-web-browser": "Preview book in web browser"
+    "preview-book-in-web-browser": "Preview book in web browser",
+    "file-not-found-title": "ZIM File Not Found",
+    "file-not-found-text": "ZIM file doesn't exist or is not readable",
+    "zim-id": "ZIM Id"
 }
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -171,5 +171,7 @@
     "preview-book-in-web-browser": "Preview book in web browser",
     "file-not-found-title": "ZIM File Not Found",
     "file-not-found-text": "ZIM file doesn't exist or is not readable",
-    "zim-id": "ZIM Id"
+    "zim-id": "Zim Id",
+    "zim-name": "Zim Name",
+    "zim-path": "Zim File Path"
 }
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -906,5 +906,6 @@ void ContentManager::updateLibraryFromDir(QString monitorDir)
 
 void ContentManager::handleDisappearedZimFile(QString bookId)
 {
-    mp_library->removeBookFromLibraryById(bookId);
+    if (!KiwixApp::instance()->getTabWidget()->getTabZimIds().contains(bookId))
+        mp_library->removeBookFromLibraryById(bookId);
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -106,10 +106,10 @@ void KiwixApp::init()
         mp_manager->asyncUpdateLibraryFromDir(monitorDir);
     });
 
-    setupDirectoryMonitoring();
-
+    /* Restore Tabs before directory monitoring to ensure we know what tabs user had. */
     restoreTabs();
     restoreWindowState();
+    setupDirectoryMonitoring();
 }
 
 void KiwixApp::setupDirectoryMonitoring()
--- a/src/tabbar.h
+++ b/src/tabbar.h
@@ -50,6 +50,7 @@ public:
     void openFindInPageBar();
     void closeTabsByZimId(const QString &id);
     QStringList getTabUrls() const;
+    QStringList getTabZimIds() const;
 
 protected:
     void mousePressEvent(QMouseEvent *event);
--- a/src/urlschemehandler.cpp
+++ b/src/urlschemehandler.cpp
@@ -44,7 +44,7 @@ UrlSchemeHandler::handleContentRequest(QWebEngineUrlRequestJob *request)
     try {
       archive = library->getArchive(zim_id);
     } catch (std::out_of_range& e) {
-      request->fail(QWebEngineUrlRequestJob::UrlNotFound);
+      replyZimNotFoundPage(request, zim_id);
       return;
     }
     try {
@@ -173,6 +173,31 @@ UrlSchemeHandler::handleSearchRequest(QWebEngineUrlRequestJob* request)
     QBuffer *buffer = new QBuffer;
     buffer->setData(content.data(), content.size());
     connect(request, &QObject::destroyed, buffer, &QObject::deleteLater);
+    request->reply("text/html", buffer);
+}
+
+void
+UrlSchemeHandler::replyZimNotFoundPage(QWebEngineUrlRequestJob *request,
+                                       const QString &zimId)
+{
+    QBuffer *buffer = new QBuffer;
+    QString contentHtml = "<section><div>"
+                          "<h1>" +
+                          gt("file-not-found-title") +
+                          "</h1>"
+                          "<p>" +
+                          gt("file-not-found-text") +
+                          "</p>"
+                          "<p>" +
+                          gt("zim-id") + ": <b>" + zimId +
+                          "</b></p>"
+                          "</div></section>";
+
+    buffer->open(QIODevice::WriteOnly);
+    buffer->write(contentHtml.toStdString().c_str());
+    buffer->close();
+
+    connect(request, SIGNAL(destroyed()), buffer, SLOT(deleteLater()));
     request->reply("text/html", buffer);
 }
 
--- a/src/urlschemehandler.cpp
+++ b/src/urlschemehandler.cpp
@@ -181,6 +181,15 @@ UrlSchemeHandler::replyZimNotFoundPage(QWebEngineUrlRequestJob *request,
                                        const QString &zimId)
 {
     QBuffer *buffer = new QBuffer;
+    QString path = "N/A", name = "N/A";
+    try 
+    {
+        auto& book = KiwixApp::instance()->getLibrary()->getBookById(zimId);
+        path = QString::fromStdString(book.getPath());
+        name = QString::fromStdString(book.getName());
+    }
+    catch (...) { /* Blank */ }
+
     QString contentHtml = "<section><div>"
                           "<h1>" +
                           gt("file-not-found-title") +
@@ -190,6 +199,12 @@ UrlSchemeHandler::replyZimNotFoundPage(QWebEngineUrlRequestJob *request,
                           "</p>"
                           "<p>" +
                           gt("zim-id") + ": <b>" + zimId +
+                          "</b></p>"
+                          "<p>" +
+                          gt("zim-name") + ": <b>" + name +
+                          "</b></p>"
+                          "<p>" +
+                          gt("zim-path") + ": <b>" + path +
                           "</b></p>"
                           "</div></section>";
 
--- a/resources/i18n/qqq.json
+++ b/resources/i18n/qqq.json
@@ -178,5 +178,7 @@
 	"path-was-copied": "Tooltip confirming that the download path from settings was copied.",
 	"file-not-found-title": "Error title text displayed when the desktop application cannot find the Zim file needed to display the web page.",
 	"file-not-found-text": "Error description text for when the desktop application cannot find the Zim file needed to display the web page.",
-	"zim-id": "The term for the unique identifier of a zim file."
+	"zim-id": "The term for the unique identifier of a zim file.",
+	"zim-name": "The term for the name of a Zim file",
+	"zim-path": "The term for the path of a Zim file"
 }
--- a/resources/i18n/qqq.json
+++ b/resources/i18n/qqq.json
@@ -175,5 +175,8 @@
 	"no-details": "A content type for Zim files representing it only has an introduction.",
 	"no-pictures": "A content type for Zim files that does not contain pictures.",
 	"no-videos": "A content type for Zim files that does not contain videos.",
-	"path-was-copied": "Tooltip confirming that the download path from settings was copied."
+	"path-was-copied": "Tooltip confirming that the download path from settings was copied.",
+	"file-not-found-title": "Error title text displayed when the desktop application cannot find the Zim file needed to display the web page.",
+	"file-not-found-text": "Error description text for when the desktop application cannot find the Zim file needed to display the web page.",
+	"zim-id": "The term for the unique identifier of a zim file."
 }
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -284,6 +284,15 @@ QStringList TabBar::getTabUrls() const {
     return idList;
 }
 
+QStringList TabBar::getTabZimIds() const
+{ 
+    QStringList idList;
+    for (int index = 0; index <= mp_stackedWidget->count(); index++)
+        if (ZimView* zv = qobject_cast<ZimView*>(mp_stackedWidget->widget(index)))
+            idList.push_back(zv->getWebView()->zimId());
+    return idList;
+}
+
 void TabBar::closeTab(int index)
 {
     // The first and last tabs (i.e. the library tab and the + (new tab) button)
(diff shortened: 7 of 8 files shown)


ShaopengLin:
Let's review this after #1120 is merged. The reopening functionality will be available after we rebase this.

src/readinglistbar.cpp
```diff
@@ -41,12 +41,16 @@ void ReadingListBar::setupList()
...
-    for(auto& bookmark:bookmarks) {
+    for(unsigned i = 0; i < bookmarks.size(); i++) {
+        auto& bookmark = bookmarks[i];
         std::shared_ptr<zim::Archive> archive;
         const QString& title = QString::fromStdString(bookmark.getTitle());
         try {
             archive = library->getArchive(QString::fromStdString(bookmark.getBookId()));
         } catch (std::out_of_range& e) {
+            /* There can be bookmarks whose information is lost. */
+            if (!bookmark.getBookId().empty())
```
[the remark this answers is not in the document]
kelson42: no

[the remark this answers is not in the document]
ShaopengLin: hmm, somehow I have a bunch of entries that don't have zimIds. It might be a bug but I am not sure how to reproduce.

src/readinglistbar.cpp
```diff
@@ -97,16 +101,17 @@ void ReadingListBar::onItemActivated(QListWidgetItem* item, Qt::MouseButtons but
...
     item->setTextAlignment(Qt::TextWordWrap);
+    item->setData(Qt::UserRole, idx);
 }
 
 void ReadingListBar::openUrl(QListWidgetItem* item, bool newTab)
 {
-    int index = ui->listWidget->row(item);
-    auto bookmark = KiwixApp::instance()->getLibrary()->getBookmarks(true).at(index);
+    int index = item->data(Qt::UserRole).toUInt();
+    auto bookmark = KiwixApp::instance()->getLibrary()->getBookmarks(false).at(index);
```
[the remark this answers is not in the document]
ShaopengLin: No, we were only adding bookmarks that are valid in setupList. The getBookmarks(true) also only looks at valid bookmarks. The order were the same so we didn't have a bug before.

ShaopengLin:
> Why were the changes related to a more informative File-Not-Found error page (reporting the ZIM path) lost?

@veloman-yunkan sorry I am not exactly sure what this means? I thought I removed everything about zim path and only did a custom error page.

veloman-yunkan:
> > Why were the changes related to a more informative File-Not-Found error page (reporting the ZIM path) lost?
> 
> @veloman-yunkan sorry I am not exactly sure what this means? I thought I removed everything about zim path and only did a custom error page.

I thought that this PR should contain all changes that were removed from #1120 when we [decided to split it into two parts](https://github.com/kiwix/kiwix-desktop/pull/1120#issuecomment-2171427504).

ShaopengLin:
I was looking at kelson's comment and thought that you meant we are going go for that. Should I still add the path? https://github.com/kiwix/kiwix-desktop/pull/1120#issuecomment-2169248914

veloman-yunkan:
> If we were to add the missing paths we will have to go back to that original complicated solution.

And that is exactly why I chose to split the PR into two parts - so that the complicated solution doesn't block merging the fix for the original issue while we keep working around problems that crop up while we pursue a secondary improvement.

src/readinglistbar.cpp
```diff
@@ -56,16 +56,9 @@ void ReadingListBar::setupList()
...
-                listWidget);
-            item->setTextAlignment(Qt::TextWordWrap);
+            addItem(title, icon);
         } catch (zim::EntryNotFound& e) {
-            auto item = new QListWidgetItem(
-                QString::fromStdString(bookmark.getTitle()),
-                listWidget);
-            item->setTextAlignment(Qt::TextWordWrap);
+            addItem(title);
         }
```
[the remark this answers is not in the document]
ShaopengLin: Done

resources/i18n/en.json
```diff
@@ -167,5 +167,8 @@
     "no-pictures": "No Pictures",
     "no-videos": "No Videos",
     "open-previous-tabs-at-startup": "Open previous tabs at startup",
-    "preview-book-in-web-browser": "Preview book in web browser"
+    "preview-book-in-web-browser": "Preview book in web browser",
+    "file-not-found-title": "Zim File Not Found",
+    "file-not-found-text": "We can't find the Zim file corresponding to this page. Please check that your file exists.",
+    "zim-url": "Zim Url"
```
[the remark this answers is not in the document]
ShaopengLin: Done

resources/i18n/en.json
```diff
@@ -167,5 +167,8 @@
     "no-pictures": "No Pictures",
     "no-videos": "No Videos",
     "open-previous-tabs-at-startup": "Open previous tabs at startup",
-    "preview-book-in-web-browser": "Preview book in web browser"
+    "preview-book-in-web-browser": "Preview book in web browser",
+    "file-not-found-title": "Zim File Not Found",
+    "file-not-found-text": "We can't find the Zim file corresponding to this page. Please check that your file exists.",
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/readinglistbar.cpp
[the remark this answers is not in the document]
veloman-yunkan: Let's try **not** to hide any invalid bookmarks. But we can display them in a special way.

Note that if you add all bookmarks to the reading list, you won't need the extra `idx` parameter in `ReadingListBar::addItem()`

src/tabbar.h
```diff
@@ -10,6 +10,24 @@
...
+
+    BookInfo() {};
+    BookInfo(const QString &name, const QString &path)
+        : name{name}, path{path} {};
+    BookInfo(const QJsonObject &json)
+        : name{json.value("name").toString()},
+          path{json.value("path").toString()} {};
+
+    QJsonObject toJson() const { return {{"name", name}, {"path", path}}; };
+};
```
veloman-yunkan, line 28: When you were referring to the complicated solution I thought that you meant the last version (that had some issues because of directory monitoring) before we decided to split the PR. Sorry.

veloman-yunkan (changes requested):
I think that we should again split this PR into two:

1. Improvement of the missing ZIM file error page. The main challenge with this one will be to resolve the issues that we faced just before the previous split (related to the side effects of directory monitoring).
2. Listing all bookmarks (including invalid ones)

src/tabbar.h
ShaopengLin: Since we cannot change the behavior of directory monitoring, we can:
1. simply save all the books that we have seen in updateFromDir, no matter if we will be removing them.
2. only remove a saved book when all traces such as opened tab and bookmarks of that book is removed.

ShaopengLin:
@veloman-yunkan This solution is a lot simpler than the last one. If you have any other suggestions I am open to try.

src/kiwixapp.cpp
```diff
@@ -577,3 +592,35 @@ void KiwixApp::saveCurrentTabIndex()
...
+    auto bookInfoMap = mp_session->value("removedZimBookInfo", QVariantMap{}).toMap();
+    auto existingZimIds = getTabWidget()->getTabZimIds();
+
+    for (const auto& zimId : bookInfoMap.keys())
+    {
+        if (!existingZimIds.contains(zimId))
+            bookInfoMap.remove(zimId);
+    }
+    mp_session->setValue("removedZimBookInfo", bookInfoMap);
+}
```
veloman-yunkan, line 626: To me something seems wrong about saving this information in the session. Please give me some time to come up with a better solution or admit that none exists.

kelson42:
@ShaopengLin Any feedback?

ShaopengLin:
> To me something seems wrong about saving this information in the session. Please give me some time to come up with a better solution or admit that none exists.

We are a little stuck on how to properly design the solution to this problem completely without changing the behaviour of the directory monitoring. I believe veloman is busy with the other issue and hasn't touched on this yet.

@veloman-yunkan @kelson42  This has been stuck for a while. We don't have an easy path to save those information otherwise, since the directory monitoring completely removes the record of zims if it disappears.

I believe a clean solution would likely need us to either create a new file like library.xml to retain the missing zim entries, or change libkiwix to save this information.

ShaopengLin:
@veloman-yunkan This has been stuck for too long. If storing this information in settings or a file doesn't make sense, we can store it in the URL paths. What do you think?

Otherwise, we should simply do what Kelson said, have a simple error message with Zim Id, and then move on to have the bookmark portion finished. Having the bookmark feature I believe is an equally important user experience.

kelson42:
@ShaopengLin I would like to intervene here as things don't move well and maybe not even in the right direction.

First of all, to me, all the work around bookmarks in that PR is unclear: what is tried to be achieved and what is the chosen approach. One reason is that there is no proper issue describing the problem and therefore no formal validation about both problem and solution approach.

Regarding the topic of "invalid" bookmarks, to me this is not a problem because:
* Bookmarks refering to a ZIM which is not in the local library should never be displayed
* All the bookmark handling should be done in the libkiwix (not in Kiwix Desktop)

For all these reasons, please:
* Keep only the first part of your PR here (about missing page error)
* Open an issue describing clearly the problem around the bookmark so we can discuss it independently of any PR

veloman-yunkan:
> @veloman-yunkan This has been stuck for too long. If storing this information in settings or a file doesn't make sense, we can store it in the URL paths. What do you think?

@ShaopengLin I still think that the approach outlined in https://github.com/kiwix/kiwix-desktop/pull/1120#issuecomment-2162262938 is the best. A proper implementation of it would require moving the directory monitoring functionality from `Library` to `ContentManager` - the former doesn't have any knowledge of a book being opened in a tab while the latter does. Then `ContentManager` will be in charge of deciding how to handle a disappeared ZIM file - if it is of special interest to the user (e.g. was open in a tab, was bookmarked, etc) then it will be preserved in the library but marked as missing, otherwise it will be simply dropped from the library.

If you think it's too complex or get stuck while pursuing that solution please let me know and I will implement it.

ShaopengLin:
@veloman-yunkan Kelson has mentioned this before:
> There are many valid reasons why the directory monitoring can remove a file from the library and all of them are external to Kiwix. We have no control on this. The rest must follow.

Doesn't this violate our solution to keep the book in library.xml? This was the reason I was not going with the solution from [#1120 (comment)](https://github.com/kiwix/kiwix-desktop/pull/1120#issuecomment-2162262938) for this PR in the first place.

veloman-yunkan:
> I propose that I implement the said change in library monitoring as a separate PR whereupon you can rebase and test your PR on top of my branch.

Done in #1149. You will have to enhance the function `ContentManager::handleDisappearedZimFile()` as described in https://github.com/kiwix/kiwix-desktop/pull/1132#issuecomment-2238540879.

ShaopengLin:
@veloman-yunkan I have rebased and tested your change. Do you mean I simply push what I have (my change rebased onto yours.) to this PR?

veloman-yunkan:
> @veloman-yunkan I have rebased and tested your change. Do you mean I simply push what I have (my change rebased onto yours.) to this PR?

Yes (and change the base branch to `directory_monitoring_improvements` by editing the PR so that only your changes show up here).

src/contentmanager.cpp
```diff
@@ -906,5 +906,6 @@ void ContentManager::updateLibraryFromDir(QString monitorDir)
 
 void ContentManager::handleDisappearedZimFile(QString bookId)
 {
-    mp_library->removeBookFromLibraryById(bookId);
+    if (!KiwixApp::instance()->getTabWidget()->getTabZimIds().contains(bookId))
+        mp_library->removeBookFromLibraryById(bookId);
```
ShaopengLin, line 910: @veloman-yunkan I checked that the book's isPathValid boolean was already set when we retrieved it. Don't think we need to make any modifications to the library.

veloman-yunkan (changes requested):
It works! Nice!

Please separate the change addressing the subtlety with directory monitoring into a separate commit. Then this PR will be complete.

ShaopengLin:
> Please separate the change addressing the subtlety with directory monitoring into a separate commit. Then this PR will be complete.

I assume you meant the setupDirectoryMonitoring subtlety

veloman-yunkan:
> > Please separate the change addressing the subtlety with directory monitoring into a separate commit. Then this PR will be complete.
> 
> I assume you meant the setupDirectoryMonitoring subtlety

No, I meant everything related to directory monitoring. Suppose that you implement the more informative file-not-found error page with directory monitoring disabled. Then you find out that turning directory monitoring on leads to certain issues. I want those issues to be addressed in a separate commit (with a proper explanation).

ShaopengLin:
@veloman-yunkan 
There are two situations where external modification to the library due to directory monitoring is problematic:

1. If the library file(s) are removed or modified, our code for file-not-found page can fault due to the library losing book information. This is, however already mitigated with the try/catch when we are unable to find the book.

2. Existing functions may rely on the previous monitoring behaviour to validate whether the zim file physically exist. The getValidBookById resolves this issue. This function retains the previous behaviour of getBookById, and should be used if we want to get a book that is physically present. Could you do a sanity check on the getBookById calls I left out or replaced to make sure I didn't make any misconception of the intentions of getBookById? Thanks!

veloman-yunkan (commented):
> @veloman-yunkan There are two situations where external modification to the library due to directory monitoring is problematic:
> 
>  1. If the library file(s) are removed or modified, our code for file-not-found page can fault due to the library losing book information. This is, however already mitigated with the try/catch when we are unable to find the book.
> 
>  2. Existing functions may rely on the previous monitoring behaviour to validate whether the zim file physically exist. The getValidBookById resolves this issue. This function retains the previous behaviour of getBookById, and should be used if we want to get a book that is physically present. Could you do a sanity check on the getBookById calls I left out or replaced to make sure I didn't make any misconception of the intentions of getBookById? Thanks!

Let's not let this PR's scope drift. The purpose of this PR is to improve the information displayed in the file-not-found error page.

Existing functionality does *not* rely on previous behaviour of directory monitoring. Directory monitoring was added relatively recently and not all of its side effects were fully understood and addressed. Let's not go after potential issues that have been introduced due to directory monitoring in the past. Just make sure that your enhancement

1. works in the absence of directory monitoring in the first place and
2. also works when directory monitoring is enabled (changes targeting this part should come in a separate commit).

kelson42:
#1149 has been merged, so if I get it right, after a rebase it should be possible tom omplete this PR!?

src/tabbar.h
```diff
@@ -50,6 +50,7 @@ class TabBar : public QTabBar
     void openFindInPageBar();
     void closeTabsByZimId(const QString &id);
     QStringList getTabUrls() const;
+    QStringList getTabZimIds() const;
```
veloman-yunkan, line 53: This function should be introduced in the next commit.

src/urlschemehandler.cpp
```diff
@@ -181,9 +181,15 @@ UrlSchemeHandler::replyZimNotFoundPage(QWebEngineUrlRequestJob *request,
...
-    QString name = QString::fromStdString(book.getName());
+    QString path = "N/A", name = "N/A";
+    try 
+    {
+        auto& book = KiwixApp::instance()->getLibrary()->getBookById(zimId);
+        path = QString::fromStdString(book.getPath());
+        name = QString::fromStdString(book.getName());
+    }
+    catch (...) { /* Blank */ }
+
```
veloman-yunkan, line 192: I think that this change is unrelated to directory monitoring - it can be merged into the previous commit.

veloman-yunkan (changes requested):
Almost there. Please fix the last two comments and also format your commit messages to meet the [commit message guidelines](https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting)

src/tabbar.h
ShaopengLin: Done

src/urlschemehandler.cpp
ShaopengLin: Done

veloman-yunkan (approved):

PR #1134, opened 2024-06-20
Implemented Save  Page

### LINKED ISSUE #77: Implement "Save page"

ShaopengLin:
Fix #77 
Changes:
- Re-enabled file menu item and shortcut Ctrl + Shift + S for save page.
- Saves the HTML of the page on the current tab to PDF.
- Saves non-HTML pages directly as files.
- Remembers the last directory saved when opening file selection dialoge.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -181,5 +181,6 @@
     "export-reading-list-error": "An error has occured during export of the reading list.",
     "import-reading-list": "Import reading list",
     "import-reading-list-error": "An error has occured during import of the reading list.",
-    "disable-sandbox": "Application was launched from a network drive. This is known to cause compatibility issues due to the sandbox. Do you want to take the risks and disable it?"
+    "disable-sandbox": "Application was launched from a network drive. This is known to cause compatibility issues due to the sandbox. Do you want to take the risks and disable it?",
+    "save-page-as": "Save As..."
 }
--- a/resources/i18n/qqq.json
+++ b/resources/i18n/qqq.json
@@ -188,5 +188,6 @@
 	"export-reading-list": "Represents the action of exporting the reading list to a file.",
 	"export-reading-list-error": "Error description text for when exporting the reading list to a file failed.",
 	"import-reading-list": "Represents the action of importing a reading list from a file.",
-	"import-reading-list-error": "Error description text for when importing a reading list from a file failed."
+	"import-reading-list-error": "Error description text for when importing a reading list from a file failed.",
+	"save-page-as": "Represents the action of saving the current tab content to a file chosen by the user."
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -19,6 +19,8 @@
 #include <QtPlatformHeaders\QWindowsWindowFunctions>
 #endif
 
+const QString DEFAULT_SAVE_DIR = QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation);
+
 ////////////////////////////////////////////////////////////////////////////////
 // KiwixApp
 ////////////////////////////////////////////////////////////////////////////////
@@ -564,4 +566,16 @@ void KiwixApp::restoreWindowState()
 void KiwixApp::saveCurrentTabIndex()
 {
   return mp_session->setValue("currentTabIndex", getTabWidget()->currentIndex());
+}
+
+void KiwixApp::savePrevSaveDir(const QString &prevSaveDir)
+{
+  mp_session->setValue("prevSaveDir", prevSaveDir);
+}
+
+QString KiwixApp::getPrevSaveDir() const
+{
+  QString prevSaveDir = mp_session->value("prevSaveDir", DEFAULT_SAVE_DIR).toString();
+  QDir dir(prevSaveDir);
+  return dir.exists() ? prevSaveDir : DEFAULT_SAVE_DIR;
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -427,11 +427,7 @@ void KiwixApp::createActions()
     CREATE_ACTION(OpenRecentAction, gt("open-recent"));
     HIDE_ACTION(OpenRecentAction);
 
-    /* TODO See https://github.com/kiwix/kiwix-desktop/issues/77
-    CREATE_ACTION(SavePageAsAction, tr("Save page as ..."));
-    // SET_SHORTCUT(SavePageAsAction, QKeySequence::SaveAs);
-    HIDE_ACTION(SavePageAsAction);
-    */
+    CREATE_ACTION_SHORTCUT(SavePageAsAction, gt("save-page-as"), QKeySequence::Save);
 
     CREATE_ACTION_SHORTCUTS(SearchArticleAction, gt("search-article"), QList<QKeySequence>({QKeySequence(Qt::Key_F6), QKeySequence(Qt::CTRL | Qt::Key_L), QKeySequence(Qt::ALT | Qt::Key_D)}));
 
--- a/src/kprofile.cpp
+++ b/src/kprofile.cpp
@@ -24,15 +24,29 @@ void KProfile::startDownload(QWebEngineDownloadItem* download)
 void KProfile::startDownload(QWebEngineDownloadRequest* download)
 #endif
 {
-    QString defaultFileName = download->url().fileName();
-    QString fileName = QFileDialog::getSaveFileName(KiwixApp::instance()->getMainWindow(),
-                                                       gt("save-file-as-window-title"), defaultFileName);
+    auto app = KiwixApp::instance();
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
+    QString defaultFileName = QUrl(download->path()).fileName();
+#else
+    QString defaultFileName = download->downloadFileName();
+#endif
+    QString extension = defaultFileName.section('.', -1);
+    QString filter = extension != '.' ? "(*" + extension + ")" : "";
+
+    QString fileName = QFileDialog::getSaveFileName(
+            app->getMainWindow(), gt("save-file-as-window-title"),
+            defaultFileName, filter);
+
     if (fileName.isEmpty()) {
         return;
     }
-    QString extension = "." + download->url().url().section('.', -1);
     if (!fileName.endsWith(extension)) {
         fileName.append(extension);
+    }
+
+    if (download->isSavePageDownload()) {
+        download->page()->printToPdf(fileName);
+        return;
     }
 #if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
     download->setPath(fileName);
--- a/src/kprofile.cpp
+++ b/src/kprofile.cpp
@@ -30,12 +30,13 @@ void KProfile::startDownload(QWebEngineDownloadRequest* download)
 #else
     QString defaultFileName = download->downloadFileName();
 #endif
+    QString suggestedPath = app->getPrevSaveDir() + "/" + defaultFileName;
     QString extension = defaultFileName.section('.', -1);
     QString filter = extension != '.' ? "(*" + extension + ")" : "";
 
     QString fileName = QFileDialog::getSaveFileName(
             app->getMainWindow(), gt("save-file-as-window-title"),
-            defaultFileName, filter);
+            suggestedPath, filter);
 
     if (fileName.isEmpty()) {
         return;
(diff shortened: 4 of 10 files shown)


src/webview.cpp
```diff
@@ -131,6 +132,42 @@ QMenu* WebView::getHistoryForwardMenu() const
...
+        auto mimeType = QByteArray::fromStdString(item.getMimetype());
+        mimeType = mimeType.split(';')[0];
+        
+        auto suggestedDir = app->getSettingsManager()->getPrevSaveDir();
+        auto suggestedPath = suggestedDir + "/" + this->url().fileName() + ".pdf";
+        if (mimeType == "text/html")
+        {
+            QString fileName = QFileDialog::getSaveFileName(
+                app->getMainWindow(), gt("save-file-as-window-title"),
+                suggestedPath, tr("PDF files (*.pdf)"));
```
ShaopengLin, line 158: Do we really need a `gt` translation here? The filter seems like it should be a pretty standard translation.

src/webview.cpp
```diff
@@ -131,6 +132,42 @@ QMenu* WebView::getHistoryForwardMenu() const
...
+        auto mimeType = QByteArray::fromStdString(item.getMimetype());
+        mimeType = mimeType.split(';')[0];
+        
+        auto suggestedDir = app->getSettingsManager()->getPrevSaveDir();
+        auto suggestedPath = suggestedDir + "/" + this->url().fileName() + ".pdf";
+        if (mimeType == "text/html")
+        {
+            QString fileName = QFileDialog::getSaveFileName(
+                app->getMainWindow(), gt("save-file-as-window-title"),
+                suggestedPath, gt("pdf-files-filter") + " (*.pdf)");
```
ShaopengLin, line 158: Is the `gt` for the filter text needed? I didn't see one for our .zim filters but I added it just in case.

ShaopengLin:
Actually, we should hold off until #1132 is merged as we have to test how that works. I am also seeing some places of improvement for this PR.

kelson42:
@ShaopengLin I will review this PR, can you please rebase and fix the conflict?

ShaopengLin:
@kelson42 done

kelson42:
* Why `CTRL + shift + S`? usual (Web Browser) shortcut is `CTRL + S`, at least on linux. (otherwise it works)
*

ShaopengLin:
> If I want to save the article "AC/DC" then it tries to create DC.pdf

@kelson42 I may be wrong I believe your case happens like this: The native/Qt window manager would treat it as if AC is a directory and DC.pdf is the name. The path will be path/AC/DC.pdf which simply fails to save. For this case, there is no way for us to know if the user intended for it to be a path element or a file name when the results get back to us. I don't think we have a choice here but just report an error.

In most cases, native Linux window manager simply doesn't allow this to happen. I don't believe windows' window manager will allow saves to invalid file names either.

ShaopengLin:
@kelson42 Done.

> Saving a mp4 video directly does not work, it proposes to save a PDF file, not directly the .mp4 file.

The video/contents that can be saved are these ones opened directly as a video file in the tab:
Screenshot from 2024-07-15 01-14-42

The ones embedded inside a html web page would we would not be able to identify those as the page mime type is html:
Screenshot from 2024-07-15 01-15-02

These videos can just be downloaded using the provided download button:
Screenshot from 2024-07-15 01-15-11

kelson42:
@ShaopengLin  Please update me on each point I have given. In general this is verz imoortant to be verz clear about what is done and why. The reviewer have a limited amount of time and therefore the goal is to implement PR so the reviewer as just to approve it without having to ask questions and obviouslz ask for changes.

ShaopengLin:
@kelson42
1. The shortcut for save is now Ctrl + S. Qt's "Save As" shortcut Ctrl + Shift + S assumes a desktop environment instead of a browser one.
2. The issue is explained in my previous [comment](https://github.com/kiwix/kiwix-desktop/pull/1134#issuecomment-2227718729).  
3. We now do a simple sanitization on the appearances of reserved characters of the OS in the file name by replacing them with '_'. This prevents our code from parsing a file name "AC/DC" as file "DC" in the directory "AC". All we care about here is preventing any side effects from using the title text as file name, as it should just be a placeholder name. Determining the file name validity is the job of the OS file window manager.

src/webview.cpp
kelson42: To me it seems mandatory, otherwise how would you get a translated message?! Or I get it wrong?! @veloman-yunkan

kelson42:
I have rebased as we are approaching the time to make the code review and the PR is already quite old.

kelson42:
@veloman-yunkan @ShaopengLin I would recommend to move the string slugification method to libkiwix `stringTools.cpp`. What do you think?

kelson42:
@ShaopengLin Looks good to me finally. Could you please add as well the the contextual menu like on a Web Browser?
@veloman-yunkan Ready to code review.

kelson42 (commented):
LGTM

ShaopengLin:
I would be fine with moving it to libkiwix. Note this doesn't guarantee the file name is 100% usable in Windows. They have a lot of rules extra rules.

ShaopengLin:
@kelson42 Do you want me to just PR from fork to libkiwix, or create a branch like we do here? If the latter I will need to be a collaborator.

kelson42:
> @kelson42 Do you want me to just PR from fork to libkiwix, or create a branch like we do here? If the latter I will need to be a collaborator.

Sorry, you should just have received an invitation to be member of the team and have write permission.

kelson42:
Blocked by https://github.com/kiwix/libkiwix/pull/1105

ShaopengLin:
@kelson42 @veloman-yunkan Ready for review. Updated change from https://github.com/kiwix/libkiwix/pull/1105.

kelson42:
@ShaopengLin I want to test it, but would be very nice if you could rebase it (once again).

ShaopengLin:
@kelson42 Done.

veloman-yunkan (changes requested):

src/kiwixapp.cpp
```diff
@@ -211,6 +213,11 @@ void KiwixApp::restoreTabs()
     getTabWidget()->setCurrentIndex(mp_session->value("currentTabIndex", 0).toInt());
 }
 
+void KiwixApp::restorePrevSaveDir()
+{
+  m_prevSaveDir = mp_session->value("prevSaveDir", DEFAULT_SAVE_DIR).toString();
+}
```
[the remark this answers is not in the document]
ShaopengLin: Forgot to include it when doing fixes. @veloman-yunkan if we do not care about the overhead of reading this directory from session for each `getPrevSaveDir()`, then we can merge them.

src/library.cpp
```diff
@@ -67,6 +68,22 @@ std::shared_ptr<zim::Archive> Library::getArchive(const QString &zimId)
...
+
+    try {
+        return archive.getEntryByPath(path);
+    } catch (zim::EntryNotFound& e) {
+        if (path.empty() || path == "/") {
+            return archive.getMainEntry();
+        }
+    }
+    throw zim::EntryNotFound("Cannot find entry for non empty path");
+}
```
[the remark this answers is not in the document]
ShaopengLin: @veloman-yunkan I think making this a static member makes more sense. Having getZimIdFromUrl is helpful, but it shouldn't belong to `Library`.

src/kiwixapp.cpp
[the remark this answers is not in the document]
veloman-yunkan: I see no reason why we should mind about that overhead

src/library.cpp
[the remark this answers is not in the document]
ShaopengLin: Yeah, it does seem this doesn't belong to Library either. Might be better suited as a free function in library.h/cpp

resources/i18n/en.json
```diff
@@ -181,5 +181,6 @@
     "export-reading-list-error": "An error has occured during export of the reading list.",
     "import-reading-list": "Import reading list",
     "import-reading-list-error": "An error has occured during import of the reading list.",
-    "disable-sandbox": "Application was launched from a network drive. This is known to cause compatibility issues due to the sandbox. Do you want to take the risks and disable it?"
+    "disable-sandbox": "Application was launched from a network drive. This is known to cause compatibility issues due to the sandbox. Do you want to take the risks and disable it?",
+    "save-as": "Save As"
```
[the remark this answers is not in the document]
ShaopengLin: Don't think the text should change as Chromium browsers also just say "Save As..."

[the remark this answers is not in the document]
ShaopengLin: Done

src/kiwixapp.cpp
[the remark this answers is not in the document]

src/library.cpp
```diff
@@ -3,10 +3,27 @@
...
+
```
veloman-yunkan, line 26: I think you can preserve this function in `src/urlschemehandler.cpp` as an enhancement of `getEntryFromPath()` (this way the change will be more conspicuous).

src/library.h
```diff
@@ -61,4 +61,6 @@ public slots:
 friend class LibraryManipulator;
 };
 
+zim::Entry getArchiveEntryFromUrl(const zim::Archive& archive, const QUrl& url);
+
```
veloman-yunkan, line 65: Declaring functions in header files is convenient but not mandatory. You can simply forward declare it in the second .cpp file (accompanied by a comment pointing to the location of the definition) and do that in the commit when that second usage is introduced. This way the issue of finding an appropriate header file for this declaration (and the ensuing sense of necessity of moving the definition in the respective cpp file) is worked around.

resources/i18n/qqq.json
```diff
@@ -188,5 +188,6 @@
 	"export-reading-list": "Represents the action of exporting the reading list to a file.",
 	"export-reading-list-error": "Error description text for when exporting the reading list to a file failed.",
 	"import-reading-list": "Represents the action of importing a reading list from a file.",
-	"import-reading-list-error": "Error description text for when importing a reading list from a file failed."
+	"import-reading-list-error": "Error description text for when importing a reading list from a file failed.",
+	"save-pagr-as": "Represents the action of saving the current tab content to a file chosen by the user."
```
veloman-yunkan, line 192: pagr → you know what

src/kprofile.cpp
```diff
@@ -25,20 +25,26 @@ void KProfile::startDownload(QWebEngineDownloadRequest* download)
 #endif
 {
     auto app = KiwixApp::instance();
-    QString defaultFileName = download->url().fileName();
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
+    QString defaultFileName = QUrl(download->path()).fileName();
+#else
+    QString defaultFileName = download->downloadFileName();
+#endif
```
veloman-yunkan: This better be done in the previous (or separate) commit

veloman-yunkan (changes requested):
If it were not for the pagr typo I might have approved. But since we are going to go through another iteration I chose not to silence the perfectionist side of me :)

src/kprofile.cpp
ShaopengLin: Done

resources/i18n/qqq.json
ShaopengLin: Done

src/library.h
ShaopengLin: Done

src/library.cpp
ShaopengLin: Done

src/urlschemehandler.cpp
```diff
@@ -13,30 +13,30 @@
...
 {
-  try {
-    return archive.getEntryByPath(path);
-  } catch (zim::EntryNotFound& e) {
-    if (path.empty() || path == "/") {
-      return archive.getMainEntry();
-    }
-  }
-  throw zim::EntryNotFound("Cannot find entry for non empty path");
 }
```
veloman-yunkan, line 34: The placement of `getArchiveEntryFromUrl()` and the change in whitespace make the diff in this commit more confusing than it could be. History of code is as important as its most recent state. Please don't introduce unnecessary difficulties in understanding the changes in the code. Cleanup actions like moving code around for better cohesion etc are welcome but they should be performed as separate steps in commits of their own. A general rule - try to not combine changes of different nature in the same commit and try keep the size of each change within an average software engineer's ability to comprehend it in a single pass.

veloman-yunkan (changes requested):
The commit that dropped category translations was lost. Was that a mistake?

src/urlschemehandler.cpp
ShaopengLin: Done, the change log does look confusing...

veloman-yunkan (approved):

PR #1138, opened 2024-06-27
Added home button

### LINKED ISSUE #58: Where is the "home" button
I'm missing the "Home" button which brings me to the welcome page of a ZIM. It seems to be absent from the design too! @schoenbaechler What that on purpose?

sgourdas:
This aims to close #58 

Added the home button beside the search bar, that works using the pre existing logic tied to `alt+home`


kelson42:
@sgourdas For me it's OK, on the usabilty side, but please make the logo a bit bigger, compare with the other logo, I guess a few pixels are missing

sgourdas:
> There is also a problem with the color of the line (same for the random button): you have but black, but if you look to the icon on the right, they are different, please fix.

@kelson42 sorry but this is not clear to me. Do you mean that the left icons (home and random) are bolder than the right ones?

sgourdas:
Checking the `.svg` of each file respectively they all use `#000` for color which is black. For me, the only visible difference is the boldness. Should I check something different?

kelson42:
@sgourdas It's not the same coloring, please see 
image

sgourdas:
@kelson42 I got what you mean. I did some improvements to get to the look intended. 

Regarding the **random** icon, I cannot get it to be thinner than this. An alternative I have ready, if you prefer, would be using this one below:
image

kelson42:
@sgourdas At a first sight, it looks indeed better but Iwill review in detaila coloring, boldness and sizing. Hope all good.

kelson42:
@veloman-yunkan Can you please validate the code?

kelson42:
@sgourdas Like other buttons related to the content, the button should be deactivated if we are in the library

sgourdas:
@kelson42 added the disabling functionality and reduced some instance calls also in the meantime.

kelson42 (approved):
LGTM from a user perspective, @veloman-yunkan ready for code review

veloman-yunkan (changes requested):
LGTM, but please split your changes into two commits:

1. Code cleanup
2. Implementation of the new feature

veloman-yunkan (commented):
The title of the second commit "Code cleanup and random.svg color alignment" suggests that two unrelated changes are combined into one.

src/tabbar.cpp
```diff
@@ -66,6 +62,12 @@ TabBar::TabBar(QWidget *parent) :
             this, SLOT(onTabMoved(int,int)), Qt::DirectConnection);
 }
 
+void TabBar::openHomePage() {
+    auto current = this->currentWebView();
+    QUITIFNULL(current);
+    current->setUrl("zim://" + current->zimId() + ".zim/");
+}
```
veloman-yunkan, line 69: This is rather a refactoring change. Maybe it better belong to the code cleanup commit.

sgourdas:
> The title of the second commit "Code cleanup and random.svg color alignment" suggests that two unrelated changes are combined into one.

Should i split that also in 2? `random.svg` is being changed in that to allow for better consistency with the rest toolbar icons.

veloman-yunkan (commented):
I'd prefer if code cleanup precedes the enhancement (since in general that is how enhancements should be done - while studying the code, you discover ways in which it could be improved, so that the enhancement can benefit from those changes too).

veloman-yunkan:
> > The title of the second commit "Code cleanup and random.svg color alignment" suggests that two unrelated changes are combined into one.
> 
> Should i split that also in 2? `random.svg` is being changed in that to allow for better consistency with the rest toolbar icons.

Yes. Changes should be atomic. I would even recommend having two separate commits (extracting a piece of code into a named function; shortening lines with the help of auxiliary local variables) for the refactoring/cleanup part of this PR

src/tabbar.cpp
sgourdas: Agree

sgourdas: BTW, do you think this change is beneficial since `openHomePage` is only called in one place?

veloman-yunkan: Of course, it is. Now now that place is less cluttered.

sgourdas:
@veloman-yunkan is this ok?

src/kiwixapp.cpp
```diff
@@ -527,6 +527,7 @@ void KiwixApp::handleItemsState(TabType tabType)
     KiwixApp::instance()->getAction(KiwixApp::ZoomOutAction)->setDisabled(libraryOrSettingsTab);
     KiwixApp::instance()->getAction(KiwixApp::ZoomResetAction)->setDisabled(libraryOrSettingsTab);
     KiwixApp::instance()->getAction(KiwixApp::RandomArticleAction)->setDisabled(libraryOrSettingsTab);
+    app->getAction(KiwixApp::OpenHomePageAction)->setDisabled(libraryOrSettingsTab);
```
veloman-yunkan, line 531: This change won't compile - there is no variable `app` yet. Please fix it by moving the last (code cleanup) commit to the beginning of the PR.

sgourdas: Fixed, sorry for this.

veloman-yunkan (approved):

PR #1139, opened 2024-07-01
Settings monitor path QOL improvements

### LINKED ISSUE #1113: Few improvements around Settings directories

sgourdas:
Closes #1113 

Added functionality for the monitor path in settings:

- Tooltip when hovering over, to show the full path.
- Copy to clipboard icon, with a disappearing tooltip to indicate success.
- Truncation of the path in case it exceeds a limit. `...` replaces the hidden path, to allow for better readability of the directory's path.

### DIFF
--- a/src/settingsview.cpp
+++ b/src/settingsview.cpp
@@ -9,7 +9,7 @@
 
 namespace 
 {
-    QString formatDownloadDir(const QString& input) {
+    QString formatSettingsDir(const QString& input) {
         const int maxLength = 40;
         if (input.length() > maxLength) {
             QString suffix = input.right(maxLength);
@@ -24,13 +24,19 @@ SettingsView::SettingsView(QWidget *parent)
     : QWidget(parent)
     , ui(new Ui::Settings)
 {
+    SettingsManager *settingsMgr = KiwixApp::instance()->getSettingsManager();
     ui->setupUi(this);
     ui->widget->setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/_settingsManager.css"));
     connect(ui->zoomPercentSpinBox, QOverload<int>::of(&QSpinBox::valueChanged), this, &SettingsView::setZoom);
     connect(ui->moveToTrashToggle, &QCheckBox::clicked, this, &SettingsView::setMoveToTrash);
     connect(ui->reopenTabToggle, &QCheckBox::clicked, this, &SettingsView::setReopenTab);
     connect(ui->browseButton, &QPushButton::clicked, this, &SettingsView::browseDownloadDir);
-    connect(ui->downloadDirPathCopy, &QPushButton::clicked, this, &SettingsView::copyDownloadPathToClipboard);
+    connect(ui->downloadDirPathCopy, &QPushButton::clicked, [this, settingsMgr]() {
+        copySettingsPathToClipboard(settingsMgr->getDownloadDir(), ui->downloadDirPathCopy);
+    });
+    connect(ui->monitorDirPathCopy, &QPushButton::clicked, [this, settingsMgr]() {
+        copySettingsPathToClipboard(settingsMgr->getMonitorDir(), ui->monitorDirPathCopy);
+    });
     connect(ui->resetButton, &QPushButton::clicked, this, &SettingsView::resetDownloadDir);
     connect(ui->monitorBrowse, &QPushButton::clicked, this, &SettingsView::browseMonitorDir);
     connect(ui->monitorClear, &QPushButton::clicked, this, &SettingsView::clearMonitorDir);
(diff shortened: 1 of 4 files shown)


src/settingsview.cpp
```diff
@@ -176,27 +180,28 @@ void SettingsView::setReopenTab(bool reopen)
...
 
-void SettingsView::copyDownloadPathToClipboard()
+void SettingsView::copySettingsPathToClipboard(QPushButton* button)
 {
-    QString downloadPath = KiwixApp::instance()->getSettingsManager()->getDownloadDir();
-    QApplication::clipboard()->setText(downloadPath);
+    auto settingsManager = KiwixApp::instance()->getSettingsManager();
+    QString pathToCopy = (button == ui->downloadDirPathCopy) 
+        ? settingsManager->getDownloadDir() 
+        : settingsManager->getMonitorDir();
```
veloman-yunkan, line 192: This (or the function signature and semantics) is a little ugly. Either pass `pathToCopy` as an argument as well, or change the parameter type to something else (e.g. an enum or a string) based on which select both `button` and `pathToCopy` inside the function body.

veloman-yunkan (changes requested):

src/settingsview.cpp
sgourdas: Yeah, did not really like it as well but was not sure which was the best option. I am going to pass it as an argument to provide modularity also for future use.

src/settingsview.cpp
```diff
@@ -30,7 +30,12 @@ SettingsView::SettingsView(QWidget *parent)
     connect(ui->moveToTrashToggle, &QCheckBox::clicked, this, &SettingsView::setMoveToTrash);
     connect(ui->reopenTabToggle, &QCheckBox::clicked, this, &SettingsView::setReopenTab);
     connect(ui->browseButton, &QPushButton::clicked, this, &SettingsView::browseDownloadDir);
-    connect(ui->downloadDirPathCopy, &QPushButton::clicked, this, &SettingsView::copyDownloadPathToClipboard);
+    connect(ui->downloadDirPathCopy, &QPushButton::clicked, [this]() {
+        copySettingsPathToClipboard(ui->downloadDirPathCopy, KiwixApp::instance()->getSettingsManager()->getDownloadDir());
```
veloman-yunkan, line 34: Too long line. Introduce a local variable `settingsMgr` and use it whereever possible in this function.

src/settingsview.cpp
```diff
@@ -176,27 +180,23 @@ void SettingsView::setReopenTab(bool reopen)
 
 void SettingsView::onDownloadDirChanged(const QString &dir)
 {
-    ui->downloadDirPath->setText(formatDownloadDir(dir));
+    ui->downloadDirPath->setText(formatSettingsDir(dir));
     ui->downloadDirPath->setToolTip(dir);
 }
 
-void SettingsView::copyDownloadPathToClipboard()
+void SettingsView::copySettingsPathToClipboard(QPushButton* button, QString pathToCopy)
```
veloman-yunkan, line 187: 1. The order of parameters is somewhat unnatural. `pathToCopy` is the primary parameter for the functionality of this procedure. `button` is passed so that it can also perform the nice-to-have (tooltip) feature.
2. Also please pass the button by `const` pointer.

veloman-yunkan (changes requested):

src/settingsview.cpp
sgourdas: > The order of parameters is somewhat unnatural. pathToCopy is the primary parameter for the functionality of this procedure. button is passed so that it can also perform the nice-to-have (tooltip) feature.

Yeah I did that initially, do not why i changed it back...

> Also please pass the button by const pointer.

The `QToolTip::showText` function does not accept a const.

sgourdas: @veloman-yunkan is it ok to keep as non const?

kelson42 (approved):
LGTM from user perspective

sgourdas:
1 more commit and we should be ready then

src/settingsview.cpp
veloman-yunkan: > @veloman-yunkan is it ok to keep as non const?

Yes

src/settingsview.cpp
veloman-yunkan: I see that the *whereever possible in this function* part had to be emphasized.

veloman-yunkan (changes requested):
While we are waiting for #1118 to be merged, please also address https://github.com/kiwix/kiwix-desktop/pull/1139#discussion_r1677691385 (and it makes sense to do it in a separate code-beautification commit).

src/settingsview.cpp
sgourdas: Yeah sorry, I missed committing this chunk of the constructor changes

veloman-yunkan (commented):
LGTM, but I want to have #1118 merged first so I will provide a formal approval later.

veloman-yunkan (approved):

PR #1143, opened 2024-07-18
Implement Basic Text to Speech

### LINKED ISSUE #44: Implement TTS
It looks like Windows has this builtin:
https://www.youtube.com/watch?v=-GCxNR3QsC8

Was reported first at:
https://sourceforge.net/p/kiwix/feature-requests/878/

ShaopengLin:
Fix #44

Changes:
- Reading the entire or selected text on a web page in different languages.
- The reading can be triggered in the Edit Menu, Shortcut, and right-click on a page.
- TTS UI to select languages and voices, and stop the reading.
- Voice configurations are remembered between sessions.
- Language determined by Zim's Locale.

Shortcut:
- ALT + SHIFT + A: read article
- ALT + SHIFT + S: read selection
- ALT + SHIFT + X: stop reading
- ALT + SHIFT + V: open voice selector
- ALT + SHIFT + L: open language selector

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -545,6 +545,11 @@ void KiwixApp::saveWindowState()
   mp_session->setValue("windowState", getMainWindow()->saveState());
 }
 
+void KiwixApp::saveVoiceName(const QString& langName, const QString& voiceName)
+{
+  mp_session->setValue("voice/" + langName, voiceName);
+}
+
 void KiwixApp::restoreWindowState()
 {
   getMainWindow()->restoreGeometry(mp_session->value("geometry").toByteArray());
@@ -559,6 +564,11 @@ void KiwixApp::saveCurrentTabIndex()
 void KiwixApp::savePrevSaveDir(const QString &prevSaveDir)
 {
   mp_session->setValue("prevSaveDir", prevSaveDir);
+}
+
+QString KiwixApp::getSavedVoiceName(const QString& langName) const
+{
+  return mp_session->value("voice/" + langName, "").toString();
 }
 
 QString KiwixApp::getPrevSaveDir() const
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -392,6 +392,10 @@ void KiwixApp::createActions()
     CREATE_ACTION_SHORTCUT(ReadArticleAction, gt("read-article"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_A));
     CREATE_ACTION_SHORTCUT(ReadTextAction, gt("read-text"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_S));
     CREATE_ACTION_SHORTCUT(ReadStopAction, gt("read-stop"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_X));
+    CREATE_ACTION_SHORTCUT(ToggleTTSLanguageAction, gt("select-read-language"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_L));
+    CREATE_ACTION_SHORTCUT(ToggleTTSVoiceAction, gt("select-read-voice"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_V));
+    mpa_actions[ToggleTTSLanguageAction]->setCheckable(true);
+    mpa_actions[ToggleTTSVoiceAction]->setCheckable(true);
 
     CREATE_ACTION_ICON_SHORTCUT(PrintAction, "print", gt("print"), QKeySequence::Print);
     connect(mpa_actions[PrintAction], &QAction::triggered,
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -389,6 +389,9 @@ void KiwixApp::createActions()
     DISABLE_ACTION(HistoryBackAction);
     DISABLE_ACTION(HistoryForwardAction);
 
+    CREATE_ACTION_SHORTCUT(ReadArticleAction, gt("read-article"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_A));
+    CREATE_ACTION_SHORTCUT(ReadTextAction, gt("read-text"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_S));
+
     CREATE_ACTION_ICON_SHORTCUT(PrintAction, "print", gt("print"), QKeySequence::Print);
     connect(mpa_actions[PrintAction], &QAction::triggered,
             this, &KiwixApp::printPage);
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -391,6 +391,7 @@ void KiwixApp::createActions()
 
     CREATE_ACTION_SHORTCUT(ReadArticleAction, gt("read-article"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_A));
     CREATE_ACTION_SHORTCUT(ReadTextAction, gt("read-text"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_S));
+    CREATE_ACTION_SHORTCUT(ReadStopAction, gt("read-stop"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_X));
 
     CREATE_ACTION_ICON_SHORTCUT(PrintAction, "print", gt("print"), QKeySequence::Print);
     connect(mpa_actions[PrintAction], &QAction::triggered,
--- a/src/texttospeechbar.cpp
+++ b/src/texttospeechbar.cpp
@@ -7,7 +7,6 @@ TextToSpeechBar::TextToSpeechBar(QWidget *parent)
 {
     ui->setupUi(this);
 
-    m_speech.setLocale(QLocale::system().language());
     connect(&m_speech, &QTextToSpeech::stateChanged, this,
             &TextToSpeechBar::onStateChanged);
 
@@ -19,6 +18,9 @@ TextToSpeechBar::TextToSpeechBar(QWidget *parent)
             &TextToSpeechBar::stop);
     connect(ui->closeButton, &QPushButton::pressed,
                 this, &TextToSpeechBar::speechClose);
+
+    setupLanguageComboBox();
+    languageSelected(ui->langComboBox->currentIndex());
 }
 
 void TextToSpeechBar::speak(const QString &text)
@@ -29,6 +31,28 @@ void TextToSpeechBar::speak(const QString &text)
 void TextToSpeechBar::stop()
 {
     m_speech.stop();
+}
+
+void TextToSpeechBar::setupLanguageComboBox()
+{
+    ui->langLabel->setText(gt("language"));
+    ui->langComboBox->setMaxVisibleItems(10);
+
+    QLocale current = QLocale::system();
+    for (const auto& locale : m_speech.availableLocales())
+    {
+        const QString name(QString("%1 (%2)")
+                        .arg(QLocale::languageToString(locale.language()))
+                        .arg(locale.nativeLanguageName()));
+
+        ui->langComboBox->addItem(name, locale);
+        if (locale.name() == current.name())
+            current = locale;
+    }
+
+    ui->langComboBox->setCurrentIndex(ui->langComboBox->findData(current));
+    connect(ui->langComboBox, QOverload<int>::of(&QComboBox::currentIndexChanged),
+                this, &TextToSpeechBar::languageSelected);
 }
 
 void TextToSpeechBar::speechClose()
@@ -47,6 +71,12 @@ void TextToSpeechBar::speechShow()
 {
     show();
     setFocus();
+}
+
+void TextToSpeechBar::languageSelected(int index)
+{
+    const QLocale locale = ui->langComboBox->itemData(index).toLocale();
+    m_speech.setLocale(locale);
 }
 
 void TextToSpeechBar::keyPressEvent(QKeyEvent *event)
--- a/src/texttospeechbar.cpp
+++ b/src/texttospeechbar.cpp
@@ -109,3 +109,10 @@ void ComboBoxLineEdit::preventSelection()
 {
     setSelection(0, 0);
 }
+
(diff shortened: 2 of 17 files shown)


ShaopengLin:
I wasn't able to implement PAUSE, RESUME, and volume. 

PAUSE/RESUME: Qt relies on the underlying system's audio engines to perform these operations and Qt has quite some buggy behavior on those engines. I tried for a long time to find solutions but none is compatible with all engines and Qt versions. This leads to the compromise of simply having Speak & STOP.

Volume: From what I tested, the Linux "speech" engines' audio doesn't change during the reading, which in my opinion makes it useless.

Considering we are shipping cross-platform, it is unreliable to have features that work on one but not on another. If we can find a way to package Qt with a working [speech engine](https://doc.qt.io/qt-6/qttexttospeech-engines.html), we might be able to implement PAUSE/RESUME. However, I have no idea how to do that. @veloman-yunkan @mgautierfr Do you by any chance have an idea of how hard it would be to package an external library like [Flite](https://github.com/festvox/flite) with Kiwix-Desktop? If it's a huge hurdle and we only get the PAUSE/RESUME, then I wouldn't say the return on the effort is great.

The priority for this PR would be to have Kiwix-Desktop be compatible with Qt6.4+, CI builds, and a satisfactory UI/UX given the existing feature.

kelson42:
@ShaopengLin Thank you very much for your PR. For the moment I'm not able to compile it on my Ubuntu 22.04 but should move soon to 24.04 and hope this will allow me to have Qt 6.4. In the meantime we need to secure the backward compatibility with older version of Qt (both version 5 and 6). Please fix the compilation script and the code to secure that this feature is only activated if Qt6.4 or higher is available, otherwise just deactive the feature in a way both compilation and run are fine.

ShaopengLin:
@kelson42 The compilation is now for Qt 6.4 +.

kelson42:
@ShaopengLin I have moved to 24.04. Would you be able please to rebase on `main` so I can finally move forward with the testing?

ShaopengLin:
@kelson42 Done.

From my own testing, TTS from Qt is rather unstable across machines and sometimes crashes the application when used for no apparent reason/pattern at all. I have been trying my best to find out if there is a solution to the core problem, though little information really floats around the internet on this topic. Please try it on your machine and see if those issues persist. If I cannot figure out why those things occur, rather than relying on Qt's flawed implementation, the goal should be improving screen reader compatibilities.

When you test, the dfki packages are likely not going to work so select the other ones.
Screenshot from 2024-09-09 02-43-01

ShaopengLin:
@kelson42 I have found the root cause to the crash. Will put up a fixup commit soon...

ShaopengLin:
@kelson42 Should be ready for testing now.

kelson42:
@ShaopengLin I see the menu entry but nothing happens if I click on it: no sound not bottom bar visible. I use Ubuntu 24.04 wieth Qt 5.15.13.

ShaopengLin:
@kelson42 As mentioned in this [comment](https://github.com/kiwix/kiwix-desktop/pull/1143#issuecomment-2238324534) we moved compilation to Qt6.4+.

kelson42:
> @kelson42 As mentioned in this [comment](https://github.com/kiwix/kiwix-desktop/pull/1143#issuecomment-2238324534) we moved compilation to Qt6.4+. I will make a change so those menu entries are invisible in Qt 6.4 and below.

@ShaopengLin Then the option should not be available (or grayed)

kelson42:
@ShaopengLin For me, does not compile with Qt 6.4.2:
```
g++ -c -pipe -std=c++17 -Werror -I/usr/local/include/ -I/usr/local/include -I/usr/include/x86_64-linux-gnu -I/usr/include/p11-kit-1 -O2 -std=gnu++1z -Wall -Wextra -fPIC -D_REENTRANT -DVERSION=2.3.1 -DQT_DEPRECATED_WARNINGS -DQT_DISABLE_DEPRECATED_BEFORE=0x060000 -DQT_NO_DEBUG -DQT_WEBENGINEWIDGETS_LIB -DQT_PRINTSUPPORT_LIB -DQT_WIDGETS_LIB -DQT_WEBENGINECORE_LIB -DQT_QUICK_LIB -DQT_OPENGL_LIB -DQT_GUI_LIB -DQT_QMLMODELS_LIB -DQT_WEBCHANNEL_LIB -DQT_QML_LIB -DQT_NETWORK_LIB -DQT_QMLINTEGRATION_LIB -DQT_POSITIONING_LIB -DQT_CORE_LIB -I. -Isubprojects/QtSingleApplication/src -I/usr/include/x86_64-linux-gnu/qt6 -I/usr/include/x86_64-linux-gnu/qt6/QtWebEngineWidgets -I/usr/include/x86_64-linux-gnu/qt6/QtPrintSupport -I/usr/include/x86_64-linux-gnu/qt6/QtWidgets -I/usr/include/x86_64-linux-gnu/qt6/QtWebEngineCore -I/usr/include/x86_64-linux-gnu/qt6/QtQuick -I/usr/include/x86_64-linux-gnu/qt6/QtOpenGL -I/usr/include/x86_64-linux-gnu/qt6/QtGui -I/usr/include/x86_64-linux-gnu/qt6/QtQmlModels -I/usr/include/x86_64-linux-gnu/qt6/QtWebChannel -I/usr/include/x86_64-linux-gnu/qt6/QtQml -I/usr/include/x86_64-linux-gnu/qt6/QtNetwork -I/usr/include/x86_64-linux-gnu/qt6/QtQmlIntegration -I/usr/include/x86_64-linux-gnu/qt6/QtPositioning -I/usr/include/x86_64-linux-gnu/qt6/QtCore -I. -I. -I/usr/lib/x86_64-linux-gnu/qt6/mkspecs/linux-g++ -o contenttypefilter.o src/contenttypefilter.cpp
src/contenttypefilter.cpp: In constructor ‘ContentTypeFilter::ContentTypeFilter(QString, QWidget*)’:
src/contenttypefilter.cpp:16:35: error: ‘checkStateChanged’ is not a member of ‘QCheckBox’
   16 |         connect(this, &QCheckBox::checkStateChanged, this, &ContentTypeFilter::onStateChanged);
      |                                   ^~~~~~~~~~~~~~~~~
make: *** [Makefile:1437: contenttypefilter.o] Error 1

```

ShaopengLin:
@kelson42 That is not my change. I believe this is from https://github.com/kiwix/kiwix-desktop/pull/1205 where I believe @sgourdas should've made the deprecation for only Qt 6.7+, and not Qt 6+.

kelson42:
@sgourdas Can you please fix compilation regression we have now with Qt6? This is blocking review of this PR.

sgourdas:
> @sgourdas Can you please fix compilation regression we have now with Qt6? This is blocking review of this PR.

@kelson42 was this not fixed with #1212 by @ShaopengLin?

kelson42:
Actually I don't have Qt6.7+ on my Ubuntu 24.04. How should i test this? @ShaopengLin @sgourdas Do you use a PPA to get the most recent version?

sgourdas:
> Actually I don't have Qt6.7+ on my Ubuntu 24.04. How should i test this? @ShaopengLin @sgourdas Do you use a PPA to get the most recent version?

If I remember correctly I had built it from the git source.

Would you like me to test something?

ShaopengLin:
@kelson42 If I remember correctly I used the Qt online installer for Linux. After install I went to `Qt/<Version>/gcc_64/bin` and installs `qmake` with instruction [here](https://github.com/kiwix/kiwix-desktop#compilation-with-qt6).

kelson42:
@sgourdas If you change the prerequisites, you need to update the compilation script (here the project file) to be sure that the compilation stops properly with a clear error if the prerequisites are not satisfied. Can you please do that?

kelson42:
@ShaopengLin Now it compile with Qt6.4, but I have no clue how to activate the TTS (I see nothing in the "Edit" menu for example).
image

Can you please be pretty explicit/precise about all ways to activate the TTS?

ShaopengLin:
@kelson42 It seems speech library doesn't come directly with the Qt install, which explains why you see nothing at all. I will get back to you with the libraries you need to install.

kelson42:
@ShaopengLin I have installed `qt6-speech-dev` and now it seems I have the feature (but no language at all!?)... but there is as well a package `libqt5texttospeech5-dev`, so I wonder a bit why we have to run this with Qt6?!
image

ShaopengLin:
@kelson42 I believe you also might need `qt6-speech-speechd-plugin` or `qt6-speech-flite-plugin`, though I haven't tested the flite plugin.

I can make it a Qt5 feature as well, not going to be huge changes. I originally actually proposed this to be a Qt 5 feature [here](https://github.com/kiwix/kiwix-desktop/issues/44#issuecomment-2204508557). But back [here](https://github.com/kiwix/kiwix-desktop/pull/1143#issuecomment-2238324534) we decided to make this a Qt6.4+ only feature.

kelson42:
@ShaopengLin For me this is OK if this is a Qt6 only feature but this will slow down its introduction and will complexify a bit the merging (qt6 switch schedule is still not clear). If we can have the feature in Qt5 without a significant effort, then it would be better.

ShaopengLin:
@kelson42 Should work on Qt 5 as well now.

kelson42:
@ShaopengLin I works fine with Qt5, here a few remarks:
* If TTS is running and voice is changed, then please relaunch automatically with new voice
* If ZIM is in French, then language of the voice should be French
* Please stick the "Stop" button to the comboxes.

Otherwise it seems good for a first version IMHO!

kelson42:
> * If TTS is running and voice is changed, then please relaunch automatically with new voice

This still fails most of the time

ShaopengLin:
@kelson42 Voice relaunch and combobox should be clickable everywhere and no longer be selectable.

kelson42:
@ShaopengLin It's not necessary to require install (in deb package) of `qtspeech5-speechd-plugin` has this is a dependence of `libqt5texttospeech5-dev`

ShaopengLin:
> Please register in settings the language chosen by the user (one per language). 
@kelson42 To clarify, do you mean store the language chosen per-tab or per-zim

kelson42:
> > Please register in settings the language chosen by the user (one per language). 
> 
> 
> 
> @kelson42 To clarify, do you mean store the language chosen per-tab or per-zim

Neither the first nor the second. just keep track of which voice is used for which lang.

ShaopengLin:
@kelson42 Voice choices are saved now.

kelson42 (approved):
LGTM from a user perspective

ShaopengLin:
@veloman-yunkan will notify you when code is ready. We can focus on the bigger ones first.

ShaopengLin:
@veloman-yunkan Ready for review.

ShaopengLin:
@veloman-yunkan Also, Qt at `CI / Linux (focal)` does not have texttospeech. Will we be adding the package or will I have to disable TTS if the package does not exist?

kelson42:
> @veloman-yunkan Also, Qt at `CI / Linux (focal)` does not have texttospeech. Will we be adding the package or will I have to disable TTS if the package does not exist?

What does that technically means "does not have texttospeech"? It means no boice is available?

ShaopengLin:
@kelson42 The Qt TextToSpeech module is missing in that CI.

kelson42:
@ShaopengLin The problem is not really the CI but what will happen if for some reason we have Qt5 without all the necessary. My best guess is that we should be prepared to such a scenario and hide or desactivate the feature if this lack is detected at start. Or is that depending to a certain version of Qt5?

ShaopengLin:
@kelson42 Don't believe it should depend on a certain version of Qt5, but I suppose having safeguards against this would be ideal. I will change this shortly.

src/kiwixapp.cpp
```diff
@@ -544,6 +544,11 @@ void KiwixApp::saveWindowState()
   mp_session->setValue("windowState", getMainWindow()->saveState());
 }
 
+void KiwixApp::saveVoiceName(const QString& langName, const QString& voiceName)
+{
+  mp_session->setValue(langName + "/voice", voiceName);
```
veloman-yunkan, line 549: This way of generating the setting name from variable data introduces a possibility of conflict with another setting with a name generated in a different way. I think it is better to at least start setting names with a fixed known prefix (either move "voice" to the beginning, or prepend "tts:" or "text2speech:")

src/texttospeechbar.cpp
```diff
@@ -82,14 +82,29 @@ void TextToSpeechBar::resetVoiceComboBox()
...
+    for (int i = 0; i < ui->voiceComboBox->count(); i++)
+    {
+        if (m_voices[i].name() == currentVoice.name())
+            voiceIndex = i;
+        if (m_voices[i].name() == savedVoiceName)
+        {
+            voiceIndex = i;
+            break;
+        }
+    }
```
veloman-yunkan, line 105: Please introduce a function `int TextToSpeechBar::getVoiceIndex(const QString& voiceName) const` and use it by trying to first obtain the saved voice, or the current voice as a fallback.

src/texttospeechbar.cpp
```diff
@@ -140,6 +141,8 @@ void TextToSpeechBar::voiceSelected(int index)
     KiwixApp::instance()->saveVoiceName(currentLang, voice.name());
 
     m_speech.setVoice(voice);
+    if (m_speech.state() == QTextToSpeech::Speaking)
+        speak(m_text);
```
veloman-yunkan, line 145: IMHO, that won't be good UX. If possible (given the limitation of pause/resuming functionality of `QTextToSpeech`) it would be better to pause the speech, change the voice and resume the speech.

veloman-yunkan (changes requested):
Considering TTS as an accessibility feature for people with vision problems, it should be primarily accessible via keyboard. I know that kiwix-desktop's accessibility in that regard is quite poor, however introducing TTS without a way of controlling it via keyboard doesn't seem wise.

src/texttospeechbar.cpp
ShaopengLin: My first thought as well, unfortunately, QTextToSpeech's pause and resume barely work.

src/texttospeechbar.cpp
ShaopengLin: Done

src/kiwixapp.cpp

ShaopengLin:
Added a new set of shortcuts for TTS.

veloman-yunkan (approved):
LGTM. Didn't test.

PR #1144, opened 2024-07-18
Fix and update a few actions

kelson42:


kelson42:
Merging as this is pretty straight and fix CI

PR #1146, opened 2024-07-21
Fixed Tab and Bookmark Right to Left Layout Drawing

### LINKED ISSUE #1142: Title tab left padding is not OK
See
image

Pretty sure this is a regression, but not clear when it has been introduced

ShaopengLin:
Improve #594 Fix #1142 

Changes:
- Tab text alignment in RTL mode fixed. It was showing the tail of the title instead of the start.
- The left border of tabs was missing in RTL. Now each tab has both a left and right border. This makes the border a little thicker than before, but better UX than no border.
- I forgot to add the border on both sides when I refactored the search bar, which meant it has no left border in RTL mode. Fixed
- Title text now also has 2px padding.

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -425,8 +425,22 @@ void TabBar::paintEvent(QPaintEvent *e)
         bool textRightToLeft = tab_title.isRightToLeft();
         bool appRightToLeft = QWidget::isRightToLeft();
 
+        // See QTabBar::tab::padding value in resources/css/style.css
+        const int padding = 4;
         QRect tabTextRect = style()->subElementRect(QStyle::SE_TabBarTabText, &tabopt, this);
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+        // See QTabBar::tab::border value in resources/css/style.css
+        const int border = 1;
 
+        // Add Padding to left, right. Padding is 4px. Add 5 to account for 
+        // Extra pixel from border.
+        tabTextRect.setX(tabTextRect.x() + padding + border);
+        tabTextRect.setWidth(tabTextRect.width() - padding - border);
+#else
+        // Qt6 correctly adds left and right padding but now incorrectly adds
+        // top or bottom padding.
+        tabTextRect.setY(tabTextRect.y() - padding);
+#endif
         QRect fontTextRect = fontMetrics().boundingRect(tab_title);
 
         if (fontTextRect.width() > tabTextRect.width())
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -422,6 +422,8 @@ void TabBar::paintEvent(QPaintEvent *e)
         initStyleOption(&tabopt, i);
 
         bool need_fade_out = false;
+        bool textRightToLeft = tab_title.isRightToLeft();
+        bool appRightToLeft = QWidget::isRightToLeft();
 
         QRect tabTextRect = style()->subElementRect(QStyle::SE_TabBarTabText, &tabopt, this);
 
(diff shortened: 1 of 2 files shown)


kelson42 (changes requested):
If you look to the icon, there is now a vertical alignement problem: I don't see why the two "Ted" Icon, and title should not be aligned!

ShaopengLin:
Screenshot from 2024-07-30 01-41-25

@kelson42 The text is technically aligned to the icon. It is just the font size on the icon is different. If I misunderstood could you provide a screen shot of your alignment situation?

ShaopengLin:
Ignore my previous comment, It does seem the text is not aligning with respect to the icon...

kelson42:
@ShaopengLin Please look at all this alignment/margin at the pixel level, so things look perfect.

ShaopengLin:
@kelson42 Additional Fixes:
1. Fixed the text to icon alignment.
2. The tab bottom border disappears when it is the current tab. The icon is up by 1px. Text painting now adjusts to this change as well.
3. Close button also does not adjust to the border disappearance. Fixed as well.

kelson42:
Sorry, but this is still wrong:
image

* OK: What is green is OK, so the text is properly ligned verticaly
* OK: Top and bottom padding are the same at 5px, which is fine to me
* NOK: But why it left and right padding at 4px?! Should be 5 as well (or put 4pixel padding everywhere)
* NOK: Icon is 2 pixels too high!

ShaopengLin:
@kelson42 The weird behavior is from our padding and border handling. Now the padding should be 4px. The bottom padding is the same, we are just painting the border white as part of the focus animation.

Screenshot from 2024-07-30 17-06-09

kelson42 (approved):
LGTM
@veloman-yunkan Good for the code review

src/tabbar.cpp
```diff
@@ -463,7 +463,7 @@ void TabBar::paintEvent(QPaintEvent *e)
 
         int align = Qt::AlignVCenter;
         if (need_fade_out) {
-            align |= (right_to_left)? Qt::AlignRight : Qt::AlignLeft;
+            align |= (textRightToLeft == appRightToLeft) ? Qt::AlignLeft : Qt::AlignRight;
```
veloman-yunkan, line 471: This looks strange. Why is the case of `textRightToLeft` and `appRightToLeft` both being `false` aligned the same way as when both of them are `true`?

ShaopengLin: I will retest this and either make a explanation and visual, or correct it. A comment I believe will be needed here to explain the textbox behaviours.

ShaopengLin: The alignment in RTL is flipped, i.e. if you set Align Left, it will actually align it to the right. https://doc.qt.io/qt-6/qtquick-positioning-righttoleft.html.

ShaopengLin:
@veloman-yunkan @kelson42  I just noticed that the padding code here isn't working as wanted in Qt6.2. I will make a change and then we can re-review.

ShaopengLin:
Done

src/tabbar.cpp
```diff
@@ -427,6 +427,16 @@ void TabBar::paintEvent(QPaintEvent *e)
...
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+        // Add Padding to left, right. Padding is 4px. Add 5 to account for 
+        // Extra pixel from border.
+        tabTextRect.setX(tabTextRect.x() + 5);
+        tabTextRect.setWidth(tabTextRect.width() - 5);
+#else
+        // Qt6 correctly adds left and right padding but now incorrectly adds
+        // top or bottom padding.
+        tabTextRect.setY(tabTextRect.y() - 4);
+#endif
```
veloman-yunkan, line 439: Magic numbers are evil. Please introduce a local constant (`const int padding = 4)` with a comment linking it to the value of `QTabBar::tab::padding` in `resources/css/style.css` (also add a sibling comment in `resources/css/style.css`).

ShaopengLin: Done

veloman-yunkan (approved):
LGTM

@kelson42 Please test again before merging

PR #1164, opened 2024-07-29
Smarter directory monitoring

### LINKED ISSUE #1150: ZIM file downloaded in a monitored directory by an external tool may fail to be added to the library automatically
When `kiwix-desktop` monitors a directory it attempts to add new ZIM files to the library only when they appear. If a ZIM file is being downloaded by an external tool (that at the beginning of the download creates the file using its final name, rather than using a temporary file name indicating that the download is in progress and renaming upon completion), the attempt to open the ZIM file will most likely fail, it will be considered invalid and won't be added to the library. If the download completes during the same `kiwix-desktop` session that will go unnoticed by `kiwix-desktop`. However, once `kiwix-desktop` is restarted the file will appear in the library.

veloman-yunkan:
Fixes #1150

This is the best attempt at fixing #1150.

Directory monitoring was enhanced to work as follows:

- if adding a new file from a monitored directory fails, attempts to add it are repeated on further updates to that file until the file is good enough to be added to the library. Note that this doesn't mean that the ZIM file under question if perfectly valid, since adding to the library depends on minimal data from the ZIM file and will succeed as long as there are no issues with that data. If the file is still being downloaded by an external tool, opening that ZIM file for browsing may fail (then it will be dropped from the library) or certain content from it may fail to be loaded/displayed (the latter scenario has been addressed by #1174).
- care has been taken to minimize attempts to add a new ZIM file while the chances of succeeding at it are low.

The PR starts with refactoring - all code responsible for directory monitoring is transferred to `ContentManager` and cleaned up and only after that enhancement to that functionality is made.

### DIFF
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -879,10 +879,27 @@ void ContentManager::handleDisappearedZimFiles(const QStringSet& zimPaths)
     }
 }
 
+ContentManager::QStringSet ContentManager::handleNewZimFiles(const QStringSet& zimPaths)
+{
+    QStringSet successfullyAddedZims;
+    const auto kiwixLib = mp_library->getKiwixLibrary();
+    kiwix::Manager manager(kiwixLib);
+    for (auto bookPath : zimPaths) {
+        DBGOUT("directory monitoring: file appeared: " << bookPath);
+        if ( mp_library->isBeingDownloadedByUs(bookPath) ) {
+            DBGOUT("                      it is being downloaded by us, ignoring...");
+        } else if ( manager.addBookFromPath(bookPath.toStdString()) ) {
+            DBGOUT("                      and was added to the library");
+            successfullyAddedZims.insert(bookPath);
+        } else {
+            DBGOUT("                      but could not be added to the library");
+        }
+    }
+    return successfullyAddedZims;
+}
+
 void ContentManager::updateLibraryFromDir(QString monitorDir)
 {
-    typedef Library::QStringSet QStringSet;
-
     QMutexLocker locker(&m_updateFromDirMutex);
     const QDir dir(monitorDir);
     const QStringSet oldDirEntries = m_knownZimsInDir[monitorDir];
@@ -892,24 +909,9 @@ void ContentManager::updateLibraryFromDir(QString monitorDir)
     }
     const QStringSet addedZims = newDirEntries - oldDirEntries;
     const QStringSet removedZims = oldDirEntries - newDirEntries;
-    const auto kiwixLib = mp_library->getKiwixLibrary();
-    kiwix::Manager manager(kiwixLib);
-    bool needsRefresh = !removedZims.empty();
     handleDisappearedZimFiles(removedZims);
-    for (auto bookPath : addedZims) {
-        if ( mp_library->isBeingDownloadedByUs(bookPath) ) {
-            DBGOUT("directory monitoring: " << bookPath
-                    << " ignored since it is being downloaded by us.");
-        } else {
-            DBGOUT("directory monitoring: file appeared: " << bookPath);
-            const bool added = manager.addBookFromPath(bookPath.toStdString());
-            DBGOUT("                     "
-                   << (added ? "and was added" : "but could not be added")
-                   << " to the library");
-            needsRefresh |= added;
-        }
-    }
-    if (needsRefresh) {
+    const auto successfullyAddedZims = handleNewZimFiles(addedZims);
+    if (!removedZims.empty() || !successfullyAddedZims.empty()) {
         mp_library->save();
         emit(booksChanged());
         setMonitorDirZims(monitorDir, newDirEntries);
--- a/src/contentmanager.cpp
+++ b/src/contentmanager.cpp
@@ -865,9 +865,14 @@ void ContentManager::setMonitoredDirectories(QStringSet dirList)
         m_watcher.removePath(path);
     }
     m_knownZimsInDir.clear();
+    MonitoredZimFileInfo libraryZimFileInfo;
+    libraryZimFileInfo.status = MonitoredZimFileInfo::ADDED_TO_THE_LIBRARY;
     for (auto dir : dirList) {
         if (dir != "") {
-            m_knownZimsInDir[dir] = mp_library->getLibraryZimsFromDir(dir);
+            auto& zimsInDir = m_knownZimsInDir[dir];
+            for ( const auto& fname : mp_library->getLibraryZimsFromDir(dir) ) {
+                zimsInDir.insert(fname, libraryZimFileInfo);
+            }
             m_watcher.addPath(dir);
             asyncUpdateLibraryFromDir(dir);
         }
@@ -919,6 +924,7 @@ namespace
 
 // indexed by MonitoredZimFileInfo::ZimFileStatus enum
 const char* monitoredDirZimFileHandlingMsgs[] = {
+    "",
     "it is being downloaded by us, ignoring...",
     "the file was added to the library",
     "the file could not be added to the library"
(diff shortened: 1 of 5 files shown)


kelson42:
@veloman-yunkan The best at this stage is IMHO to merge straight, we need to move forward and I have trust the code is good.

veloman-yunkan:
> @veloman-yunkan The best at this stage is IMHO to merge straight, we need to move forward and I have trust the code is good.

Nope. This is too big a change to be merged without a review.

kelson42:
@ShaopengLin Can you please have a look? You are at this stage almost the only one who can do it.

veloman-yunkan:
> @ShaopengLin Can you please have a look? You are - at this stage - almost the only one who can do it.

Yes, please. And feel free to take revenge for all the pickiness on my side when reviewing your PRs :)

src/contentmanager.cpp
```diff
@@ -942,14 +943,32 @@ bool ContentManager::handleZimFileInMonitoredDirLogged(QString dir, QString file
...
-    Q_UNUSED(fileName);
-    // TODO: implement properly
-    MonitoredZimFileInfo zfi;
-    zfi.status = MonitoredZimFileInfo::PROCESS_NOW;
-    return zfi;
+    const auto bookPath = QDir::toNativeSeparators(dir + "/" + fileName);
+
+    MonitoredZimFileInfo zimFileInfo;
+
+    zimFileInfo.lastModified = QFileInfo(bookPath).lastModified();
```
mgautierfr, line 979: We may use appropriate constructor here.
I don't like to see structure which may be not fully initialized.

src/contentmanager.cpp
```diff
@@ -943,12 +945,26 @@ bool ContentManager::handleZimFileInMonitoredDirLogged(QString dir, QString file
...
+}
+
 void ContentManager::MonitoredZimFileInfo::updateStatus(const MonitoredZimFileInfo& prevInfo)
 {
     Q_ASSERT(prevInfo.status != ADDED_TO_THE_LIBRARY);
 
     if ( this->lastModified == prevInfo.lastModified ) {
         this->status = UNCHANGED_KNOWN_BAD_ZIM_FILE;
+    } else if ( prevInfo.status == PROCESS_LATER ) {
+        this->status = DEFERRED_PROCESSING_ALREADY_PENDING;
```
mgautierfr, line 965: I'm not sure to properly understand the status evolution of a zim being downloaded (behind our back)

`DEFERRED_PROCESSING_ALREADY_PENDING` is set only here and never read. So I suppose it is just a status to not set something else (which may trigger different action).

But when we go out of this status ? Is it handleZimFileInMonitoredDirDeferred which change it ?
But it changes it to PROCESS_NOW and so we add to the library "right now". Should we check again that file has been modified for a second ?

ShaopengLin: I believe `deferHandlingOfZimFileInMonitoredDir` is called first which sets the timer of 1 second, then `handleZimFileInMonitoredDirDeferred` is called to set it to PROCESS_NOW.

ShaopengLin (changes requested):
Mostly Stylistical requests. The addition process makes sense IMO.

src/contentmanager.cpp
ShaopengLin: I believe `deferHandlingOfZimFileInMonitoredDir` is called first which sets the timer of 1 second, then `handleZimFileInMonitoredDirDeferred` is called to set it to PROCESS_NOW.

src/contentmanager.cpp
```diff
@@ -864,10 +864,11 @@ void ContentManager::asyncUpdateLibraryFromDir(QString dir)
     });
 }
 
-void ContentManager::handleDisappearedZimFiles(const QStringSet& zimPaths)
+void ContentManager::handleDisappearedZimFiles(const QString& dirPath, const QStringSet& fileNames)
```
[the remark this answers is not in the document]
ShaopengLin: They actually do, but agree it isn't obvious from the diff log:

`m_knownZimFiles` is set by `setMonitorDirZims`, of which the passed-in value has changed;

`getLibraryZimsFromDir()` is changed, the name didn't appear as it just got truncated in the `src/library.cpp` change

src/contentmanager.cpp
```diff
@@ -22,6 +22,12 @@
 #include "contentmanagerheader.h"
 #include <QDesktopServices>
 
+#ifndef QT_NO_DEBUG
+#define DBGOUT(X) qDebug() << "DBG: " << X
+#else
+#define DBGOUT(X)
```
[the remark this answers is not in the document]
veloman-yunkan: This macro doesn't expand to a compound statement, hence no `do {} while(0)` is necessary here

src/contentmanager.h
```diff
@@ -123,7 +126,7 @@ public slots:
             COULD_NOT_BE_ADDED_TO_THE_LIBRARY
         };
 
-        ZimFileStatus status = ADDED_TO_THE_LIBRARY;
+        ZimFileStatus status = NO_INFO;
```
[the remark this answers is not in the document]
veloman-yunkan: `ContentManager::handleZimFileInMonitoredDir()` was (already) updated in the same commit to account for that change.

src/contentmanager.cpp
```diff
@@ -880,25 +895,17 @@ void ContentManager::updateLibraryFromDir(QString monitorDir)
...
         } else {
-            DBGOUT("ContentManager::updateLibraryFromDir(): "
-                   << "file appeared: " << bookPath);
+            DBGOUT("directory monitoring: file appeared: " << bookPath);
             const bool added = manager.addBookFromPath(bookPath.toStdString());
-            DBGOUT("                                        "
+            DBGOUT("                     "
                    << (added ? "and was added" : "but could not be added")
-                   << " to the library");
+                   << "to the library");
```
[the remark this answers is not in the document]
veloman-yunkan: Thanks for the `qDebug().nospace()` suggestion! I added two fixup commits in their respective positions in the PR history addressing this concern.

src/contentmanager.cpp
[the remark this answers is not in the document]
veloman-yunkan: @ShaopengLin's analysis is correct. I slightly rebalanced the changes between this commit and a preceding cleanup commit so that the diff is smaller and the reader's attention is not distracted by insignificant changes. I also amended the commit messages with an explanation.

src/contentmanager.cpp
veloman-yunkan: In a sense, `MonitoredZimFileInfo::lastModified` is an optional field. It is not used in the other location (in `ContentManager::setMonitoredDirectories()` where a `MonitoredZimFileInfo` object is constructed. IMHO, introducing a constructor for just this incarnation of `MonitoredZimFileInfo` is not justified (besides, an explicit assignment to `.lastModified` make the code more readable).

src/contentmanager.cpp
veloman-yunkan: > `DEFERRED_PROCESSING_ALREADY_PENDING` is set only here and never read. So I suppose it is just a status to not set something else (which may trigger different action).

Mostly yes (though `DEFERRED_PROCESSING_ALREADY_PENDING` is read implicitly by the logging code, but that is only for debugging).

> But when we go out of this status ? Is it handleZimFileInMonitoredDirDeferred which change it ? But it changes it to PROCESS_NOW and so we add to the library "right now". Should we check again that file has been modified for a second ?

This status is set only on a temporary variable and is never recorded in directory monitoring's state (`m_knownZimsInDir`). The same is true for the `PROCESS_NOW`, `BEING_DOWNLOADED_BY_US`, and `UNCHANGED_KNOWN_BAD_ZIM_FILE` statuses, however, I guess with them the situation is less confusing since they do not imply any delayed action - a decision how to proceed is made immediately (in `ContentManager::handleZimFileInMonitoredDir()`). In a sense, that is also the case with `DEFERRED_PROCESSING_ALREADY_PENDING` - as soon as it is discovered that a deferred action is already registered, we simply stop processing the current event. From that point of view `BEING_DOWNLOADED_BY_US`, `UNCHANGED_KNOWN_BAD_ZIM_FILE` and `DEFERRED_PROCESSING_ALREADY_PENDING` could be reduced to a single value `NOTHING_TO_BE_DONE` and exist merely for logging/debugging purposes.

veloman-yunkan:
Rebased on main and addressed review comments in fixup commits inserted at the right points in the PR history. Also moved some of the changes from the commit "Directory monitoring works via file names instead of paths" into "Renamed a few vars in ContentManager::updateLibraryFromDir()" and updated the description of the former change.

ShaopengLin (approved):
LGTM

Issue #1107, opened 2024-05-12
Windows 10 - Not Settings Menu, All Files Does Nothing  (windows_x64_2.0-alpha6-1)


Intrepid-MR:
RE: kiwix-desktop_windows_x64_2.0-alpha6-1
Windows 10 22H2

After running Kiwix, the following issues occur:
1. "Settings" is missing from the menu.
2.  Clicking on "All Files" does nothing.

kelson42:
This is a super old verwion of the software. Plase try with latest release https://kiwix.org/en/applications/

Intrepid-MR:
The ONLY Windows version available at https://kiwix.org/en/applications/
is in fact:  2.0-alpha6-1
There is NO more recent Windows version.

kelson42:
Sorry we have created a regression here

rgaudin:
@Intrepid-MR thank you for reporting ; we only provide a single link to the latest version. You made no mistake but we introduced a bug in the redirection (the link is always the same but its redirection target changes with new releases).

PR #1169, opened 2024-08-04
Migrate getDataDirectory from libkiwix to kiwix-desktop

sgourdas:
This migrates the `getDataDirectory()` function from libkiwix, after some cleanups.
 - Also removes `KIWIX_DATA_DIR`

Fixes kiwix/libkiwix#1087 
Fixes kiwix/libkiwix#1085

Blocked by https://github.com/kiwix/libkiwix/pull/1107

### DIFF
--- a/src/downloadmanagement.cpp
+++ b/src/downloadmanagement.cpp
@@ -105,7 +105,7 @@ namespace
 kiwix::Downloader* createDownloader()
 {
     try {
-        return new kiwix::Downloader();
+        return new kiwix::Downloader(getDataDirectory().toStdString());
     } catch (std::exception& e) {
         QMessageBox::critical(nullptr, gt("error-downloader-window-title"),
         gt("error-downloader-launch-message") + "<br><br>" + e.what());
@@ -283,19 +283,17 @@ void DownloadManager::checkThatBookCanBeDownloaded(const kiwix::Book& book, cons
 
 std::string DownloadManager::startDownload(const kiwix::Book& book, const QString& downloadDirPath)
 {
-    typedef std::vector<std::pair<std::string, std::string>> DownloadOptions;
-
     const std::string& url = book.getUrl();
     const QString bookId = QString::fromStdString(book.getId());
-    const DownloadOptions downloadOptions{{"dir", downloadDirPath.toStdString()}};
 
     std::string downloadId;
     try {
-        const auto d = mp_downloader->startDownload(url, downloadOptions);
+        const auto d = mp_downloader->startDownload(url, downloadDirPath.toStdString());
         downloadId = d->getDid();
     } catch (std::exception& e) {
         throwDownloadUnavailableError();
     }
+    
     return downloadId;
 }
 
--- a/src/settingsmanager.cpp
+++ b/src/settingsmanager.cpp
@@ -7,6 +7,16 @@
 #include <QLocale>
 #include <QList>
 
+QString getDataDirectory()
+{
+    QString dataDir = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+    
+    if (!dataDir.isEmpty() && QDir().mkpath(dataDir))
+        return dataDir;
+    
+    return QString::fromStdString(kiwix::getCurrentDirectory());
+}
+
 SettingsManager::SettingsManager(QObject *parent)
     : QObject(parent),
     m_settings("Kiwix", "Kiwix-desktop"),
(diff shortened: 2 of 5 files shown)


src/downloadmanagement.h
```diff
@@ -207,3 +207,5 @@ class DownloadManager : public QObject
 };
 
 #endif // DOWNLOADMANAGEMENT_H
+
+std::string getDataDirectory();
```
veloman-yunkan, line 211: I don't think that this function should be declared in this header file, but if there is no better place then it should at least be moved to before the `#endif`

src/kiwixapp.h
```diff
@@ -10,6 +10,7 @@
 #include "kprofile.h"
 #include "settingsmanager.h"
 #include "translation.h"
+#include "downloadmanagement.h"
```
veloman-yunkan, line 13: Better include in the .cpp file

src/downloadmanagement.h
veloman-yunkan: I think that `settingsmanager.h/cpp` is a more appropriate home for this function's declaration and definition.

veloman-yunkan (changes requested):

src/kiwixapp.h
sgourdas: Just for future reference, how should I decide where to include each header file?

veloman-yunkan: Try to minimize dependencies (and don't introduce false dependencies). "kiwixapp.h" doesn't depend on "downloadmanagement.h" (the former doesn't use anything declared in the latter). By including "downloadmanagement.h" in "kiwixapp.h" you grow the list of formal dependencies of "downloadmanagement.h". Then if anything changes in "downloadmanagement.h" more code will be recompiled. If you include "downloadmanagement.h" only in the .cpp files that use stuff from that header, then recompilation in case of touching it will be faster.

kelson42:
@sgourdas @veloman-yunkan What is the status here, it seems it needs a new review?

sgourdas:
@kelson42 yes, ready for a new review

src/settingsmanager.cpp
```diff
@@ -7,6 +7,50 @@
...
+    dataDir = cDataDir;
+  } else {
+    cDataDir = ::getenv("HOME");
+    if (cDataDir != nullptr) {
+      dataDir = cDataDir;
+      dataDir = kiwix::appendToDirectory(dataDir, ".local");
+      dataDir = kiwix::appendToDirectory(dataDir, "share");
+    }
+  }
+#endif
```
mgautierfr, line 45: I wonder if we could remove this and replace it with Qt function (writableLocation)[`https://doc.qt.io/qt-6/qstandardpaths.html#writableLocation`] (or (standardLocation)[https://doc.qt.io/qt-6/qstandardpaths.html#standardLocations], depending if we want to read data (configuration, zim files) or write it)

`QStandardPaths::AppDataLocation` seems to return what the commented lines are doing (at least on Windows and Linux)

veloman-yunkan:
> @sgourdas @veloman-yunkan What is the status here, it seems it needs a new review?

@kelson42 I will review it as soon as kiwix/libkiwix#1107 is merged

src/downloadmanagement.cpp
```diff
@@ -102,10 +102,10 @@ void DownloadState::changeState(Action action)
 namespace
 {
 
-kiwix::Downloader* createDownloader()
+kiwix::Downloader* createDownloader(std::string dataDir)
```
veloman-yunkan, line 105: Undo this change. Call `getDataDirectory()` inside this function

mgautierfr:
> I will review it as soon as https://github.com/kiwix/libkiwix/pull/1107 is merged

I have revert the merge of #1107 as it breaks the nightly CI until this PR is merged.
Please review this PR as if #1107 is merged and I will revert the revert PR (https://github.com/kiwix/libkiwix/pull/1110)

src/settingsmanager.cpp
sgourdas: Good idea. I looked into it and the only issue I see is the intermediate differences between Qt and us currently For example, I have the resolutions below: 

## Windows
Current:

1. `C:\Users\<USER>\AppData\Roaming\kiwix`
2. `C:\Users\<USER>\kiwix`
3. `<APPDIR>`

AppDataLocation:

1. `C:/Users/<USER>/AppData/Roaming/kiwix`
2. `C:/ProgramData/kiwix`
4. `<APPDIR>`
5. `<APPDIR>/data`
6. `<APPDIR>/data/kiwix`

## Rest
Current:

1. `~/.local/share/kiwix`
2. `<APPDIR>`

AppDataLocation:

1. `~/.local/share/<APPNAME>`
3. `/usr/local/share/<APPNAME>`
4. `/usr/share/<APPNAME>`

How should we go about this?

kelson42:
@veloman-yunkan We need a review now, before merging the libkiwix part, otherwise in the meantime the whole kiwix desktop CI will be broken.

sgourdas:
On this I am waiting for a reply to [my comment](https://github.com/kiwix/kiwix-desktop/pull/1169/files#r1714974074) on if it is a good way to proceed.

sgourdas:
I have made a suggestion commit to showcase the new implementation. Using `writableLocation` we get only the first valid path from the possible list of `AppDataLocation` and create it if not already there. This should work as intended.

kelson42:
@veloman-yunkan We need a new review pass I guess.

veloman-yunkan (commented):
There are too many commits in the PR history

veloman-yunkan (approved):

PR #1170, opened 2024-08-05
Disable sandbox when run from Windows network share

### LINKED ISSUE #885: Failing Kiwix if started from Windows Network Drive (was: Kiwix does not start on Windows: "QT5Core.dll was not found")
I noticed that Kiwix (version: kiwix-desktop_windows_x64_2.3.0) can't start properly from mapped network shares.
I got "The code execution cannot proceed because QT5Core.dll was not found. Reinstalling the program may fix this problem."
On regular drive everything works fine.

sgourdas:
This aims to close #885 by adding the `--no-sandbox` command line argument.

This allows the user to manage the risk and turn off the sandbox for example when running from a network mounted drive on Windows.

Also, so as to apply the no sandbox flag before the KiwixApp has been instantiated, the argument check is being done with the usual way of checking `argv` instead of using `QCommandLineParser` which runs on `KiwixApp::a`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5,6 +5,10 @@
 #include <QCommandLineParser>
 #include <iostream>
 #include <sstream>
+#ifdef Q_OS_WIN
+    #include <kiwix/tools.h>
+    #include <windows.h>
+#endif
 #if QT_VERSION >= QT_VERSION_CHECK(5, 12, 0)
   #include <QWebEngineUrlScheme>
 #endif
@@ -18,6 +22,15 @@ int main(int argc, char *argv[])
     }
 // End of hack ^^^
 
+#ifdef Q_OS_WIN
+    std::string driveLetter = kiwix::getExecutablePath().substr(0, 3);
+    qInfo() << driverLetter;
+    UINT driveType = GetDriveTypeA(driveLetter.c_str());
+    if(driveType == DRIVE_REMOTE) {
+        qputenv("QTWEBENGINE_CHROMIUM_FLAGS", QByteArray("--no-sandbox"));
+        qInfo() << "Disabled sandbox";
+    }
+#endif
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     // High DPI Scaling is enabled by default in Qt6. This attribute no longer exists in 6.0 and later
     QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);


sgourdas:
How should I test this since it is Windows specific? nightly?

kelson42:
I guess we have no way to test this properly before mergin unfortunately.

kelson42 (changes requested):
Please update the `usage()` so if we type `kiwix-desktop --help` we get a proper explanation. Regarding the parsing of the command line, we don't have already a lib, a place (already there) doing that properly?

sgourdas:
> Please update the usage() so if we type kiwix-desktop --help we get a proper explanation.

I am not sure I understand. What else should be populated?
image

> Regarding the parsing of the command line, we don't have already a lib, a place (already there) doing that properly? We should be able to handle this in a better manner.

We do indeed, I assume you are talking about QCommandLinerParser. The issue is that it needs the app to be instantiated before it can check the parameters. In our case we need to check and apply the no sandbox variable before app init so this creates an issue. The only workaround I have found is creating a dummy `QCoreApplication` which I do not really like as a solution.

veloman-yunkan (commented):
Do we really need the `--no-sandbox` option?

Under Windows, GUI applications are rarely started from the command line. Of course we can provide a wrapper bat file `kiwix-desktop-no-sandbox.bat` for starting `kiwix-desktop` in that mode, but then we can simply set the `QTWEBENGINE_CHROMIUM_FLAGS` env var in that script directly.

kelson42:
I can only agree, we could/should better provide a `.bat` launcher. But by doing so, having a C++ piece of code.

In general, I don't like to provide necessary features to users via ENV because this is not clear/transparent and an error happens very easily without a chance to really detect it easily. But I get that here there is really a risk of overengineering if we provide a `.bat` launcher.

All in one, I believe I would still prefer to have this properly in C++, but whatever approach chosen at the end, this is not a blocker to me.

kelson42:
> > Please update the usage() so if we type kiwix-desktop --help we get a proper explanation.
> 
> I am not sure I understand. What else should be populated? image

Juste me been blind, sorry. Please add something like "Necessary to run Kiwix from Windows network drive." So this is clear why we have this option.

> > Regarding the parsing of the command line, we don't have already a lib, a place (already there) doing that properly? We should be able to handle this in a better manner.
> 
> We do indeed, I assume you are talking about QCommandLinerParser. The issue is that it needs the app to be instantiated before it can check the parameters. In our case we need to check and apply the no sandbox variable before app init so this creates an issue. The only workaround I have found is creating a dummy `QCoreApplication` which I do not really like as a solution.

OK, I let you see the details with @veloman-yunkan. LGTM then.

kelson42 (approved):
Looks good from user perspective (even if not tested on Windows)

veloman-yunkan:
> All in one, I believe I would still prefer to have this properly in C++, but whatever approach chosen at the end, this is not a blocker to me.

I would agree with a C++ implementation if it works as follows:

1. detects that `kiwix-desktop` is started from a network location,
2. tells the user about it and asks for confirmation to continue under relaxed security constraints.

Otherwise, given that we provide a `.bat` file (that will hide from the user how that issue is solved under the hood), I can't justify a command-line option being added to `kiwix-desktop`.

kelson42:
> > All in one, I believe I would still prefer to have this properly in C++, but whatever approach chosen at the end, this is not a blocker to me.
> 
> I would agree with a C++ implementation if it works as follows:
> 
> 1. detects that `kiwix-desktop` is started from a network location,
> 2. tells the user about it and asks for confirmation to continue under relaxed security constraints.

@sgourdas doable?

sgourdas:
@kelson42, definitely. I think that since we only want to do this for Windows an easy and valid check should be checking the absolute filepath's prefix for `\\`

veloman-yunkan:
> @kelson42, definitely. I think that since we only want to do this for Windows an easy and valid check should be checking the absolute filepath's prefix for `\\`

This won't work for mapped network shares (that's the case of the ticket, their path starts with `Q:\`).

sgourdas:
> > @kelson42, definitely. I think that since we only want to do this for Windows an easy and valid check should be checking the absolute filepath's prefix for `\\`
> 
> This won't work for mapped network shares (that's the case of the ticket, their path starts with `Q:\`).

I can see. And I cannot really easily test it sadly. 

What seems like a more reliable way is [GetDriveType](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getdrivetypea) from `windows.h`. I will investigate and come back with a commit.

sgourdas:
@kelson42 @veloman-yunkan can we use nightly to test? Is there another option?

veloman-yunkan:
> @kelson42 @veloman-yunkan can we use nightly to test? Is there another option?

I think that you can clone the `kiwix-build` repository and modify the [ci](https://github.com/kiwix/kiwix-build/blob/main/.github/workflows/ci.yml) github flow and/or some other scripts so that it uses your branch of `kiwix-desktop`. This will build under Windows natively (via the MSVC compiler).

I think that it is also possible to cross-compile for Windows under Linux (using one of the `kiwix-build` options `--config win32_dyn` or `--config win32_static`).

kelson42:
@veloman-yunkan We are really getting rid of cross-compilation and moving all compalition for Windows on Windows. This is something @mgautierfr has been working on since months.

mgautierfr:
> I think that it is also possible to cross-compile for Windows under Linux (using one of the kiwix-build options --config win32_dyn or --config win32_static).

Today we are building kiwix-desktop for windows only on the appveyor CI (as @kelson42 said, I'm working on it to be able to build it locally/in GH CI).
One with commit rights on kiwix-build can trigger a test build of kiwix-desktop with commits like the last two ones in https://github.com/kiwix/kiwix-build/tree/build_dev_kiwix-desktop
(This also means that a build of this PR will be available soon in https://tmp.kiwix.org/ci/dev_preview/build_dev_kiwix-desktop/)

kelson42:
All of this is far two complicated. I merge and we will see. The solution is clearly to have:
* An easy compilation on Windows (relying on a prebuild of libkiwix/libzim)
* A Windows CI in this repo.

PR #1174, opened 2024-08-13
Handling of invalid ZIM files

### LINKED ISSUE #1128: App should not crash at start if one of the ZIM file is corrupted
I have achieved to get a corrupted ZIM file, see https://github.com/kiwix/kiwix-desktop/issues/1127 for the details.

Trying to open the file, Kiwix just crashes. Here the details:
https://tmp.kiwix.org/corrupted_zim_crash/

When I restart Kiwix, it crash at start and I can not use Kiwix at all anymore

What I expect:
If there is crash by reading a ZIM file, I expect Kiwix to detect it and stop (somehow) crashing when I restart it.

veloman-yunkan:
Fixes #1128

When trying to display resources from an invalid ZIM file that fail to be retrieved due to problems with the ZIM data, now an HTML error page is returned instead. Note that the resource being accessed may in fact be an image, a CSS or a javascript file and the user may not see the error (though the appearance of the webpage trying to load and use that resource will be affected to various degrees).

To test the fix use the data from https://tmp.kiwix.org/corrupted_zim_crash/ and override the `library.xml` and `kiwix-desktop.session` files in your `~/.local/share/kiwix` directory with those found in the [attached ZIP archive](https://github.com/user-attachments/files/16597904/issue1128.zip). After `kiwix-desktop` is started it will contain two tabs:

1. Home page of the ZIM file - 4 books ("*Les misérables Tome I: Fantine*", "*Histoire de France 1440-1465 (Volume 7/19)*", "*Anciennes loix des François, conservées dans les coutumes angloises, recueillies par Littleton, Vol. I*", and "*Lexicon Latinum : $b Universae phraseologiae corpus congestum etc.*") will have their thumbnails replaced with a placeholder image because of the inability to load the image from the ZIM file (an HTML error page is instead served). As you can see it is a very weak indication that something is wrong with the ZIM file
2. The second tab refers to one of those images directly and the error is obvious


ShaopengLin (commented):
I still only got run_time error after following the instructions. Otherwise the change LGTM

resources/i18n/en.json
```diff
@@ -174,6 +174,9 @@
     "zim-id": "Zim Id",
     "zim-name": "Zim Name",
     "zim-path": "Zim File Path",
+    "bad-zim-file-error-page-title": "Invalid ZIM File",
+    "bad-zim-file-error-page-text": "ZIM file is invalid or corrupted.",
```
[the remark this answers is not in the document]
kelson42: This should always be 'ZIM'

kelson42 (changes requested):
I have tested and it works fine. What I like is that to a large extend the app tries to deliver the content and fails only is a specific content is missing.

But we need:
* To put "ZIM" in strings (instead of "zim")
* The error page is not UTF8 charset declared, see:
* 
image

src/urlschemehandler.cpp
```diff
@@ -216,6 +218,33 @@ UrlSchemeHandler::replyZimNotFoundPage(QWebEngineUrlRequestJob *request,
...
+                          "<p>"  + gt("zim-path") + ": <b>" + path  + "</b></p>"
+                          "<p>" +  gt("zim-entry-path") + ": <b>" + zimEntryPath + "</b></p>"
+                          "</div></section>";
+
+    buffer->open(QIODevice::WriteOnly);
+    buffer->write(contentHtml.toStdString().c_str());
+    buffer->close();
+
+    connect(request, SIGNAL(destroyed()), buffer, SLOT(deleteLater()));
+    request->reply("text/html", buffer);
```
mgautierfr, line 245: I wonder if we should not set a error code to something else than 200.
Request may be made by js script which will check the error code.
But I haven't found an API allowing to return a specific content and set error code in the same time.

src/urlschemehandler.cpp
```diff
@@ -216,6 +218,33 @@ UrlSchemeHandler::replyZimNotFoundPage(QWebEngineUrlRequestJob *request,
...
+    const QString zimEntryPath = request->requestUrl().path();
+
+    QString contentHtml = "<section><div>"
+                          "<h1>" + gt("bad-zim-file-error-page-title") + "</h1>"
+                          "<p>"  + gt("bad-zim-file-error-page-text") + "</p>"
+                          "<p>"  + gt("zim-id")   + ": <b>" + zimId + "</b></p>"
+                          "<p>"  + gt("zim-name") + ": <b>" + name  + "</b></p>"
+                          "<p>"  + gt("zim-path") + ": <b>" + path  + "</b></p>"
+                          "<p>" +  gt("zim-entry-path") + ": <b>" + zimEntryPath + "</b></p>"
+                          "</div></section>";
```
mgautierfr, line 238: Shouldn't we put all this text as a translatable template ?
RTL language may need to be formatted differently.

mgautierfr (changes requested):

src/urlschemehandler.cpp
veloman-yunkan: I had the same difficulty while reviewing #1120. But since there we didn't have such a potential problem with accessing a resource via js, I [didn't mind](https://github.com/kiwix/kiwix-desktop/pull/1120#issuecomment-2151808886) about that scenario. Should we use [`QWebEngineUrlRequestJob::fail()`](https://doc.qt.io/qt-5/qwebengineurlrequestjob.html#fail) in this case (but then no special information about the error will be shown to the user)?

src/urlschemehandler.cpp
veloman-yunkan: This problem was introduced in #1132, I only duplicated it. Maybe we can address it in a separate PR.

veloman-yunkan:
Rebased on main and addressed review comments of @kelson42 and @ShaopengLin in new commits. @mgautierfr has identified a [good point](https://github.com/kiwix/kiwix-desktop/pull/1174#discussion_r1721720011) that has to be answered before we merge this PR or decide to change its implementation.

src/urlschemehandler.cpp
kelson42: I don't know about RTL, but the current way of doing is the one recommended by TW.

veloman-yunkan: The concern is about "*key*: *value*" lines. How will they be displayed if *key* is in an RTL language/script and *value* is in a LTR script?

kelson42: Honestly pretty sure it will be all-right. Merging and we will fix if not, but this is anyway a general problem then.

veloman-yunkan:
@kelson42 I think that you missed the [question](https://github.com/kiwix/kiwix-desktop/pull/1174#discussion_r1721720011) that had to be answered before merging the PR.

kelson42:
@veloman-yunkan mist, was sure there was only one indeed.

Issue #460, opened 2020-06-09
Kiwix Desktop does not ignore other libraries when local library.xml is available


kelson42:
I have a local Kiwix Desktop for Linux on my USB stick, with a local empty library.
```
$ find .
.
./library.xml
./kiwix-desktop_x86_64_2.0.1-1.appimage
kelson@camber:/media/kelson/Kiwix/Kiwix$ cat library.xml 
<library version="20110515"> </library>
```

But if I started, I don't get an empty local library:
![image](https://user-images.githubusercontent.com/1029718/84135784-ab82cf80-aa4a-11ea-9665-46d75a44bf99.png)

It seems to still read my other libraries.

kelson42:
Will be fixed at the same time as #1166 IMHO

sgourdas:
@kelson42 isn't this due to the monitor directory in settings? If you reset it, it should not transfer the zim files in the new `library.xml`.

```
All ZIM files in this directory will be automatically added to the library.
```

kelson42:
@sgourdas AFAIK there is current a logic to load `library.xml` from local and profile directory if it exists. Can you please secure that it has been removed properly/it's not the case anymore?

sgourdas:
@kelson42 it is not clear to me. From my experience and knowledge the only thing that loads information from other sources is through the "monitor" function. Am I wrong on this?

kelson42:
@sgourdas I just verified and it works like in the documentation. To me this principle of mixing local and system configuration is made deprecated by the implementation of the `.portable` approach and should be therefore be removed.

sgourdas:
@kelson42 the `KIWIX_DATA_DIR` option was removed with #1169.

I imagine you are referring to this and you are correct.
>As Kiwix Desktop can download and update the library.xml for you. It may be easier to use it. If your default library correspond to the files you want to put on your USB drive. You can simply copy the library.xml and all ZIM files in the user directory top the same directory than the Kiwix Desktop executable.

I will be coming back with a short PR to stop library loading from the executable's path as priority.

sgourdas:
Should I also properly update the [relevant section](https://github.com/kiwix/kiwix-desktop/wiki/Kiwix-desktop-on-an-USB-Drive-(Portable-version)#principle) after merging?

kelson42:
@sgourdas Not necessary, this is a topic for #461

PR #1179, opened 2024-08-16
Implement portable mode

### LINKED ISSUE #1166: Portable Kiwix, allow to custom all local storage/configuration locations
In a case of a portable Kiwix (running on a portable USB drive for example), we want to keep everything local on USB drive. We don't want to interfere with a potential local instance. Therefore, none of the [data and configuration files](https://github.com/kiwix/kiwix-desktop/wiki/Data-and-Configuration-Files) should be written [at the usual locations](https://github.com/kiwix/kiwix-desktop/wiki/Data-and-Configuration-Files).

A portable Kiwix is technically defined when there is (an empty) file `.portable` just beside the Kiwix binary. Detecting the presence of that file will activate the portable mode.

The portable mode will read/write the data files to another (predefined) location positioned relatively position to the Kiwix binary:
* `data/kiwix-desktop.session`
* `data/kiwix.session`
* `data/*.zim`
* ...

In such a mode nothing should be written outside the local Kiwix `data` directory.

This issue superseeds [older issuer #630](https://github.com/kiwix/kiwix-desktop/issues/630)

sgourdas:
This adds a check for the `.portable` file besides the executable. If it exists all application files will be saved in the `data` directory.

Fixes #1166

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -142,38 +142,36 @@ void KiwixApp::newTab()
 
 QString KiwixApp::findLibraryDirectory()
 {
-  // Check for library.xml in the same directory than the executable (portable kiwix-desktop)
-  auto currentDataDir = QString::fromStdString(kiwix::removeLastPathElement(kiwix::getExecutablePath()));
-  auto libraryFile = QFileInfo(currentDataDir, "library.xml");
-  if (libraryFile.exists())
-    return currentDataDir;
+    auto currentDataDir = QString::fromStdString(kiwix::removeLastPathElement(kiwix::getExecutablePath()));
+    // Check for library.xml in the same directory as the executable.
+    auto libraryFile = QFileInfo(currentDataDir, "library.xml");
+    if (libraryFile.exists())
+        return currentDataDir;
 
-  // Check for default dataDirectory.
-  currentDataDir = getDataDirectory();
-  libraryFile = QFileInfo(currentDataDir, "library.xml");
-  if (libraryFile.exists())
-    return currentDataDir;
+    // Check for default dataDirectory.
+    currentDataDir = getDataDirectory();
+    libraryFile = QFileInfo(currentDataDir, "library.xml");
+    if (libraryFile.exists())
+        return currentDataDir;
 
-  // There is no library.xml in default dataDirectory.
-  // Either, it is a first launch, or user used a pre-release version with wrong directory.
-  // Let's try to move data from old directory to new one if needed.
-  auto oldDataDir = QDir(currentDataDir);
-  oldDataDir.cdUp();
-  libraryFile = QFileInfo(oldDataDir, "library.xml");
-  if (libraryFile.exists()) {
-    // We have to move all zims file and xml file to the new dataDir
-    for (auto& fileInfo: oldDataDir.entryInfoList({"*.zim", "library*.xml"})) {
-      auto newFileInfo = QFileInfo(currentDataDir, fileInfo.fileName());
-      QFile::rename(fileInfo.absoluteFilePath(), newFileInfo.absoluteFilePath());
+    // Check if this is a pre-release version with wrong directory.
+    auto oldDataDir = QDir(currentDataDir);
+    oldDataDir.cdUp();
+    libraryFile = QFileInfo(oldDataDir, "library.xml");
+    if (libraryFile.exists()) {
+        // We have to move all zim files and xml file to the new dataDir
+        for (auto& fileInfo: oldDataDir.entryInfoList({"*.zim", "library*.xml"})) {
+            auto newFileInfo = QFileInfo(currentDataDir, fileInfo.fileName());
+            QFile::rename(fileInfo.absoluteFilePath(), newFileInfo.absoluteFilePath());
+        }
+        // Aria2 stores information about the current download using absolute path.
+        // Let's remove everything. User will lose its ongoing download but it should be pretty rare.
+        for (auto& fileInfo: oldDataDir.entryInfoList({"*.aria2", "*.meta4", "kiwix.session"}))
+            QFile::remove(fileInfo.absoluteFilePath());
     }
-    // Aria2 store informations about the current download using absolute path.
-    // Let's remove everything. User will loose its ongoing download but it should be pretty rare.
-    for (auto& fileInfo: oldDataDir.entryInfoList({"*.aria2", "*.meta4", "kiwix.session"})) {
-      QFile::remove(fileInfo.absoluteFilePath());
-    }
-  }
 
-  return currentDataDir;
+    // This is a first launch
+    return currentDataDir;
 }
 
 void KiwixApp::restoreTabs()
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -143,6 +143,10 @@ void KiwixApp::newTab()
 QString KiwixApp::findLibraryDirectory()
 {
     auto currentDataDir = QString::fromStdString(kiwix::removeLastPathElement(kiwix::getExecutablePath()));
+    
+    if (isPortableMode())
+        return currentDataDir + QDir::separator() + "data";
+
     // Check for library.xml in the same directory as the executable.
     auto libraryFile = QFileInfo(currentDataDir, "library.xml");
     if (libraryFile.exists())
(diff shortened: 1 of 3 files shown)


kelson42:
@sgourdas What is exactly the status here? It is mark ready to review, but still WIP and no reviewer? It's unclear.

sgourdas:
Changing to draft since getDataDirectory is not merged yet. I have finished work on the others.

kelson42:
@sgourdas If there is a blocker, please write it clearly. Now that the data directory PR are merge, does that mean this PR is ready to review?
@sgourdas Please rebase an fix conflict.

sgourdas:
Rebased with main and completed changes. Ready for review.

src/kiwixapp.cpp
```diff
@@ -150,38 +150,36 @@ void KiwixApp::newTab()
 
 QString KiwixApp::findLibraryDirectory()
 {
-  // Check for library.xml in the same directory than the executable (portable kiwix-desktop)
```
veloman-yunkan, line 153: The comment is about the commit message. Please change it to "Fixed indentation and comments"

src/kiwixapp.cpp
```diff
@@ -151,6 +151,10 @@ void KiwixApp::newTab()
 QString KiwixApp::findLibraryDirectory()
 {
     auto currentDataDir = QString::fromStdString(kiwix::removeLastPathElement(kiwix::getExecutablePath()));
+    auto portableFile = QFileInfo(currentDataDir, ".portable");
+    if (portableFile.exists())
+        return currentDataDir + "/data";
```
veloman-yunkan, line 156: Do we (intend to) provide a portable package for Windows? If yes, then the directory separator should depend on the the platform.

src/settingsmanager.cpp
```diff
@@ -9,6 +9,11 @@
 
 QString getDataDirectory()
 {
+    auto currentDataDir = QString::fromStdString(kiwix::removeLastPathElement(kiwix::getExecutablePath()));
+    auto portableFile = QFileInfo(currentDataDir, ".portable");
+    if (portableFile.exists())
+        return currentDataDir + "/data";
```
veloman-yunkan, line 15: Do we (intend to) provide a portable package for Windows? If yes, then the directory separator should depend on the the platform.

veloman-yunkan:
WIP qualifier is no longer pertinent, I guess

src/kiwixapp.cpp
sgourdas: Absolutely. Fixing...

veloman-yunkan (approved):

kelson42:
@sgourdas Sorry for the lack of feedback here, will look ASAP

kelson42 (changes requested):
I have checked finally your feature. This is on the good way, but it seems that the configuration is still take from the older "system" conf.
image

I see also no `kiwix-desktop.conf` in the `data` directory. Please verify that all configuration files from https://github.com/kiwix/kiwix-desktop/wiki/Data-and-Configuration-Files are read/write locally. Please update https://github.com/kiwix/kiwix-desktop/wiki/Data-and-Configuration-Files with a new "portable" column as well.

kelson42:
@sgourdas Any update?

sgourdas:
Sorry for the delay. I will update ASAP.

kelson42:
@sgourdas Please ensure this PR fixes #460 properly

sgourdas:
>Please update https://github.com/kiwix/kiwix-desktop/wiki/Data-and-Configuration-Files with a new "portable" column as well. 

@kelson42 what should the "portable" column have as information?

kelson42:
> >Please update https://github.com/kiwix/kiwix-desktop/wiki/Data-and-Configuration-Files with a new "portable" column as well. 
> 
> 
> 
> @kelson42 what should the "portable" column have as information? 

The relative paths where the file are read/written

sgourdas:
> > > Please update https://github.com/kiwix/kiwix-desktop/wiki/Data-and-Configuration-Files with a new "portable" column as well.
> > 
> > 
> > @kelson42 what should the "portable" column have as information?
> 
> The relative paths where the file are read/written

I have set it up so in portable mode all files are written in the `data` folder besides the executable. Since it is all the same maybe just a note under the table would be enough?

sgourdas:
@veloman-yunkan I added support also for the download directory and `.conf` file to work appropriately in portable mode.

kelson42 (approved):
To me, this PR seems to perfctly implement #1166

src/settingsmanager.h
```diff
@@ -78,4 +79,5 @@ public slots:
 };
 
 QString getDataDirectory();
+bool isPortableMode();
 #endif // SETTINGSMANAGER_H
```
veloman-yunkan, line 83: Add a blank line before `#endif`

src/settingsmanager.h
```diff
@@ -17,6 +17,7 @@ class SettingsManager : public QObject
     explicit SettingsManager(QObject *parent = nullptr);
     virtual ~SettingsManager() {};
 
+    QString getSettingsConfPath();
```
veloman-yunkan, line 20: This function doesn't need to be part of the `SettingsManager` public interface (or belong to `SettingsManager` at all). It can be defined in an unnamed namespace in `settingsmanager.cpp`.

veloman-yunkan (changes requested):
Regarding the current changes I have only a couple of insignificant proposals for improvement.

But regarding the portable functionality I have the following questions:
1. Should we allow the user to change/set the download and monitored directories in portable mode?
2. Should we allow adding ZIM files from outside the USB drive to the library in portable mode?

sgourdas:
>But regarding the portable functionality I have the following questions:
>1. Should we allow the user to change/set the download and monitored directories in portable mode?
>2. Should we allow adding ZIM files from outside the USB drive to the library in portable mode?

I was thinking about this too and how strict we should be with the portable mode. 

IMO we should not be very restrictive and let the user choose. But, with the current implementation, if the download directory is changed and mode is portable, it would be set back to the `data` folder on app restart, which I dont think is proper, but adds some complexity if we want to tackle.

veloman-yunkan:
> But, with the current implementation, if the download directory is changed and mode is portable, it would be set back to the `data` folder on app restart

How come it works like that? Isn't the download directory setting saved in the `kiwix-desktop.session` file?

veloman-yunkan (approved):
LGTM.

sgourdas:
> > But, with the current implementation, if the download directory is changed and mode is portable, it would be set back to the `data` folder on app restart
> 
> How come it works like that? Isn't the download directory setting saved in the `kiwix-desktop.session` file?

It does usually, but not in portable mode as per this PR. 

[`settingsmanager.cpp:192`](https://github.com/kiwix/kiwix-desktop/pull/1179/commits/3edee85d637948af126ce4be2a1fe785caaf9564#diff-a7b0c45be7e799343cfc8b712bc1a489f7229e1c1397d246084411f167664dbaR192)

src/settingsmanager.cpp
```diff
@@ -162,7 +188,8 @@ void SettingsManager::initSettings()
 {
     m_kiwixServerPort = m_settings.value("localKiwixServer/port", 8080).toInt();
     m_zoomFactor = m_settings.value("view/zoomFactor", 1).toDouble();
-    m_downloadDir = m_settings.value("download/dir", getDataDirectory()).toString();
+    QString dataDir = getDataDirectory();
+    m_downloadDir = isPortableMode() ? dataDir : m_settings.value("download/dir", dataDir).toString();
```
veloman-yunkan, line 192: OK, I see the answer to my question. Didn't pay proper attention to this change.

veloman-yunkan: If we decide that portable mode should be sandboxed to the USB drive, then the better approach would be disabling the download and monitored directory settings in that mode (as well as disallowing adding external ZIM files). But that's up to @kelson42 to decide.

kelson42:
>     1. Should we allow the user to change/set the download and monitored directories in portable mode?

Why not? But to me this is something different: I have created https://github.com/kiwix/kiwix-desktop/issues/1203

>     2. Should we allow adding ZIM files from outside the USB drive to the library in portable mode?

Why, not? At the end this is also a topic which is relevant for #1203 IMHO

PR #1184, opened 2024-08-21
[CI] Compile on Windows

mgautierfr:


scripts/pkg-config-wrapper.py
```diff
@@ -0,0 +1,58 @@
...
+
+def forward_to_pkg_config():
+    completeProcess = subprocess.run(["pkg-config", *sys.argv[1:]], capture_output=True, check=True, text=True)
+    return completeProcess.stdout
+   
+def search_static_lib(lib_name, search_pathes):
+    for path in search_pathes:
+        lib_path = path/f"lib{lib_name}.a"
+        if lib_path.exists():
+            return f"lib{lib_name}.a"
```
veloman-yunkan, line 35: Shouldn't you return the absolute path here as claimed in the doc comment and commit message?

scripts/pkg-config-wrapper.py
```diff
@@ -0,0 +1,58 @@
...
+# On Unix, they are in the form of "libfoo.a"
+# On Windows, "foo.lib" can also the definition of symbols in a foo.dll.
+# So you can link to "foo.lib", whatever you are doing static or dynamic linking and you are good.
+# However, meson is always creating static library as "libfoo.a"
+# 'to avoid a potential name clash with shared libraries which also generate import libraries with a lib suffix.' [1]
+# On top of that, qmake is replacing all `-lfoo` in LIBS by `foo.lib` (on Windows).
+# So at the end, we try to link with `foo.lib` but we have `libfoo.a`
+# Solution could be :
+# - Rename `libfoo.a` to `foo.lib`, but it would mean modify deps libraries on the FS
+# - Don't use LIBS and direction set QMAKE_LFLAGS but we would have to handle different command line option format
```
veloman-yunkan, line 13: direction → directly (also fix in the commit message)

mgautierfr: Done in the fixup commit.
Commit message already fixed

scripts/pkg-config-wrapper.py
mgautierfr: Indeed, I made it with few try and error and few wrong modification on kiwix-desktop.pro side.

It seems that we mainly have to replace `-lfoo` with the static library name, whatever if it is absolute path or not (assuming we have a `-Lpath` too).

I've keep the absolute path as it is what meson is doing.

kelson42 (approved):
LGTM

veloman-yunkan (commented):
LGTM. Formal approval will be granted after the fixup commit is gone.

mgautierfr:
> ...after the fixup commit is gone.

Done

veloman-yunkan:
> > ...after the fixup commit is gone.
> 
> Done

Just for your information - it was done in a slightly wrong way (applied to the wrong commit)

PR #1198, opened 2024-09-07
Introduce Scroll Buttons to Replace Ones from Qt's Built-in TabBar.

### LINKED ISSUE #157: What is the strategy to deal with high density screens
How do  you secure that the application layout is displayed properly on high density screens on Windows & Linux?

ShaopengLin:
Fix #157
This PR solves the scroll button issue using the second approach in [this comment](https://github.com/kiwix/kiwix-desktop/issues/157#issuecomment-2171870239). The comment has been stale for a long time and I am using my best judgment to select the most intuitive approach.

Changes:
- The original scroll buttons are now hidden.
- Added refactor to new tab button. There is now a persistent new tab button on the side if the new tab 'tab' is hidden from view. Due to technical issues, the prior new tab button prevents this change from working properly.
- Introduce new scroll buttons on each side of the tab bar.
- Buttons scroll by moving the current tab to their respective direction.
- Respective buttons are hidden once the first or the new tab button tab is fully visible.


kelson42:
@ShaopengLin The computation deciding when the arrow should be displayed or not seems buggy. See here:
image

All the tabs are visibles, so now arrow should be visible. 

There is also some kind of strange blue artefact on the left arrow:
image

ShaopengLin:
@kelson42 I have fixed this issue with the need to introduce a refactor to the new tab 'tab'. After I fixed the issue you mentioned [here](https://github.com/kiwix/kiwix-desktop/pull/1198#issuecomment-2336833303), I could no longer scroll to the new tab 'tab' due to internal code and Qt limitations. With some digging, the only solution that I have is to introduce a persistent new tab button when the new tab 'tab' is hidden from view. This no longer required us to scroll to the new tab 'tab'. Could you re-test the UI?

The blue tear is a Qt built-in feature. I have disabled it.

kelson42:
By keeping adding new tabs via mouse double-click on links, I achieve a situation where I have a lots of tabs overflowed, but the right arrow is still not displayed:
image

ShaopengLin:
@kelson42 Should be fixed. Though I know where in the code that could have caused this, I don't seem to have the ability to double click on a link to open it in a new tab. The double click simply opens the link in the current tab for me. Is there a trick to doing this?

kelson42 (approved):
LGTM from user perspective
@veloman-yunkan  ready for code review

kelson42:
> @kelson42 Should be fixed. Though I know where in the code that could have caused this, I don't seem to have the ability to double click on a link to open it in a new tab. The double click simply opens the link in the current tab for me. Is there a trick to doing this?

double-click with the mouse middle button.

veloman-yunkan (commented):
I starting reviewing the changes and testing the code without reading the prehistory of this PR. Then I noticed the inconsistency between the terms used and the actual functionality, wrote a few comments but then I felt like there might be some reason for it, paused the review and read the long unanswered comment in the issue.

My feeling is that the scrolling buttons should just scroll (and, hopefully, that can be achieved with [`QTabBar::scroll()`](https://doc.qt.io/qt-5/qwidget.html#scroll)). I also think that the buttons should not disappear when you fully scroll to the respective end.

ShaopengLin:
@veloman-yunkan IMO there aren't any intuitive solutions to really "scroll". These buttons should be good enough replacements for those who do not know keyboard shortcuts to switch tabs or do not have a mouse scroller. If we figure out how to actually "scroll", those can be easily adapted. I believe we can get this merged first if you are also stuck on the scrolling behavior.

@kelson42 could you verify the UI again?

kelson42 (approved):
To my opinion, from a user perspective, it seems to work perfectly!
@veloman-yunkan Requireing code approval for merging,

src/mainwindow.cpp
```diff
@@ -109,6 +114,34 @@ void MainWindow::showTabAndTop() {
...
+    QRect tabBarRect = getTabBar()->rect();
+    QRect newButtonTabRect = tabBar->tabRect(tabBar->count() - 1);
+
+    /* Decision is made at half way of the new button tab for smoothness */
+    newButtonTabRect.setWidth(newButtonTabRect.width() / 2);
+    bool newTabVisible = tabBarRect.contains(newButtonTabRect);
+    if (mp_ui->newTabSideButton->isHidden())
+        mp_ui->newTabSideButton->setVisible(!newTabVisible);
+    else
+        mp_ui->newTabSideButton->setHidden(newTabVisible);
```
veloman-yunkan, line 129: Can't the `if/else` statement be reduced to `mp_ui->newTabSideButton->setVisible(!newTabVisible);`?

src/mainwindow.cpp
```diff
@@ -109,6 +114,34 @@ void MainWindow::showTabAndTop() {
...
+
+    QRect firstTabRect = tabBar->tabRect(0);
+    QRect lastRealTabRect = tabBar->tabRect(tabBar->count() - 2);
```
veloman-yunkan, line 132: Better use `tabBar->realTabCount() - 1`

src/mainwindow.cpp
```diff
@@ -109,6 +114,34 @@ void MainWindow::showTabAndTop() {
...
+
+    bool firstVisible = tabBarRect.contains(firstTabRect);
+    bool lastVisible = tabBarRect.contains(lastRealTabRect);
+    bool noOverFlow = firstVisible && lastVisible;
+
+    mp_ui->prevTabButton->setHidden(noOverFlow);
+    mp_ui->nextTabButton->setHidden(noOverFlow);
+
+    mp_ui->prevTabButton->setDisabled(firstVisible);
+    mp_ui->nextTabButton->setDisabled(lastVisible);
```
veloman-yunkan, line 142: This makes sense only if the stated purpose of the `prevTabButton` and `nextTabButton` is to scroll the tabbar (though with an undesirable but hard-to-eliminate side-effect of changing the active tab). However if the users perceive the function of those buttons as switching to an adjacent tab via GUI, then the fact that that operation is disabled before the first/last tab has been reached becomes controversial.

src/mainwindow.cpp
ShaopengLin: Done

src/mainwindow.cpp
ShaopengLin: Done, made method public.

veloman-yunkan (approved):
One last tiny concern of the same kind as in https://github.com/kiwix/kiwix-desktop/pull/1198#discussion_r1760800168 (caused by the scrolling-vs-switch-to-adjacent-tab confusion) is that one is deprived of the next/previous tab buttons unless they have enough tabs open. However I don't consider it as a blocker. Just want to draw @kelson42's attention to the small change in behavior (caused by my feedback) since his last testing.

kelson42:
> One last tiny concern of the same kind as in [#1198 (comment)](https://github.com/kiwix/kiwix-desktop/pull/1198#discussion_r1760800168) (caused by the scrolling-vs-switch-to-adjacent-tab confusion) is that one is deprived of the next/previous tab buttons unless they have enough tabs open. However I don't consider it as a blocker. Just want to draw @kelson42's attention to the small change in behavior (caused by my feedback) since his last testing.

IMHO works properly like this. I will merge.

PR #1208, opened 2024-09-13
Stop library from prioritizing executable directory

### LINKED ISSUE #460: Kiwix Desktop does not ignore other libraries when local library.xml is available

sgourdas:
This removes the prioritization of the executable's directory for the library of the app, as now the "portable" functionality has been implemented in a different way (#1166).

Fixes #460

### DIFF
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -142,30 +142,22 @@ void KiwixApp::newTab()
 
 QString KiwixApp::findLibraryDirectory()
 {
-    auto currentDataDir = QString::fromStdString(kiwix::removeLastPathElement(kiwix::getExecutablePath()));
-    
+    auto dataDir = getDataDirectory();
     if (isPortableMode())
-        return currentDataDir + QDir::separator() + "data";
+        return dataDir;
 
-    // Check for library.xml in the same directory as the executable.
-    auto libraryFile = QFileInfo(currentDataDir, "library.xml");
+    auto libraryFile = QFileInfo(dataDir, "library.xml");
     if (libraryFile.exists())
-        return currentDataDir;
-
-    // Check for default dataDirectory.
-    currentDataDir = getDataDirectory();
-    libraryFile = QFileInfo(currentDataDir, "library.xml");
-    if (libraryFile.exists())
-        return currentDataDir;
+        return dataDir;
 
     // Check if this is a pre-release version with wrong directory.
-    auto oldDataDir = QDir(currentDataDir);
+    auto oldDataDir = QDir(dataDir);
     oldDataDir.cdUp();
     libraryFile = QFileInfo(oldDataDir, "library.xml");
     if (libraryFile.exists()) {
         // We have to move all zim files and xml file to the new dataDir
         for (auto& fileInfo: oldDataDir.entryInfoList({"*.zim", "library*.xml"})) {
-            auto newFileInfo = QFileInfo(currentDataDir, fileInfo.fileName());
+            auto newFileInfo = QFileInfo(dataDir, fileInfo.fileName());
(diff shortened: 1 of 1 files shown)


src/kiwixapp.cpp
```diff
@@ -147,14 +147,9 @@ QString KiwixApp::findLibraryDirectory()
     if (isPortableMode())
         return currentDataDir + QDir::separator() + "data";
 
-    // Check for library.xml in the same directory as the executable.
-    auto libraryFile = QFileInfo(currentDataDir, "library.xml");
-    if (libraryFile.exists())
-        return currentDataDir;
-
     // Check for default dataDirectory.
     currentDataDir = getDataDirectory();
```
veloman-yunkan, line 151: Now, taking into account the current implementation of `getDataDirectory()`, the beginning of this function's body can be simplified as follows:

```c++
{
    auto currentDataDir = getDataDirectory();
    if (isPortableMode())
        return currentDataDir;

   ...
```

sgourdas: `getDataDirectory()`'s fallback/final evaluation is the current directory. I do not think this is ideal for the use we need it.

Am I missing something?

sgourdas: Maybe I should refactor the variable names which were like this from the libkiwix logic, to have the code be more understandable as well?

veloman-yunkan: When `isPortableMode() == true`, `getDataDirectory()` returns the `data` subdirectory of executable's directory, doesn't it?

src/kiwixapp.cpp
```diff
@@ -142,25 +142,19 @@ void KiwixApp::newTab()
...
-        return currentDataDir + QDir::separator() + "data";
-
-    // Check for default dataDirectory.
-    currentDataDir = getDataDirectory();
-    auto libraryFile = QFileInfo(currentDataDir, "library.xml");
+    auto dataDir = getDataDirectory();
+    auto libraryFile = QFileInfo(dataDir, "library.xml");
     if (libraryFile.exists())
-        return currentDataDir;
+        return dataDir;
```
veloman-yunkan, line 148: This will fail to immediately return the portable data directory in portable mode if for some reason the library.xml file is missing (as far as I understand that's what happens during the first launch). Though the continued execution of this function will most likely have the same intended effect it is better to change this piece of code to:

```c++
    auto dataDir = getDataDirectory();
    if (isPortableMode())
        return dataDir;
```

veloman-yunkan (changes requested):

src/kiwixapp.cpp
sgourdas: Sorry to postpone this, but why would that be a beneficial change? Isn't that just unneeded code?

Based on the current commit's implementation, the function would work as intended on first run, by returning on the last `return`

veloman-yunkan: Mostly yes. But I am confident that you can imagine the unlikely circumstances under which the heuristics implemented by the code below that line may be triggered wrongfully. Therefore the early return is a safety measure.

sgourdas: Fair point. I will proceed then.

veloman-yunkan (approved):

sgourdas:
@kelson42 ready for merge.

PR #1209, opened 2024-09-13
Monitor data directory on portable mode

### LINKED ISSUE #461: Simplification of portable USB creation
For people wanting to distribute USB flash drives with Kiwix AND ZIM files, we have a working process. The process is documented at https://github.com/kiwix/kiwix-desktop/wiki/Kiwix-desktop-on-an-USB-Drive-(Portable-version).

The problem is that this is still a bit complicated. This ticket requires to simplify it and has this new workflow proposal: The local `library.xml` would be automatically based on ZIM files available at the same level like the binary if:
* There is no local `library.xml` already
* There is no other `library.xml` which include known local ZIM files.

sgourdas:
This simplifies the creation of a portable USB by automatically monitoring the data directory in portable mode. This way we allow the user to just place needed .zim files in the data folder and the app will handle the rest.

- Adds also immutability to the monitor and download paths through settings on portable mode

Fixes #461

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -166,6 +166,7 @@
     "import-reading-list-error": "An error has occured during import of the reading list.",
     "disable-sandbox": "Application was launched from a network drive. This is known to cause compatibility issues due to the sandbox. Do you want to take the risks and disable it?",
     "save-page-as": "Save As...",
+    "portable-disabled-tooltip": "Function disabled in portable mode",
     "scroll-next-tab": "Scroll to next tab",
     "scroll-previous-tab": "Scroll to previous tab"
 }
--- a/src/settingsview.cpp
+++ b/src/settingsview.cpp
@@ -18,6 +18,14 @@ namespace
         }
         return input;
     }
+
+    void disableInPortableMode(QWidget* w)
+    {
+        w->setEnabled(false); 
+        QString disabledStyle = "color: gray;";
+        w->setStyleSheet(disabledStyle);
+        w->setToolTip(gt("portable-disabled-tooltip"));
+    }
 }
 
 SettingsView::SettingsView(QWidget *parent)
@@ -62,6 +70,12 @@ SettingsView::SettingsView(QWidget *parent)
     ui->monitorHelp->setToolTip(gt("monitor-directory-tooltip"));
     ui->moveToTrashLabel->setText(gt("move-files-to-trash"));
     ui->reopenTabLabel->setText(gt("open-previous-tabs-at-startup"));
+    if(isPortableMode()) {
+        disableInPortableMode(ui->browseButton);
+        disableInPortableMode(ui->resetButton);
+        disableInPortableMode(ui->monitorBrowse);
+        disableInPortableMode(ui->monitorClear);
+    }
 #if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
     ui->line_5->hide();
     ui->moveToTrashLabel->hide();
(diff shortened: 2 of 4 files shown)


kelson42 (approved):
LGTM from user perspective

kelson42:
@sgourdas Might that be that settings UI about monitoring directory should be greyed, as any value will be overriden at start by the code?

veloman-yunkan (approved):

sgourdas:
@kelson42 sure, but should that be the case only for the monitor directory?

kelson42:
@sgourdas Yes, I think so... because whatever this settings is set it will be reseted at start right? For the other settings, the user can change them and they will be kept!

sgourdas:
> @sgourdas Yes, I think so... because whatever this settings is set it will be reseted at start right? For the other settings, the user can change them and they will be kept!

No, that is not the case at the moment per this commit. Monitor **and** download directory are currently hardcoded to the data directory on **every portable start**.

What would the desired behavior be?

kelson42:
> > @sgourdas Yes, I think so... because whatever this settings is set it will be reseted at start right? For the other settings, the user can change them and they will be kept!
> 
> 
> 
> @kelson42 no, that is not the case at the moment per this commit. Monitor **and** download directory are currently hardcoded to the data directory on **every portable start**.
> 
> 
> 
> What would the desired behavior be?

Then I guess both should be greyed in portable mode (only).

sgourdas:
> > > @sgourdas Yes, I think so... because whatever this settings is set it will be reseted at start right? For the other settings, the user can change them and they will be kept!
> > 
> > 
> > @kelson42 no, that is not the case at the moment per this commit. Monitor **and** download directory are currently hardcoded to the data directory on **every portable start**.
> > What would the desired behavior be?
> 
> Then I guess both should be greyed in portable mode (only).

@kelson42 I added immutability on these two in portable mode. It includes grayed out text, disabled buttons and tooltip to explain the disabling.

kelson42:
@sgourdas LGTM, please rebase so we can proceed with merging.

resources/i18n/en.json
```diff
@@ -165,5 +165,6 @@
     "import-reading-list": "Import reading list",
     "import-reading-list-error": "An error has occured during import of the reading list.",
     "disable-sandbox": "Application was launched from a network drive. This is known to cause compatibility issues due to the sandbox. Do you want to take the risks and disable it?",
+    "portable-disabled-tooltip": "Function disabled in portable mode",
```
veloman-yunkan, line 168: qqq.json must be updated too

src/settingsview.cpp
```diff
@@ -62,6 +62,21 @@ SettingsView::SettingsView(QWidget *parent)
...
+        ui->monitorClear->setEnabled(false); 
+        QString disabledStyle = "color: gray;";
+        ui->browseButton->setStyleSheet(disabledStyle);
+        ui->resetButton->setStyleSheet(disabledStyle);
+        ui->monitorBrowse->setStyleSheet(disabledStyle);
+        ui->monitorClear->setStyleSheet(disabledStyle);
+        ui->browseButton->setToolTip(gt("portable-disabled-tooltip"));
+        ui->resetButton->setToolTip(gt("portable-disabled-tooltip"));
+        ui->monitorBrowse->setToolTip(gt("portable-disabled-tooltip"));
+        ui->monitorClear->setToolTip(gt("portable-disabled-tooltip"));
```
veloman-yunkan, line 78: Please rewrite this fragment in a shorter way after introducing the following helper function in an unnamed namespace

```c++
void disableInPortableMode(QWidget* w)
{
        w->setEnabled(false); 
        QString disabledStyle = "color: gray;";
        w->setStyleSheet(disabledStyle);
        w->setToolTip(gt("portable-disabled-tooltip"));
}
```

veloman-yunkan (changes requested):

veloman-yunkan (approved):

Issue #1199, opened 2024-09-08
Can not compile anymore with Qt6


sgourdas:
Running `sudo make install` on latest libkiwix build outputs:

```
[...]

PR #1212, opened 2024-09-14
Corrected version for checkStateChanged

ShaopengLin:
Follow up for #1199

The `checkStateChanged` method is introduced in Qt 6.7 instead of 6.0, as per the [documentation](https://doc.qt.io/qt-6/qcheckbox.html):

> (since 6.7) void	[checkStateChanged](https://doc.qt.io/qt-6/qcheckbox.html#checkStateChanged)(Qt::CheckState state)


ShaopengLin:
@kelson42 You can merge this since. I don't believe this needs any code review at all. This fixes the compilation error you mentioned [here](https://github.com/kiwix/kiwix-desktop/pull/1143#issuecomment-2350979515).

PR #1213, opened 2024-09-16
libkiwix now identifies itself as libkiwix

veloman-yunkan:
Depends on kiwix/libkiwix#1133


kelson42 (changes requested):
Not sure why, but CI does not pass.

veloman-yunkan:
> Not sure why, but CI does not pass.

That was tricky, but now CI passes.

Issue #315, opened 2019-11-29
Implement multizim search/suggestion properly


kelson42:
And display the results.

Smarre:
How hard would it be to implement this feature? If it’s easy enough, I probably could take a look at it.

kelson42:
Blocked by https://github.com/openzim/libzim/issues/932

Issue #42, opened 2018-10-09
Implement table of content


kelson42:
This is a really old feature request, for Kiwix Desktop 1.0 too.

See this old ticket: https://sourceforge.net/p/kiwix/feature-requests/1000/

kelson42:
The strategy here is:
* Implement ZIM API to retrieve TOC (here is the ticket for MWoffliner for example https://github.com/openzim/mwoffliner/issues/596)
* Implement client feature in Kiwix Desktop

Jaifroid:
@knowledge-is-power Just to be clear, the ToC is not in the Wikipedia archives (not even hidden) so it needs to be constructed dynamically from the headers in the article. This isn't too hard, and some work was done on it in [this PR](https://github.com/openzim/mwoffliner/pull/1237), but then stalled. Reader software can implement a generic ToC if an article has H1, H2, H3 etc., but it's probably best done inside the ZIM.

Navboxes, etc., are another matter. These _are_ mostly in the Wikipedia ZIMs, but are hidden by the mobile style, so have to be force-unhidden (unless desktop CSS can be applied).

Jaifroid:
@knowledge-is-power Just so you're aware, Kiwix JS isn't just a PWA, we also have installable packages for various platforms, and packages for Linux as well as Windows. See http://kiwix.github.io/kiwix-js-windows/kiwix-js-electron.html. Although it's not the full-featured Desktop app, it works very well with Wikimedia archives (and many others too). It's good to have options!

kelson42:
Reminder: please one topic per ticket, here you mixed two topics which have nothing in common:
* box collapsing
* TOC

For the rest, the way forward is pretty known, we just need to have the development resources. It's an old problem, but this is on the roadmap and this will be fixed/implemented at some point.

knowledge-is-power:
@kelson42  yeah I thought both issues were part of the same problem that's why I made a single ticket, sorry about that. now that its clear to me that they are unrelated, is it okay if i open a dedicated ticket for wiki navboxes/templates? (cuz this ticket is dedicated to ToC issue)

juuz0:
I created a ticket here #869

ShaopengLin:
@kelson42 Javascript Injection works fine. QWebEngineScript allows us to inject Javascript in separate files and we can also write the CSS styles over there. I can create a Mock up and you can advise on how to adjust the presentation (This is injected so it will feel foreign and we need a balance).
[...]

PR #1218, opened 2024-09-28
UI for selecting the book to be used for title search

### LINKED ISSUE #413: Tune search suggestions overlay
Here is how it should looks like (from the mockup):

image

Without implementing everything now, I want:
* To have a proper font (size)
* The overall look&file of the widget
* The icon of the ZIM in which the content is found
* "Endless" suggestion

ShaopengLin:
Fixes (part of) #315 

Relies on #1189

Changes:
- Added MultiZim button to SearchBar. Ctrl + Shift + L to open popup.
- Searching is now allowed on library, settings, and blank tabs.
- Popup shows a list of Zims that could be searched. The selection is singular as we do not have *true multi-zim search (yet).
- Users can now search a different zim than the current tab.
- Added placeholder UI for 'select-all' button. Hidden for now until *true multi-zim search is implemented.

### DIFF
--- a/resources/i18n/qqq.json
+++ b/resources/i18n/qqq.json
@@ -180,5 +180,6 @@
 	"portable-disabled-tooltip": "Tooltip used to explain disabled components in the portable version.",
 	"scroll-next-tab": "Represents the action of scrolling to the next tab of the current tab which toward the end of the tab bar.",
 	"scroll-previous-tab": "Represents the action of scrolling to the previous tab of the current tab which toward the start of the tab bar.",
-	"kiwix-search": "Title text for a list of search results, which notes to the user those are from Kiwix's Search Engine"
+	"kiwix-search": "Title text for a list of search results, which notes to the user those are from Kiwix's Search Engine",
+	"search-options": "The term for a collection of additional search filtering options to help users narrow down search results."
 }
--- a/src/kiwixapp.cpp
+++ b/src/kiwixapp.cpp
@@ -438,6 +438,8 @@ void KiwixApp::createActions()
     CREATE_ACTION_SHORTCUT(ToggleTOCAction, gt("table-of-content"), QKeySequence(Qt::CTRL | Qt::SHIFT | Qt::Key_1));
     HIDE_ACTION(ToggleTOCAction);
 
+    CREATE_ACTION_ICON_SHORTCUT(OpenMultiZimAction, "filter", gt("search-options"), QKeySequence(Qt::CTRL | Qt::SHIFT | Qt::Key_L));
+
     CREATE_ACTION_ONOFF_ICON_SHORTCUT(ToggleReadingListAction, "reading-list-active", "reading-list", gt("reading-list"), QKeySequence(Qt::CTRL | Qt::Key_B));
 
     CREATE_ACTION(ExportReadingListAction, gt("export-reading-list"));
--- a/src/multizimbutton.cpp
+++ b/src/multizimbutton.cpp
@@ -1,0 +1,75 @@
+#include <QListWidget>
+#include <QMenu>
+#include <QWidgetAction>
+#include <QButtonGroup>
+#include <QRadioButton>
+#include "kiwixapp.h"
+#include "multizimbutton.h"
+
+MultiZimButton::MultiZimButton(QWidget *parent) :
+    QToolButton(parent), 
+    mp_buttonList(new QListWidget),
+    mp_radioButtonGroup(new QButtonGroup(this))
+{
+    setMenu(new QMenu(this));
+    setPopupMode(QToolButton::InstantPopup);
+
+    const auto popupAction = new QWidgetAction(menu());
+    popupAction->setDefaultWidget(mp_buttonList);
+    menu()->addAction(popupAction);
+}
+
+void MultiZimButton::updateDisplay()
+{
+    mp_buttonList->clear();
+    for (const auto& button : mp_radioButtonGroup->buttons())
+        mp_radioButtonGroup->removeButton(button);
+
+    const auto library = KiwixApp::instance()->getLibrary();
+    for (const auto& bookId : library->getBookIds())
+    {
+        try
+        {
+            library->getArchive(bookId);
+        } catch (...) { continue; }
+
+        const QString bookTitle = QString::fromStdString(library->getBookById(bookId).getTitle());
+        const QIcon zimIcon = library->getBookIcon(bookId);
+        
+        const auto item = new QListWidgetItem();
+        item->setData(Qt::UserRole, bookId);
+        item->setData(Qt::DisplayRole, bookTitle);
+
+        const auto radioBt = new QRadioButton(bookTitle);
+        radioBt->setIcon(zimIcon);
+        mp_radioButtonGroup->addButton(radioBt);
+
+        mp_buttonList->addItem(item);
+        mp_buttonList->setItemWidget(item, radioBt);
+    }
+
+    mp_buttonList->sortItems();
+
+    /* Display should not be used other than for sorting. */
+    for (int i = 0; i < mp_buttonList->count(); i++)
+        mp_buttonList->item(i)->setData(Qt::DisplayRole, QVariant());
+
+    setDisabled(mp_buttonList->model()->rowCount() == 0);
+
+    const auto firstWidget = mp_buttonList->itemWidget(mp_buttonList->item(0));
+    if (const auto firstBt = qobject_cast<QRadioButton *>(firstWidget))
+        firstBt->setChecked(true);
+}
+
+QStringList MultiZimButton::getZimIds() const
+{
+    QStringList idList;
+    for (int row = 0; row < mp_buttonList->count(); row++)
+    {
+        const auto listItem = mp_buttonList->item(row);
+        const auto radioBt = qobject_cast<QRadioButton *>(mp_buttonList->itemWidget(listItem));
+        if (radioBt && radioBt->isChecked())
+            idList.append(listItem->data(Qt::UserRole).toString());
+    }
+    return idList;
+}
--- a/src/multizimbutton.cpp
+++ b/src/multizimbutton.cpp
@@ -19,6 +19,11 @@ MultiZimButton::MultiZimButton(QWidget *parent) :
     const auto popupAction = new QWidgetAction(menu());
     popupAction->setDefaultWidget(mp_buttonList);
     menu()->addAction(popupAction);
+
+    connect(mp_buttonList, &QListWidget::currentRowChanged, this, [=](int row){
+        if (const auto widget = getZimWidget(row))
+            widget->getRadioButton()->setChecked(true);
+    });
 }
 
 void MultiZimButton::updateDisplay()
@@ -72,8 +77,7 @@ void MultiZimButton::updateDisplay()
     setDisabled(mp_buttonList->model()->rowCount() == 0);
 
     mp_buttonList->scrollToTop();
-    if (const auto firstWidget = getZimWidget(0))
-        firstWidget->getRadioButton()->setChecked(true);
+    mp_buttonList->setCurrentRow(0);
 }
 
 QStringList MultiZimButton::getZimIds() const
--- a/src/multizimbutton.cpp
+++ b/src/multizimbutton.cpp
@@ -28,6 +28,9 @@ void MultiZimButton::updateDisplay()
         mp_radioButtonGroup->removeButton(button);
 
     const auto library = KiwixApp::instance()->getLibrary();
+    const auto view = KiwixApp::instance()->getTabWidget()->currentWebView();
+    QListWidgetItem* currentItem = nullptr;
+    QIcon currentIcon;
     for (const auto& bookId : library->getBookIds())
     {
         try
@@ -42,11 +45,25 @@ void MultiZimButton::updateDisplay()
         item->setData(Qt::UserRole, bookId);
         item->setData(Qt::DisplayRole, bookTitle);
 
+        if (view && view->zimId() == bookId)
+        {
+            currentItem = item;
+            currentIcon = zimIcon;
+            continue;
+        }
+
         mp_buttonList->addItem(item);
         setItemZimWidget(item, bookTitle, zimIcon);
     }
 
     mp_buttonList->sortItems();
+    if (currentItem)
+    {
+        mp_buttonList->insertItem(0, currentItem);
+
+        const auto title = currentItem->data(Qt::DisplayRole).toString();
+        setItemZimWidget(currentItem, "*" + title, currentIcon);
+    }
 
     /* Display should not be used other than for sorting. */
     for (int i = 0; i < mp_buttonList->count(); i++)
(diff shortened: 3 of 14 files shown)


kelson42:
@ShaopengLin I guess this is the next one to complete the work on the new searchbar?

ShaopengLin:
@kelson42 This is the next change.

@veloman-yunkan Depending on how #1230 discussion goes we can combine them here.

veloman-yunkan:
> @veloman-yunkan Depending on how #1230 discussion goes we can combine them here.

@ShaopengLin This PR is about Multi-ZIM search which has nothing to do with #1230. Please provide a bugfix for #1230 as a separate PR.

ShaopengLin:
@veloman-yunkan The code ios ready for review.

veloman-yunkan:
Before looking at the code I tested the new feature and here is my feedback:

1. The most obvious new possibility that could be enabled by the new button is not working - searching while on the library page.
2. The list of ZIMs doesn't seem to be sorted in any particular order.
3. The new UI (like probably a lot of our old UI) requires a mouse - it cannot be used via keyboard only
4. Are there plans to support filtering the list of ZIMs shown in the ZIM selector for search? Currently it shows the full list of ZIMs. May be it should be in sync with the library ZIM file filter?

veloman-yunkan:
> Fixes (part of) #413

#413 is unrelated to multi-zim search (and probably can already be closed). Doesn't this PR rather address part of #315?

ShaopengLin:
@veloman-yunkan Kelson decided to merge the two together back here https://github.com/kiwix/kiwix-desktop/issues/413#issuecomment-2283072087. What we have here isn't true multi-Zim, since we can only search one zim at a time.

1. That particular problem has been there since the search bar introduction. Searching also doesn't work on blank tabs, which I believe was mentioned in some issues but not sure where. Likely need to do this fix separately.

2. Sure I can sort it. Currently, it's ordered in the same way it appears in the `Local Files` library.

3. I will investigate and re-commit.

4. One problem I see is there isn't a strong correlation for users to think library tab filters are applied to multi-zim. It can be hard to comprehend and create bad UX. Adding filters will likely need to be done within multi-zim.

veloman-yunkan:
> 1. That particular problem has been there since the search bar introduction. Searching also doesn't work on blank tabs, which I believe was mentioned in some issues but not sure where. Likely need to do this fix separately.

Then it made sense because search was supposed to work only for the current book and there was no current book in the library view. Now that you can select the book to search in via dedicated UI such behavior is no longer justified.

veloman-yunkan:
> @veloman-yunkan Kelson decided to merge the two together back here [#413 (comment)](https://github.com/kiwix/kiwix-desktop/issues/413#issuecomment-2283072087). What we have here isn't true multi-Zim, since we can only search one zim at a time.

#413 and #315 are about completely different features. This PR is definitely a step toward #315 rather than the final touch of #413.

ShaopengLin:
@kelson42 Should we mark this PR for #315 instead?

kelson42:
> @kelson42 We should mark this PR for #315 instead if we would like to keep that Issue.

Sorry but I'm really lost now about what fixes what. As long as the issues are implemented at the end, I'm fine with it.

ShaopengLin:
@veloman-yunkan 
- Keyboard controls should be available for the MultiZim popup now. 
- searching in library page is now working.
- Sorted by alphabetic order.

src/multizimbutton.cpp
```diff
@@ -44,12 +44,8 @@ void MultiZimButton::update_display()
         item->setData(Qt::UserRole, bookId);
         item->setData(Qt::DisplayRole, bookTitle);
 
-        const auto radioBt = new QRadioButton(bookTitle);
-        radioBt->setIcon(zimIcon);
-        mp_radioButtonGroup->addButton(radioBt);
-
         mp_buttonList->addItem(item);
-        mp_buttonList->setItemWidget(item, radioBt);
+        setItemZimWidget(item, bookTitle, zimIcon);
```
veloman-yunkan, line 78: As of this commit `setItemZimWidget()` looks artificial. I would replace it with a fatter function `MultiZimButton::addBook(QString bookId)` and maybe move all code in the loop body into it. But let's see if further evolution of this PR justifies your implementation.

src/multizimbutton.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    for (int row = 0; row < mp_buttonList->count(); row++)
+    {
+        const auto listItem = mp_buttonList->item(row);
+        const auto radioBt = qobject_cast<QRadioButton *>(mp_buttonList->itemWidget(listItem));
+        if (radioBt && radioBt->isChecked())
+            return listItem->data(Qt::UserRole).toString();
+    }
+    return QString{};
+}
```
veloman-yunkan: Though this is not really a fully fledged implementation of multi-ZIM search, is there any reason why this function cannot be introduced in its final form `QStringList MultiZimButton::getZimIds() const`? Only at the point of usage will you assert that the result contains exactly one element.

src/multizimbutton.cpp
veloman-yunkan: I see, your way of bringing the current ZIM to top stands in the way of such proposal.

src/multizimbutton.cpp
```diff
@@ -36,6 +37,8 @@ void MultiZimButton::update_display()
     const auto view = KiwixApp::instance()->getTabWidget()->currentWebView();
     QListWidgetItem* currentItem = nullptr;
     QIcon currentIcon;
+    const int paddigTopBot = CSS::MultiZimButton::QListWidget::paddingVertical * 2;
```
veloman-yunkan, line 51: paddig (typos can cause some headache when you try to look up some code by memory and then it turns out that it contained a typo you never noticed that's why I pay attention to them).

resources/i18n/qqq.json
```diff
@@ -181,5 +181,6 @@
 	"scroll-next-tab": "Represents the action of scrolling to the next tab of the current tab which toward the end of the tab bar.",
 	"scroll-previous-tab": "Represents the action of scrolling to the previous tab of the current tab which toward the start of the tab bar.",
 	"kiwix-search": "Title text for a list of search results, which notes to the user those are from Kiwix's Search Engine",
-	"multi-zim-search": "Represents the action of searching suggrestion results from one or more ZIM files in the library."
+	"multi-zim-search": "Represents the action of searching suggrestion results from one or more ZIM files in the library.",
+	"select-all": "Represents the action of selecting all ZIM in the library to be search targets for suggestions."
```
veloman-yunkan, line 185: The too specific description of this message id suggests that it should be renamed to "select-all-zims". Or you can leave the message id as is but remove any reference to ZIMs from its description (so that it can be reused in other contexts too).

src/multizimbutton.cpp
```diff
@@ -10,16 +10,27 @@
...
     setPopupMode(QToolButton::InstantPopup);
     setDefaultAction(KiwixApp::instance()->getAction(KiwixApp::OpenMultiZimAction));
     connect(this, &QToolButton::triggered, this, &MultiZimButton::showMenu);
 
-    const auto popupAction = new QWidgetAction(menu());
-    popupAction->setDefaultWidget(mp_buttonList);
-    menu()->addAction(popupAction);
+    /* TODO: connect with list widget after 'real' multi-zim implementation */
+    mp_selectAllButton->setDisabled(true);
+    mp_selectAllButton->hide();
```
veloman-yunkan, line 23: Frankly, I don't think that it makes sense to include an invisible "Select all" button in this PR.

src/multizimbutton.cpp
```diff
@@ -165,9 +165,21 @@ ZimItemWidget::ZimItemWidget(QString text, QIcon icon, QWidget *parent) :
...
+    const int textLength = labelWidth - elideMarkerLength;
+    QString elidedText = metrics.elidedText(text, Qt::ElideRight, textLength);
+    if (elidedText != text)
+    {
+        /* Remove built-in elide marker */
+        elidedText.chop(1);
+        textLabel->setText(elidedText.trimmed() + "(...)");
+    }
+    else
+        textLabel->setText(text);
```
veloman-yunkan, line 182: Is there code duplication with custom elision of suggestions?

veloman-yunkan (changes requested):

src/multizimbutton.h
```diff
@@ -0,0 +1,25 @@
...
+class QButtonGroup;
+
+class MultiZimButton : public QToolButton {
+    Q_OBJECT
+
+public:
+    MultiZimButton(QWidget *parent = nullptr);
+
+public slots:
+    void update_display();
```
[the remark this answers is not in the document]
ShaopengLin: Its technically a legacy naming from BookmarkButton. I can change it.

src/multizimbutton.cpp
ShaopengLin: Before the if statement yes. The handling of elide is different since Qt handles LineEdit differently among other text displaying widget (quiet a pain)

src/multizimbutton.cpp
ShaopengLin: I believe Kelson just wants this UI to be complete before the actual multi-Zim is implemented.

src/multizimbutton.cpp
ShaopengLin: Done

resources/i18n/qqq.json
ShaopengLin: Done

src/multizimbutton.cpp
ShaopengLin: Done

src/multizimbutton.cpp
ShaopengLin: Done

src/multizimbutton.h
```diff
@@ -0,0 +1,25 @@
...
+    QString getZimId() const;
+
+private:
+    QListWidget* mp_buttonList;
+    QButtonGroup* mp_radioButtonGroup;
+    QString m_zimId;
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/multizimbutton.cpp
```diff
@@ -0,0 +1,76 @@
...
+void MultiZimButton::update_display()
+{
+    mp_buttonList->clear();
+    for (const auto& button : mp_radioButtonGroup->buttons())
+        mp_radioButtonGroup->removeButton(button);
+
+    const auto library = KiwixApp::instance()->getLibrary();
+    for (const auto& bookId : library->getBookIds())
+    {
+        std::shared_ptr<zim::Archive> archive;
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.h
```diff
@@ -77,12 +78,14 @@ class SearchBar : public QToolBar {
 public:
     SearchBar(QWidget *parent = nullptr);
     SearchBarLineEdit& getLineEdit() { return m_searchBarLineEdit; };
+    MultiZimButton& getMultiZimButtom() { return m_multiZimButton; };
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/multizimbutton.h
```diff
@@ -0,0 +1,25 @@
...
+#include <QToolButton>
+
+class QListWidget;
+class QButtonGroup;
+
+class MultiZimButton : public QToolButton {
+    Q_OBJECT
+
+public:
+    MultiZimButton(QWidget *parent = nullptr);
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/multizimbutton.h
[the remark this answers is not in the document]
ShaopengLin: Done

resources/i18n/qqq.json
```diff
@@ -180,5 +180,6 @@
 	"portable-disabled-tooltip": "Tooltip used to explain disabled components in the portable version.",
 	"scroll-next-tab": "Represents the action of scrolling to the next tab of the current tab which toward the end of the tab bar.",
 	"scroll-previous-tab": "Represents the action of scrolling to the previous tab of the current tab which toward the start of the tab bar.",
-	"kiwix-search": "Title text for a list of search results, which notes to the user those are from Kiwix's Search Engine"
+	"kiwix-search": "Title text for a list of search results, which notes to the user those are from Kiwix's Search Engine",
+	"multi-zim-search": "Represents the action of searching suggrestion results from one or more ZIM files in the library."
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/multizimbutton.cpp
veloman-yunkan: I am not saying that this code should be deleted and forgotten. Please move it in a separate draft PR.

src/multizimbutton.cpp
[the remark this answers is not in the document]
veloman-yunkan: Why did you remove the entire `try{}catch(){}` statement? My understanding was that you wanted to exclude those books that didn't have a valid ZIM archive associated with them (e.g. books that are still being downloaded, or books with ZIM archives on absent removable media).

src/multizimbutton.cpp
ShaopengLin: Done #1234

src/multizimbutton.cpp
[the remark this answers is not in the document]
ShaopengLin: Done, my bad forgot about the check.

ShaopengLin:
@veloman-yunkan
- Renamed button
- Resolved https://github.com/kiwix/kiwix-desktop/pull/1218/files#diff-f40632351954f93a1ef92d4db691acd31dbb8e10b6d76081983c601a04faa0ea and https://github.com/kiwix/kiwix-desktop/pull/1218/files#diff-f40632351954f93a1ef92d4db691acd31dbb8e10b6d76081983c601a04faa0ea

ShaopengLin:
I just realized the shortcut I gave search options is supposed to be used by #42. Let me change it to another one.

src/kiwixapp.cpp
```diff
@@ -438,6 +438,8 @@ void KiwixApp::createActions()
     CREATE_ACTION_SHORTCUT(ToggleTOCAction, gt("table-of-content"), QKeySequence(Qt::CTRL | Qt::SHIFT | Qt::Key_1));
     HIDE_ACTION(ToggleTOCAction);
 
+    CREATE_ACTION_ICON_SHORTCUT(OpenMultiZimAction, "filter", gt("search-options"), QKeySequence(Qt::CTRL | Qt::Key_M));
```
veloman-yunkan, line 441: @kelson42 Is the CTRL+M shortcut OK for a button named "Search options"?

veloman-yunkan (approved):
LGTM. Probably it would be better to get rid of the MultiZim (in source code) vs "Search options"  (as presented to the user) inconsistency but that's not a sound enough reason to delay the merging of this PR.

src/kiwixapp.cpp
ShaopengLin: I was aware of this as well. I will look around to find an appropriate one or what is the usual shortcut for this.

ShaopengLin: Done, switched to Ctrl + Shift + L which is a common shortcut for opening and closing search options.

PR #1220, opened 2024-09-29
Add support for new getBestPublicIp

sgourdas:
This makes the appropriate changes to support the libkiwix API change of `getBestPublicIp` implemented in kiwix/libkiwix#1132


sgourdas:
@kelson42 it is not a major change so we can easily merge it if you want.

sgourdas:
It is even reverse compatible, so we can merge before libkiwix.

Issue #1148, opened 2024-07-22
Check that the download can be saved to the target file-system


veloman-yunkan:
One of the issues behind #916 seems to be that a large download is saved to a FAT32 filesystem that has enough free space but cannot host files larger than 4GiB. Our pre-download checks must be enhanced.

charanbhatia:
hi, can you please assign this issue to me. I want to work on this.

kelson42:
@charanbhatia You should create the primitive in the libkiwix and do it in a way it is very quick to execute and without side effect. How would such a function look like and work to your opinion?

PR #1221, opened 2024-09-29
Add ipv6 support to local server

### LINKED ISSUE #1119: Add support of IPv6 IP to the HTTP daemon mode
Once https://github.com/kiwix/libkiwix/pull/1074 is merged

sgourdas:
This adds ipv6 support to the local kiwix server of the app.

Fixes #1119


kelson42 (changes requested):
I have compiled with libkiwix #1132 and even if I see now `::1` in the list, each time I start the server, then whole kiwix-desktop crashes. Actually this is the same with kiwix-desktop `main`.

kelson42 (approved):
Works as expected from a user perspective. Ready for code review.

sgourdas:
I would like to do some additional checks here before reviews, if possible.

sgourdas:
@veloman-yunkan this is ready for code review. 

P.S. It would be good for testing even to have in place the fix of `Server::setAddress` from kiwix/libkiwix#1147.

kelson42:
So Inguess this is the last PR of our IPv6 effort :)

sgourdas:
Now that ipv6 works correctly on Windows give me some additional time to catch any issues before we close this up please

sgourdas:
It would be good to merge kiwix/libkiwix#1151 first, as noted [here](https://github.com/kiwix/kiwix-desktop/pull/1221#issuecomment-2395041468), before finalizing this as well.

sgourdas:
It seems like the root cause helping the discovery of #1148 is the fact that the Kiwix Desktop local server uses `getNetworkInterfacesIPv4Or6` which does return `169.254` prefixed IPs. These are added to the dropdown of IPs "available" for use, which as we discussed leads to issues.

@veloman-yunkan should I filter these out when adding to the list, or should we modify the `getNetworkInterfacesIPv4Or6` behavior to not include them?

veloman-yunkan:
@sgourdas Neither seems a good option to me. Why don't we present the full list of IPs to the users and let them find out which one works best for them? We can add a note that not all of the IPs in the list may actually work. We can filter the list later after we have enough experience with problematic IPs and/or complaints from our users.

sgourdas:
> @sgourdas Neither seems a good option to me. Why don't we present the full list of IPs to the users and let them find out which one works best for them? We can add a note that not all of the IPs in the list may actually work. We can filter the list later after we have enough experience with problematic IPs and/or complaints from our users.

I can see that being best in a "text box scenario" where the input is coming from the user. IMHO if an option is given from a dropdown it should work correctly and have a reason to be presented/add clutter in our app.

On the other hand, filtering after we get more experience with problematic IPs is a good approach. But despite that, this prefix seems to get in the way and kind of obscure what works and what doesn't with how many IPs it provides.

veloman-yunkan:
@sgourdas If you can do some research, find out the peculiarity of `169.254.x.x` addresses and substantiate why they must be excluded from the IP addresses that we consider, then we can proceed based on that information.

sgourdas:
> @sgourdas If you can do some research, find out the peculiarity of `169.254.x.x` addresses and substantiate why they must be excluded from the IP addresses that we consider, then we can proceed based on that information.

@veloman-yunkan 159.254.0.0/16 addresses are link local addresses which should be excluded in `getNetworkInterfacesWin` like its already being done with the ipv6 equivalent, since their use is not destined for this purpose since they are non routable.

sgourdas:
> > @sgourdas If you can do some research, find out the peculiarity of `169.254.x.x` addresses and substantiate why they must be excluded from the IP addresses that we consider, then we can proceed based on that information.
> 
> @veloman-yunkan 169.254.0.0/16 addresses are link local addresses which IMO should be excluded in `getNetworkInterfacesWin` since its already being done with the ipv6 equivalent. Their use is not destined for this purpose since they are non routable.

The corresponding change should also be probably done in `getNetworkInterfacesPosix` in the same logic.

veloman-yunkan:
> Their use is not destined for this purpose since they are non routable.

@sgourdas Can you provide a reference to some document supporting your statement?

veloman-yunkan:
> > Their use is not destined for this purpose since they are non routable.
> 
> @sgourdas Can you provide a reference to some document supporting your statement?

Such addresses are not globally routable, but they should be usable within the local subnetwork, in which respect they are no different from, for example, 192.168.x.x addresses.

sgourdas:
@veloman-yunkan main sources of my response where [this](https://serverfault.com/questions/427018/what-is-this-ip-address-169-254-169-254) and [this](https://www.whatismyip.com/169-254-ip-address/#:~:text=Why%20can't%20169.254%20be,the%20link%2Dlocal%20address%20range.)

sgourdas:
@veloman-yunkan is this enough info to justify such a change?

veloman-yunkan:
@sgourdas I am not convinced. The main argument for suppressing 169.254.x.x IPs is that they seem to become of significance only when there is a misconfiguration or some other issue with the network, however that also can be thought of as an argument for not suppressing those IPs. I'll let @kelson42 decide.

sgourdas:
@veloman-yunkan do you happen to know what is the reason we are excluding IPv6 link locals but not doing the same for IPv4 currently?

veloman-yunkan:
> @veloman-yunkan do you happen to know what is the reason we are excluding IPv6 link locals but not doing the same for IPv4 currently?

No, I don't.

sgourdas:
> > @veloman-yunkan do you happen to know what is the reason we are excluding IPv6 link locals but not doing the same for IPv4 currently?
> 
> No, I don't.

Alright, lets continue with the main purpose of this PR maybe for the moment then.

src/localkiwixserver.cpp
```diff
@@ -43,6 +43,8 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
     for (const auto &interfacePair : interfacesMap) {
         QString ip = QString::fromStdString(interfacePair.second.addr);
         interfaces.push_back(ip);
```
veloman-yunkan: Won't an `if (!ip.isEmpty())` check be justified here?

src/localkiwixserver.cpp
```diff
@@ -43,6 +43,8 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
...
+        QString ipv6 = QString::fromStdString(interfacePair.second.addr6);
+        if (!ipv6.isEmpty()) interfaces.push_back(ipv6);
     }
     std::sort(interfaces.begin(), interfaces.end());
     interfaces.push_front(QString(gt("all")));
```
veloman-yunkan, line 50: Shouldn't we provide "All IPv4" and "All IPv6" options too?

src/localkiwixserver.cpp
sgourdas: Would not hurt to add for sure. I will go ahead and do it.

src/localkiwixserver.cpp
sgourdas: Sure, will add it besides All

sgourdas: Should I rename to:
- Dual Stack Mode
- IPv4 Mode
- IPv6 mode

kelson42:
> @sgourdas I am not convinced. The main argument for suppressing 169.254.x.x IPs is that they seem to become of significance only when there is a misconfiguration or some other issue with the network, however that also can be thought of as an argument for not suppressing those IPs. I'll let @kelson42 decide.

I'm pretty much in favour of not proposing this kind of IPs - if they are not reachable by non-local users - in Kiwix Desktop. This will do more arm than good IMHO. I believe the answer is slightly different for Kiwix Server, which is used by tech-savy users. So I would just propose to mask it here in Kiwix Desktop, but keep it available in libkiwix/kiwix-serve (but never choose it per default). Hope this helps.

resources/i18n/en.json
```diff
@@ -97,7 +97,9 @@
     "open-in-browser":"Open in browser",
     "start-kiwix-server":"Start Kiwix Server",
     "stop-kiwix-server":"Stop Kiwix Server",
-    "all":"All",
+    "all":"All - Dual Stack Mode",
```
veloman-yunkan, line 100: Such text contradicts the corresponding entry in `qqq.json` (which reads `{{Identical|All}}`). There will be a problem for `gt("all")` used in other context (for example, for filtering the library by category). If we want a more elaborate text for IPs we should introduce a new translation entry (e.g. `all_ips`).

resources/i18n/qqq.json
```diff
@@ -106,6 +106,8 @@
 	"start-kiwix-server": "Represents the action of starting the kiwix server.",
 	"stop-kiwix-server": "Represents the action of stopping the running kiwix server.",
 	"all": "{{Identical|All}}",
+	"ipv4": "Indicates IPv4 mode on local server.",
+    "ipv6": "Indicates IPv6 mode on local server",
```
veloman-yunkan, line 110: Fix the indentation (use tabs/spaces consistently)

veloman-yunkan (changes requested):

src/localkiwixserver.cpp
```diff
@@ -84,11 +88,21 @@ void LocalKiwixServer::runOrStopServer()
...
+        } else if (ui->IpChooser->currentText() == gt("ipv4")) {
+            mp_server->setIpMode(kiwix::IpMode::IPV4);
+            m_ipAddress = QString::fromStdString(kiwix::getBestPublicIps().addr);
+        } else if (ui->IpChooser->currentText() == gt("ipv6")) {
+            mp_server->setIpMode(kiwix::IpMode::IPV6);
+            m_ipAddress = QString::fromStdString(kiwix::getBestPublicIps().addr6);
+        } else {
+            m_ipAddress = ui->IpChooser->currentText();
+            mp_server->setAddress(m_ipAddress.toStdString());
+        }
```
veloman-yunkan, line 103: Isn't it better to set `m_ipAddress` from the result of `mp_server->getAddress()` after the server is started?

veloman-yunkan (changes requested):

src/localkiwixserver.cpp
```diff
@@ -84,11 +88,21 @@ void LocalKiwixServer::runOrStopServer()
...
+            m_ipAddress = QString::fromStdString(kiwix::getBestPublicIp());
+        } else if (ui->IpChooser->currentText() == gt("ipv4")) {
+            mp_server->setIpMode(kiwix::IpMode::IPV4);
+            m_ipAddress = QString::fromStdString(kiwix::getBestPublicIps().addr);
+        } else if (ui->IpChooser->currentText() == gt("ipv6")) {
+            mp_server->setIpMode(kiwix::IpMode::IPV6);
+            m_ipAddress = QString::fromStdString(kiwix::getBestPublicIps().addr6);
+        } else {
+            m_ipAddress = ui->IpChooser->currentText();
+            mp_server->setAddress(m_ipAddress.toStdString());
```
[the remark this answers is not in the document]
sgourdas: I tried to be smart since it auto initializes as empty, but we need it indeed for multiple invocations. I should also set mode to every case for the same reason.

src/localkiwixserver.cpp
sgourdas: That involves moving the relevant logic of `ui->IpAddress->setText` after the server start right? If we don't have an issue with that, then definitely.

src/localkiwixserver.cpp
```diff
@@ -84,18 +88,24 @@ void LocalKiwixServer::runOrStopServer()
...
-        mp_server->setAddress(m_ipAddress.toStdString());
-        m_ipAddress = (m_ipAddress != "0.0.0.0") ? m_ipAddress : QString::fromStdString(kiwix::getBestPublicIp());
-        ui->IpAddress->setText("http://" + m_ipAddress + ":" + QString::number(m_port));
-        ui->IpAddress->setReadOnly(true);
+        mp_server->setIpMode(kiwix::IpMode::AUTO);
+        mp_server->setAddress("");
+        (ui->IpChooser->currentText() == gt("all_ips")) ? mp_server->setIpMode(kiwix::IpMode::ALL)  :
+        (ui->IpChooser->currentText() == gt("ipv4"))    ? mp_server->setIpMode(kiwix::IpMode::IPV4) :
+        (ui->IpChooser->currentText() == gt("ipv6"))    ? mp_server->setIpMode(kiwix::IpMode::IPV6) :
+                                                          mp_server->setAddress(ui->IpChooser->currentText().toStdString());
```
veloman-yunkan, line 97: Conditional expressions shouldn't be used in place of statements, i.e. when you care about the side-effect rather than the value of the expression. Change to `if() {} else if () {} ...`

src/localkiwixserver.cpp
```diff
@@ -84,18 +88,24 @@ void LocalKiwixServer::runOrStopServer()
...
+        settingsManager->setKiwixServerIpAddress(ui->IpChooser->currentText());
```
veloman-yunkan, line 98: What will happen if you change the user language/locale? The safe approach is to not store translated text in settings.

veloman-yunkan (changes requested):

src/localkiwixserver.cpp
```diff
@@ -41,15 +41,20 @@ LocalKiwixServer::LocalKiwixServer(QWidget *parent) :
...
-    interfaces.push_front(QString(gt("all")));
+    interfaces.push_front(QString(gt("ipv6")));
+    interfaces.push_front(QString(gt("ipv4")));
+    interfaces.push_front(QString(gt("all_ips")));
     for (const auto &interface : interfaces) {
         ui->IpChooser->addItem(interface);
     }
-    ui->IpChooser->setCurrentText(KiwixApp::instance()->getSettingsManager()->getKiwixServerIpAddress());
+    QString ipAddress = KiwixApp::instance()->getSettingsManager()->getKiwixServerIpAddress();
+    ui->IpChooser->setCurrentText(ipAddress == "all_ips" || ipAddress == "ipv4" || ipAddress == "ipv6" ? gt(ipAddress) : ipAddress);
```
veloman-yunkan, line 57: Please add backward compatibility by converting `"0.0.0.0"` to `"ipv4"`.

sgourdas: Good catch, have to start being more aware of that

sgourdas: Shouldnt it be converted to "all" though?

veloman-yunkan: I thought about that too, but chose to be conservative.

@kelson42 What's your opinion on this? Should we upgrade the old setting of the local server IP of "0.0.0.0" to all IPs (dual stack IPv4 & IPv6 mode) or just all IPv4 IPs?

sgourdas: @kelson42 do you have any suggestion on how to handle this?

kelson42: Not an expert, but to me 0.0.0.0 is the same as IPv4 all... and yes woukd be better to convert,

sgourdas: @kelson42 "0.0.0.0" in this case was just the internal representation. Better phrased I think the question would be "If a user of the previous implementation had chosen `All`, should that now be transferred to `All - Dual Stack Mode` or `IPv4 Only Mode`?"

kelson42: It's not an internal representation, see https://en.wikipedia.org/wiki/0.0.0.0

sgourdas: Yes, that is true, but I probably did not properly phrase what I meant. I mean that in this case it is not a matter of what the setting was saved as (which is 0.0.0.0 -- because we were only working with IPv4). It is a matter of what the user was selecting which was `All`. What should now be the equivalent of the previous `All`, based on the new options?

I hope this makes more sense.

kelson42: I believe to have written already twice that it should be all IPv4. If someone wants something different it will have to change it manually.

sgourdas: Clear, ok. Will proceed based on that.

veloman-yunkan (approved):

PR #1224, opened 2024-10-05
Implement Endless Zim Search Suggestions

ShaopengLin:
Fixes (part of) #413 

Changes:
  - Added a custom model to enable endless suggestions
  - Added basic UI elements such as the search header and icon of the zim for convenience.
  - Endless can be triggered by up/down/page-down keypresses and mouse scrolls.

Known Issues to be fixed in #1189: The first suggestion is hidden by the header due to QCompleter's incorrect size management.

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -172,7 +172,8 @@ void SearchBarLineEdit::focusInEvent( QFocusEvent* event)
         event->reason() == Qt::MouseFocusReason ||
         event->reason() == Qt::ShortcutFocusReason) {
         connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
-        this, QOverload<const QModelIndex &>::of(&SearchBarLineEdit::openCompletion));
+                this, QOverload<const QModelIndex &>::of(&SearchBarLineEdit::openCompletion),
+                Qt::UniqueConnection);
     }
     QLineEdit::focusInEvent(event);
 }
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -80,7 +80,7 @@ SearchBarLineEdit::SearchBarLineEdit(QWidget *parent) :
     m_suggestionView->setRootIsDecorated(false);
     m_suggestionView->setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/popup.css"));
 
-    connect(m_completer.popup()->verticalScrollBar(), &QScrollBar::valueChanged,
+    connect(m_suggestionView->verticalScrollBar(), &QScrollBar::valueChanged,
             this, &SearchBarLineEdit::onScroll);
 
     qRegisterMetaType<QList<SuggestionData>>("QList<SuggestionData>");
@@ -119,7 +119,7 @@ SearchBarLineEdit::SearchBarLineEdit(QWidget *parent) :
 
 void SearchBarLineEdit::hideSuggestions()
 {
-    m_completer.popup()->hide();
+    m_suggestionView->hide();
 }
 
 bool SearchBarLineEdit::eventFilter(QObject *, QEvent *event)
@@ -222,15 +222,15 @@ void SearchBarLineEdit::onScroll(int value)
        to intercept the scrolling in eventFilter so, until we find out how, this
        code is here to stay.
     */
-    if (!m_completer.popup()->currentIndex().isValid())
+    if (!m_suggestionView->currentIndex().isValid())
     {
-        const auto old = m_completer.popup()->verticalScrollBar()->blockSignals(true);
-        m_completer.popup()->scrollToBottom();
-        m_completer.popup()->verticalScrollBar()->blockSignals(old);
+        const auto old = m_suggestionView->verticalScrollBar()->blockSignals(true);
+        m_suggestionView->scrollToBottom();
+        m_suggestionView->verticalScrollBar()->blockSignals(old);
         return;
     }
 
-    const auto suggestionScroller = m_completer.popup()->verticalScrollBar();
+    const auto suggestionScroller = m_suggestionView->verticalScrollBar();
     const auto scrollMin = suggestionScroller->minimum();
     const auto scrollMax = suggestionScroller->maximum();
     const bool scrolledToEnd = value == suggestionScroller->maximum();
@@ -277,8 +277,8 @@ void SearchBarLineEdit::onInitialSuggestions(int)
         m_completer.complete();
 
         /* Make row 0 appear but do not highlight it */
-        const auto completerFirstIdx = m_completer.popup()->model()->index(0, 0);
-        const auto completerSelModel = m_completer.popup()->selectionModel();
+        const auto completerFirstIdx = m_suggestionView->model()->index(0, 0);
+        const auto completerSelModel = m_suggestionView->selectionModel();
         completerSelModel->setCurrentIndex(completerFirstIdx, QItemSelectionModel::Current);
     }
 }
@@ -286,9 +286,9 @@ void SearchBarLineEdit::onInitialSuggestions(int)
 void SearchBarLineEdit::onAdditionalSuggestions(int start)
 {
     /* Set selection to be at the last row of the previous list */
-    const auto completerStartIdx = m_completer.popup()->model()->index(start, 0);
-    m_completer.popup()->setCurrentIndex(completerStartIdx);
-    m_completer.popup()->show();
+    const auto completerStartIdx = m_suggestionView->model()->index(start, 0);
+    m_suggestionView->setCurrentIndex(completerStartIdx);
+    m_suggestionView->show();
 }
 
 void SearchBarLineEdit::fetchSuggestions(NewSuggestionHandlerFuncPtr callback)
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -162,7 +162,7 @@ void SearchBarLineEdit::updateCompletion()
         return;
     }
     m_token++;
-    auto suggestionWorker = new SuggestionListWorker(m_searchbarInput, m_token, this);
+    auto suggestionWorker = new SuggestionListWorker(m_searchbarInput, m_token, 0, this);
     connect(suggestionWorker, &SuggestionListWorker::searchFinished, this,
     [=] (const QList<SuggestionData>& suggestionList, int token) {
         if (token != m_token) {
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -49,7 +49,7 @@ void BookmarkButton::on_buttonClicked()
 
 SearchBarLineEdit::SearchBarLineEdit(QWidget *parent) :
     QLineEdit(parent),
-    m_completer(&m_completionModel, this)
+    m_completer(&m_suggestionModel, this)
 {
     setAlignment(KiwixApp::isRightToLeft() ? Qt::AlignRight : Qt::AlignLeft);
     mp_typingTimer = new QTimer(this);
@@ -111,8 +111,7 @@ void SearchBarLineEdit::hideSuggestions()
 
 void SearchBarLineEdit::clearSuggestions()
 {
-    QStringList empty;
-    m_completionModel.setStringList(empty);
+    m_suggestionModel.resetSuggestions();
     m_urlList.clear();
 }
 
@@ -175,7 +174,7 @@ void SearchBarLineEdit::updateCompletion()
             openCompletion(suggestions.first(), 0);
             return;
         }
-        m_completionModel.setStringList(suggestions);
+        m_suggestionModel.append(suggestions);
         m_completer.complete();
     });
     connect(suggestionWorker, &SuggestionListWorker::finished, suggestionWorker, &QObject::deleteLater);
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -2,6 +2,7 @@
 
 #include <QCompleter>
 #include <QFocusEvent>
+#include <QScrollBar>
 
 #include "kiwixapp.h"
 #include "suggestionlistworker.h"
@@ -65,10 +66,14 @@ SearchBarLineEdit::SearchBarLineEdit(QWidget *parent) :
     setToolTip(gt("search"));
     m_completer.setCompletionMode(QCompleter::UnfilteredPopupCompletion);
     m_completer.setCaseSensitivity(Qt::CaseInsensitive);
-    m_completer.setMaxVisibleItems(16);
+
+    /* The items should be less than fetch size to enable scrolling. */
+    m_completer.setMaxVisibleItems(SuggestionListWorker::getFetchSize() / 2);
     setCompleter(&m_completer);
 
     m_completer.popup()->setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/popup.css"));
+    connect(m_completer.popup()->verticalScrollBar(), &QScrollBar::valueChanged,
+            this, &SearchBarLineEdit::onScroll);
 
     qRegisterMetaType<QList<SuggestionData>>("QList<SuggestionData>");
     connect(mp_typingTimer, &QTimer::timeout, this, &SearchBarLineEdit::updateCompletion);
@@ -165,6 +170,47 @@ void SearchBarLineEdit::updateCompletion()
     fetchSuggestions(&SearchBarLineEdit::onInitialSuggestions);
 }
 
+void SearchBarLineEdit::fetchMoreSuggestions()
+{
+    /* TODO: Refactor suggestion worker to re-use zim::SuggestionSearcher for
+       fetching more suggestion in a single archive. Currently we create a 
+       searcher for every fetch, and discarded after one use.
+    */
+    fetchSuggestions(&SearchBarLineEdit::onAdditionalSuggestions);
+}
+
+void SearchBarLineEdit::onScroll(int value)
+{
+    const auto suggestionScroller = m_completer.popup()->verticalScrollBar();
+    const auto scrollMin = suggestionScroller->minimum();
+    const auto scrollMax = suggestionScroller->maximum();
+    const bool scrolledToEnd = value == suggestionScroller->maximum();
+    if (m_aboutToScrollPastEnd)
+    {
+        if (scrolledToEnd)
+        {
+            /* The user's intention to scroll past end has been confirmed */
+            fetchMoreSuggestions();
+            m_aboutToScrollPastEnd = false; /* Relax until next time */
+        }
+        else
+        {
+            /* Scrolling past end did not happen - remove the extra scroll
+               room created for detecting the intention of scrolling past end
+            */
+            suggestionScroller->setRange(scrollMin, scrollMax - 1);
+            m_aboutToScrollPastEnd = false; /* ... and relax */
+        }
+    }
+    else if (scrolledToEnd)
+    {
+        /* The user has scrolled to end - monitor for furthur scrolling */
+        m_aboutToScrollPastEnd = true;
+        /* Create some fictitious room for an extra scroll */
+        suggestionScroller->setRange(scrollMin, scrollMax + 1);
+    }
+}
+
 void SearchBarLineEdit::openCompletion(const QModelIndex &index)
 {
     if (index.isValid())
@@ -181,6 +227,14 @@ void SearchBarLineEdit::onInitialSuggestions(int)
     } else {
         m_completer.complete();
     }
+}
+
+void SearchBarLineEdit::onAdditionalSuggestions(int start)
+{
+    /* Set selection to be at the last row of the previous list */
+    const auto completerStartIdx = m_completer.popup()->model()->index(start, 0);
+    m_completer.popup()->setCurrentIndex(completerStartIdx);
+    m_completer.popup()->show();
 }
 
 void SearchBarLineEdit::fetchSuggestions(NewSuggestionHandlerFuncPtr callback)
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -71,7 +71,7 @@ SearchBarLineEdit::SearchBarLineEdit(QWidget *parent) :
 
     /* The items should be less than fetch size to enable scrolling. */
     m_completer.setMaxVisibleItems(SuggestionListWorker::getFetchSize() / 2);
-    setCompleter(&m_completer);
+    m_completer.setWidget(this);
 
     /* QCompleter's uses default list views, which do not have headers. */
     m_completer.setPopup(m_suggestionView);
@@ -171,8 +171,17 @@ void SearchBarLineEdit::focusInEvent( QFocusEvent* event)
     if (event->reason() == Qt::ActiveWindowFocusReason ||
         event->reason() == Qt::MouseFocusReason ||
         event->reason() == Qt::ShortcutFocusReason) {
+        connect(&m_completer, QOverload<const QString &>::of(&QCompleter::activated),
+                this, &QLineEdit::setText,Qt::UniqueConnection);
+
         connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
                 this, QOverload<const QModelIndex &>::of(&SearchBarLineEdit::openCompletion),
+                Qt::UniqueConnection);
+
+        connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::highlighted), this,
+                [=](const QModelIndex &index){
+                    setText(index.isValid() ? index.data().toString() : m_searchbarInput);
+                },
                 Qt::UniqueConnection);
     }
     QLineEdit::focusInEvent(event);
@@ -185,6 +194,7 @@ void SearchBarLineEdit::focusOutEvent(QFocusEvent* event)
         setText(m_title);
     }
     deselect();
+    disconnect(&m_completer, nullptr, this, nullptr);
     return QLineEdit::focusOutEvent(event);
 }
 
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -117,7 +117,7 @@ void SearchBarLineEdit::hideSuggestions()
 
 bool SearchBarLineEdit::eventFilter(QObject *, QEvent *event)
 {
-    if (!m_aboutToScrollPastEnd)
+    if (!(m_aboutToScrollPastEnd && m_moreSuggestionsAreAvailable))
         return false;
 
     if (const auto e = dynamic_cast<QKeyEvent *>(event))
@@ -139,6 +139,7 @@ bool SearchBarLineEdit::eventFilter(QObject *, QEvent *event)
 void SearchBarLineEdit::clearSuggestions()
 {
     m_suggestionModel.resetSuggestions();
+    m_moreSuggestionsAreAvailable = false;
 }
 
 void SearchBarLineEdit::on_currentTitleChanged(const QString& title)
@@ -203,6 +204,12 @@ void SearchBarLineEdit::fetchMoreSuggestions()
 
 void SearchBarLineEdit::onScroll(int value)
 {
+    if (!m_moreSuggestionsAreAvailable)
+    {
+        m_aboutToScrollPastEnd = false;
+        return;
+    }
+
     /* Scrolling using key_down past end will teleport scroller to the top.
        We undo this here. Block signal to avoid recursion. We cannot find a way
        to intercept the scrolling in eventFilter so, until we find out how, this
@@ -283,6 +290,10 @@ void SearchBarLineEdit::fetchSuggestions(NewSuggestionHandlerFuncPtr callback)
                 }
 
                 m_suggestionModel.append(suggestionList);
+                const int listSize = suggestionList.size();
+                const bool hasFullText = listSize > 0 && suggestionList.back().isFullTextSearchSuggestion();
+                const int maxFetchSize = SuggestionListWorker::getFetchSize() + hasFullText;
+                m_moreSuggestionsAreAvailable = listSize >= maxFetchSize;
                 (this->*callback)(start);
             });
     connect(suggestionWorker, &SuggestionListWorker::finished, suggestionWorker, &QObject::deleteLater);
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -274,7 +274,7 @@ void SearchBarLineEdit::onAdditionalSuggestions(int start)
 
 void SearchBarLineEdit::fetchSuggestions(NewSuggestionHandlerFuncPtr callback)
 {
-    const int start = m_suggestionModel.rowCount();
+    const int start = m_suggestionModel.countOfRegularSuggestions();
     const auto suggestionWorker = new SuggestionListWorker(m_searchbarInput, m_token, start, this);
     connect(suggestionWorker, &SuggestionListWorker::searchFinished, this,
             [=] (const QList<SuggestionData>& suggestionList, int token) {
@@ -296,11 +296,13 @@ QModelIndex SearchBarLineEdit::getDefaulSuggestionIndex() const
         return firstSuggestionIndex;
 
     /* If the first entry matches the typed text, use it as default, otherwise
-       use the last entry which is the fulltext search. */
+       use the last entry if fulltext search exist. */
     const auto firstSuggestionText = firstSuggestionIndex.data().toString();
     if (this->text().compare(firstSuggestionText, Qt::CaseInsensitive) == 0)
         return firstSuggestionIndex;
-    return m_suggestionModel.index(m_suggestionModel.rowCount() - 1);
+    else if (m_suggestionModel.hasFullTextSearchSuggestion())
+        return m_suggestionModel.index(m_suggestionModel.rowCount() - 1);
+    return QModelIndex();
 }
 
 SearchBar::SearchBar(QWidget *parent) :
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -171,7 +171,7 @@ void SearchBarLineEdit::updateCompletion()
 
         m_suggestionModel.append(suggestionList);
         if (m_returnPressed) {
-            openCompletion(suggestionList.first().text, 0);
+            openCompletion(m_suggestionModel.index(0));
             return;
         }
         m_completer.complete();
@@ -182,20 +182,18 @@ void SearchBarLineEdit::updateCompletion()
 
 void SearchBarLineEdit::openCompletion(const QModelIndex &index)
 {
-    if (m_suggestionModel.rowCount() != 0) {
-        openCompletion(index.data().toString(), index.row());
-    }
-}
+    if (index.isValid())
+    {
+        QUrl url;
+        auto selectedSuggestionText = index.data().toString();
 
-void SearchBarLineEdit::openCompletion(const QString& text, int index)
-{
-    QUrl url;
-    if (this->text().compare(text, Qt::CaseInsensitive) == 0) {
-        url = m_suggestionModel.index(index).data(Qt::UserRole).toUrl();
-    } else {
-        url = m_suggestionModel.index(m_suggestionModel.rowCount() - 1).data(Qt::UserRole).toUrl();
+        if (this->text().compare(selectedSuggestionText, Qt::CaseInsensitive) == 0) {
+            url = index.data(Qt::UserRole).toUrl();
+        } else {
+            url = m_suggestionModel.index(m_suggestionModel.rowCount() - 1).data(Qt::UserRole).toUrl();
+        }
+        QTimer::singleShot(0, [=](){KiwixApp::instance()->openUrl(url, false);});
     }
-    QTimer::singleShot(0, [=](){KiwixApp::instance()->openUrl(url, false);});
 }
 
 SearchBar::SearchBar(QWidget *parent) :
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -70,7 +70,7 @@ SearchBarLineEdit::SearchBarLineEdit(QWidget *parent) :
 
     m_completer.popup()->setStyleSheet(KiwixApp::instance()->parseStyleFromFile(":/css/popup.css"));
 
-    qRegisterMetaType<QVector<QUrl>>("QVector<QUrl>");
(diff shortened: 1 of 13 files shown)


src/searchbar.cpp
```diff
@@ -183,20 +182,17 @@ void SearchBarLineEdit::updateCompletion()
...
+        auto editText = index.data().toString();
+        if (this->text().compare(editText, Qt::CaseInsensitive) == 0) {
+            url = index.data(Qt::UserRole).toUrl();
+        } else {
+            url = m_suggestionModel.lastIndex().data(Qt::UserRole).toUrl();
+        }
+        QTimer::singleShot(0, [=](){KiwixApp::instance()->openUrl(url, false);});
     }
-    QTimer::singleShot(0, [=](){KiwixApp::instance()->openUrl(url, false);});
 }
```
veloman-yunkan: I think it's better to remove the parameter from `SearchBarLineEdit::openCompletion()` and rename it to `openDefaultCompletion()`

src/searchbar.cpp
```diff
@@ -183,20 +182,17 @@ void SearchBarLineEdit::updateCompletion()
...
-        url = m_urlList.last();
+    if (m_suggestionModel.rowCount() != 0) 
+    {
+        QUrl url;
+        auto editText = index.data().toString();
+        if (this->text().compare(editText, Qt::CaseInsensitive) == 0) {
+            url = index.data(Qt::UserRole).toUrl();
+        } else {
+            url = m_suggestionModel.lastIndex().data(Qt::UserRole).toUrl();
+        }
```
veloman-yunkan, line 291: I guess the logic here is select the first suggestion if it fully matches the search text otherwise perform full text search which should be at the end. However if the ZIM file doesn't contain a full text search database then the full text search suggestion is not generated.

src/searchbar.cpp
```diff
@@ -65,10 +66,14 @@ SearchBarLineEdit::SearchBarLineEdit(QWidget *parent) :
     setToolTip(gt("search"));
     m_completer.setCompletionMode(QCompleter::UnfilteredPopupCompletion);
     m_completer.setCaseSensitivity(Qt::CaseInsensitive);
-    m_completer.setMaxVisibleItems(16);
+
+    /* The items should be less than fetch size to enable scrolling. */
+    m_completer.setMaxVisibleItems(8);
```
veloman-yunkan, line 73: 8 is a magic number. Why not `SuggestionListWorker::getFetchSize() - 1` (which also provides for a longer list of visible suggestions)?

src/searchbar.cpp
```diff
@@ -184,11 +184,12 @@ void SearchBarLineEdit::updateCompletion()
     m_token++;
     auto suggestionWorker = new SuggestionListWorker(m_searchbarInput, m_token, 0, this);
     connect(suggestionWorker, &SuggestionListWorker::searchFinished, this,
-    [=] (const QStringList& suggestions, const QVector<QUrl>& urlList, int token) {
+    [=] (const QStringList& suggestions, const QVector<QUrl>& urlList, bool hasFullText, int token) {
```
veloman-yunkan, line 194: A more straightforward solution (assuming the rest of my design change proposals are implemented) is to introduce a member function `bool SuggestionData::isFullTextSearchSuggestion() const` that works off the contents of the host part of the url (checks if it ends with `.search`).

veloman-yunkan (changes requested):

veloman-yunkan:
Also there is the following bug:

1. Open the [Ray Charles ZIM file](https://github.com/kiwix/libkiwix/blob/13.1.0/test/data/zimfile.zim) 
2. Start typing **beatles** in the search box

As soon as you have typed the first three letters (**bea**) the text in the search box automatically changes to **bea (Fulltext search)**.

src/searchbar.cpp
ShaopengLin: I will use `SuggestionListWorker::getFetchSize() / 2` here. Reason being endless suggestions need a scroller to work (otherwise no scrolling could happen...), which the view wouldn't have if it can contain all the entries without scrolling.

src/searchbar.cpp
ShaopengLin: I don't believe we can remove the parameter, this function opens the completion based on the user-selected index in `focusInEvent`.

src/suggestionlistmodel.h
```diff
@@ -15,10 +16,12 @@ class SuggestionListModel : public QAbstractListModel
...
 
     void resetSuggestions();
-    void append(const QStringList& suggestions);
+    void append(const QStringList& suggestions, const QVector<QUrl>& urlList);
+
+    QModelIndex lastIndex() const;
 
 private:
-    QList<QString> m_suggestions;
+    QList<QPair<QString, QUrl>> m_suggestions;
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done, added clause to simply return if we do not have full text in this situation.

src/searchbar.cpp
ShaopengLin: Done

ShaopengLin:
@veloman-yunkan The bug is caused by the absence of the commits that prevent suggestion completer flickering. I added those and the bug should be solved.

src/searchbar.cpp
veloman-yunkan: > I will use `SuggestionListWorker::getFetchSize() / 2` here.

Why won't `SuggestionListWorker::getFetchSize() - 1` work?

src/searchbar.cpp
veloman-yunkan: I see. I only payed attention to its usage in `SearchBarLineEdit::updateCompletion()` where it is called directly with an argument of `m_suggestionModel.index(0)` and missed its usage in `SearchBarLineEdit::focusInEvent()`. But then the code logic is a little confusing for the scenario when a suggestion is selected from the list - it must be clarified that upon selection of an item from the completer popup the content of the search box is updated with the text of the selected item.

Besides the name of the variable `editText` seems to be misleading - that semantics better fits the expression `this->text()` rather than `index.data().toString()` which is better described by `selectedSuggestionText` or `selectedItemText`.

src/suggestionlistmodel.cpp
```diff
@@ -55,6 +57,7 @@ void SuggestionListModel::append(const QList<SuggestionData>& suggestionList)
 
     for (const auto& suggestion : suggestionList)
         m_suggestions.append(suggestion);
+    setNoMoreSuggestion(suggestionList.size());
```
veloman-yunkan, line 75: This is a bad way for establishing exhaustion of suggestions as it means that `SuggestionListModel::append()` must be called only on the output of `SuggestionListWorker`. A better solution (though still a hack) may be passing the information on availability of additional suggestions via a special entry (e.g. with empty url) in `suggestionList`. But a simpler solution is to maintain the information on availability of more suggestions in `SearchLineEdit` since it is the class that arranges fetching of suggestions.

veloman-yunkan (changes requested):

src/searchbar.h
```diff
@@ -49,7 +48,6 @@ public slots:
 private slots:
     void updateCompletion();
     void openCompletion(const QModelIndex& index);
-    void openCompletion(const QString& text, int index);
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -180,6 +185,47 @@ void SearchBarLineEdit::updateCompletion()
...
+
+        m_suggestionModel.append(suggestionList);
+
+        /* Set selection to be at the last row of the previous list */
+        m_completer.popup()->setCurrentIndex(m_suggestionModel.index(start));
+        m_completer.popup()->show();
+    });
+    connect(suggestionWorker, &SuggestionListWorker::finished, suggestionWorker, &QObject::deleteLater);
+    suggestionWorker->start();
+}
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/suggestionlistmodel.cpp
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.h
```diff
@@ -44,9 +44,12 @@ public slots:
     bool m_returnPressed = false;
     QTimer* mp_typingTimer;
     int m_token;
+    bool m_scrolledEndBefore = false;
 
 private slots:
     void updateCompletion();
+    void fetchMoreSuggestion();
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -180,6 +185,47 @@ void SearchBarLineEdit::updateCompletion()
...
+
+void SearchBarLineEdit::onScrollToEnd(int value)
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -202,7 +203,7 @@ void SearchBarLineEdit::updateCompletion()
 
 void SearchBarLineEdit::fetchMoreSuggestion()
 {
-    int start = m_suggestionModel.lastIndex().row();
+    int start = m_suggestionModel.fetchEndIndex().row() + 1;
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -190,8 +191,20 @@ void SearchBarLineEdit::fetchMoreSuggestions()
     fetchSuggestions(&SearchBarLineEdit::onAdditionalSuggestions);
 }
 
+void SearchBarLineEdit::setNoMoreSuggestion(int fetchedSize)
+{
+    int trueFetchSize = fetchedSize - (m_suggestionModel.hasFullTextSearchSuggestion() ? 1 : 0);
+    m_noMoreSuggestion = trueFetchSize < SuggestionListWorker::getFetchSize();
+}
```
veloman-yunkan, line 215: Such a method is not fool proof. Its implementation assumes that it is called after the fetched suggestions have been appended to `m_suggestionModel`. Always try to avoid such tacit dependencies between members of your API. I see no justification for this member function. Just inline it in its sole point of usage and compute the status without any reference to `m_suggestionModel` (use only the incoming list of suggestions).

src/searchbar.cpp
```diff
@@ -119,7 +119,7 @@ bool SearchBarLineEdit::eventFilter(QObject *, QEvent *event)
 {
     if (auto e = static_cast<QKeyEvent *>(event))
     {
-        if (e->key() == Qt::Key_PageDown && m_scrolledEndBefore)
+        if (e->key() == Qt::Key_PageDown && m_scrolledEndBefore && !m_noMoreSuggestion)
```
veloman-yunkan, line 129: Double negation can be confusing. Please replace the `m_noMoreSuggestion` data member with the inverse variable `m_moreSuggestionsAreAvailable`

src/searchbar.h
```diff
@@ -46,11 +46,15 @@ public slots:
     bool m_returnPressed = false;
     QTimer* mp_typingTimer;
     int m_token;
+    bool m_scrolledEndBefore = false;
```
veloman-yunkan, line 53: 1. Explain the meaning of this data member here
2. Consider renaming it to something better (e.g. `m_haveAlreadyScrolledToEnd`, `m_onTheVergeOfScrollingPastEnd`, `m_aboutToScrollPastEnd`, `m_justShortOfScrollingPastEnd`, etc)

src/searchbar.cpp
```diff
@@ -266,6 +266,11 @@ void SearchBarLineEdit::onInitialSuggestions(int)
         openCompletion(m_suggestionModel.index(0));
     } else {
         m_completer.complete();
+
+        /* Make row 0 appear but does not highlight it */
```
veloman-yunkan, line 287: does -> do

src/searchbar.cpp
```diff
@@ -164,9 +164,17 @@ void SearchBarLineEdit::focusInEvent( QFocusEvent* event)
...
+        this, &QLineEdit::setText,Qt::UniqueConnection);
+
         connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::activated),
         this, QOverload<const QModelIndex &>::of(&SearchBarLineEdit::openCompletion),
         Qt::UniqueConnection);
+
+        connect(&m_completer, QOverload<const QModelIndex &>::of(&QCompleter::highlighted),
+        this, [=](const QModelIndex &index){
+            setText(index.isValid() ? index.data().toString() : m_searchbarInput);
+        }, Qt::UniqueConnection);
```
veloman-yunkan, line 177: Indent the 2nd and subsequent lines of every `connect()` expression so that  they are aligned with the first argument on the first line (fix this formatting in the earliest commit where the respective expression is modified).

veloman-yunkan (changes requested):
You can guess from the changed nature of most comments that we are now very close to merging this PR. The next iteration can be the final one.

src/suggestionlistworker.h
```diff
@@ -13,7 +15,7 @@ class SuggestionListWorker : public QThread
     void run() override;
 
 signals:
-    void searchFinished(const QStringList& suggestions, const QVector<QUrl>& urlList, int token);
+    void searchFinished(const QList<SuggestionData>& suggestionList, const QVector<QUrl>& urlList, int token);
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -182,20 +182,21 @@ void SearchBarLineEdit::updateCompletion()
...
+        auto selectedSuggestionText = index.data().toString();
+
+        /* On selection, text in lineedit is edited to selectedSuggestionText
+           by the compeleter.
+        */
+        if (this->text().compare(selectedSuggestionText, Qt::CaseInsensitive) == 0) {
+            url = index.data(Qt::UserRole).toUrl();
+        } else {
+            url = m_suggestionModel.lastIndex().data(Qt::UserRole).toUrl();
+        }
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -199,6 +184,32 @@ void SearchBarLineEdit::openCompletion(const QModelIndex &index)
...
+    int start = m_suggestionModel.rowCount();
+    auto suggestionWorker = new SuggestionListWorker(m_searchbarInput, m_token, start, this);
+    connect(suggestionWorker, &SuggestionListWorker::searchFinished, this,
+    [=] (const QList<SuggestionData>& suggestionList, int token) {
+        if (token != m_token) {
+            return;
+        }
+
+        m_suggestionModel.append(suggestionList);
+        (this->*callback)(start);
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.h
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -177,6 +177,21 @@ void SearchBarLineEdit::fetchMoreSuggestions()
...
+
+        /* m_completer.popup()->currentIndex() being invalid means the list
+           has been scrolled from bottom to top. We undo this here, as it avoids
+           scroll bar flicker as well. Block signal to avoid recursion.
+        */
+        auto old = m_completer.popup()->verticalScrollBar()->blockSignals(true);
+        m_completer.popup()->scrollToBottom();
+        m_completer.popup()->verticalScrollBar()->blockSignals(old);
+        
+        return fetchMoreSuggestions();
```
veloman-yunkan: What about packaging this piece of code (without the comment which should stay here) into a new helper method `SearchBarLineEdit::onScrollPastEnd()` and reusing it for the other call of `fetchMoreSuggestions()`?

src/searchbar.cpp
```diff
@@ -114,6 +115,20 @@ void SearchBarLineEdit::hideSuggestions()
...
 }
 
+bool SearchBarLineEdit::eventFilter(QObject *, QEvent *event)
+{
+    if (auto e = static_cast<QKeyEvent *>(event))
+    {
+        if (e->key() == Qt::Key_PageDown && m_aboutToScrollPastEnd)
+        {
+            m_aboutToScrollPastEnd = false;
+            fetchMoreSuggestions();
```
veloman-yunkan, line 132: Can reuse `onScrollPastEnd()` here too.

src/searchbar.cpp
```diff
@@ -263,6 +270,9 @@ void SearchBarLineEdit::fetchSuggestions(NewSuggestionHandlerFuncPtr callback)
                 }
 
                 m_suggestionModel.append(suggestionList);
+                bool hasFullText = suggestionList.size() > 0 && suggestionList.back().isFullTextSearchSuggestion();
+                int trueFetchSize = suggestionList.size() - (hasFullText ? 1 : 0);
+                m_moreSuggestionsAreAvailable = trueFetchSize >= SuggestionListWorker::getFetchSize();
```
veloman-yunkan, line 298: Risk of underflow - when `suggestionList.size() == 0` the initializer expression of `trueFetchSize` is questionable if the type of `suggestionList.size()` is unsigned. The following should be safer:

```c++
m_moreSuggestionsAreAvailable = suggestionList.size() >= SuggestionListWorker::getFetchSize() + hasFullText;
```

veloman-yunkan (changes requested):

src/searchbar.cpp
```diff
@@ -184,18 +184,16 @@ void SearchBarLineEdit::openCompletion(const QModelIndex &index)
...
+        QUrl url = index.data(Qt::UserRole).toUrl();
         QTimer::singleShot(0, [=](){KiwixApp::instance()->openUrl(url, false);});
     }
 }
 
+QModelIndex SearchBarLineEdit::getDefaulSuggestionIndex() const
+{
+    return m_suggestionModel.index(m_suggestionModel.rowCount() - 1);
+}
+
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -189,6 +174,32 @@ void SearchBarLineEdit::openCompletion(const QModelIndex &index)
     }
 }
 
+void SearchBarLineEdit::onInitialSuggestions(int)
+{
+    if (m_returnPressed) {
+        openCompletion(m_suggestionModel.index(0));
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -165,6 +170,32 @@ void SearchBarLineEdit::updateCompletion()
...
+    auto scrollMin = suggestionScroller->minimum();
+    auto scrollMax = suggestionScroller->maximum();
+    if (m_aboutToScrollPastEnd && scrolledToEnd)
+        fetchMoreSuggestions();
+    else if (scrolledToEnd)
+        suggestionScroller->setRange(scrollMin, scrollMax + 1);
+    else if (m_aboutToScrollPastEnd)
+        suggestionScroller->setRange(scrollMin, scrollMax - 1);
+
+    m_aboutToScrollPastEnd = !m_aboutToScrollPastEnd && scrolledToEnd;
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

veloman-yunkan: After going to bed I figured out that here there was no risk of underflow because `hasFullText` is `false` when `suggestionList.size() == 0` and wanted to delete this comment in the morning. But the general advice still holds - when operating in the vicinity of `0`, addition is safer than subtraction. Similarly, when operating in the vicinity of the upper limit of an integer numeric type, subtraction is safer than addition.

veloman-yunkan (changes requested):
There are still a couple of bugs (one old and one new) so let's brush up the PR a bit more.

src/searchbar.cpp
```diff
@@ -184,18 +184,25 @@ void SearchBarLineEdit::openCompletion(const QModelIndex &index)
 {
     if (m_suggestionModel.rowCount() != 0)
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -189,6 +174,32 @@ void SearchBarLineEdit::openCompletion(const QModelIndex &index)
...
+    int start = m_suggestionModel.rowCount();
+    auto suggestionWorker = new SuggestionListWorker(m_searchbarInput, m_token, start, this);
+    connect(suggestionWorker, &SuggestionListWorker::searchFinished, this,
+            [=] (const QList<SuggestionData>& suggestionList, int token) {
+                if (token != m_token) {
+                    return;
+                }
+
+                m_suggestionModel.append(suggestionList);
+                (this->*callback)(start);
```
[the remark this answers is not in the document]
ShaopengLin: Unchanged

src/searchbar.cpp
```diff
@@ -165,6 +170,44 @@ void SearchBarLineEdit::updateCompletion()
...
+{
+    fetchSuggestions(&SearchBarLineEdit::onAdditionalSuggestions);
+}
+
+void SearchBarLineEdit::onScroll(int value)
+{
+    auto suggestionScroller = m_completer.popup()->verticalScrollBar();
+    auto scrollMin = suggestionScroller->minimum();
+    auto scrollMax = suggestionScroller->maximum();
+    bool scrolledToEnd = value == suggestionScroller->maximum();
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -208,6 +208,21 @@ void SearchBarLineEdit::onScroll(int value)
...
+
+    /* m_completer.popup()->currentIndex() being invalid means the list
+       has been scrolled from bottom to top. We undo this here, as it avoids
+       scroll bar flicker as well. Block signal to avoid recursion.
+    */
+    auto old = m_completer.popup()->verticalScrollBar()->blockSignals(true);
+    m_completer.popup()->scrollToBottom();
+    m_completer.popup()->verticalScrollBar()->blockSignals(old);
+    
+    return fetchMoreSuggestions();
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -165,6 +170,44 @@ void SearchBarLineEdit::updateCompletion()
...
+void SearchBarLineEdit::onScroll(int value)
+{
+    auto suggestionScroller = m_completer.popup()->verticalScrollBar();
+    auto scrollMin = suggestionScroller->minimum();
+    auto scrollMax = suggestionScroller->maximum();
+    bool scrolledToEnd = value == suggestionScroller->maximum();
+    
+    if (m_aboutToScrollPastEnd)
+    {
+        /* The user's intention to scroll past end has been confirmed */
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -208,6 +208,21 @@ void SearchBarLineEdit::onScroll(int value)
...
 }
 
+void SearchBarLineEdit::onScrollPastEnd()
+{
+    m_aboutToScrollPastEnd = false;
+
+    /* m_completer.popup()->currentIndex() being invalid means the list
+       has been scrolled from bottom to top. We undo this here, as it avoids
+       scroll bar flicker as well. Block signal to avoid recursion.
+    */
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -114,6 +115,19 @@ void SearchBarLineEdit::hideSuggestions()
     m_completer.popup()->hide();
 }
 
+bool SearchBarLineEdit::eventFilter(QObject *, QEvent *event)
+{
+    if (auto e = static_cast<QKeyEvent *>(event))
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -177,6 +177,9 @@ void SearchBarLineEdit::fetchMoreSuggestions()
 
 void SearchBarLineEdit::onScroll(int value)
 {
+    if (!m_completer.popup()->currentIndex().isValid())
+        onScrollPastEnd();
```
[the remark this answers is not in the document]
ShaopengLin: Done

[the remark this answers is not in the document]
veloman-yunkan: > 1. You must `return` here - the `return` statement inside `onScrollPastEnd()` doesn't return out of this function.

According to [the summary of changes relative to the previous review](https://github.com/kiwix/kiwix-desktop/compare/b6a6c06025035196ade818f9756b5a182401aeb2..1bc6c8218991ac33acf2f1afbce7a92a2a4e5385) above bug was not fixed

src/searchbar.cpp
```diff
@@ -177,19 +177,21 @@ void SearchBarLineEdit::fetchMoreSuggestions()
...
-
     if (m_aboutToScrollPastEnd)
     {
         if (scrolledToEnd)
-        {
             /* The user's intention to scroll past end has been confirmed */
-            fetchMoreSuggestions();
-            m_aboutToScrollPastEnd = false; /* Relax until next time */
-        }
+            onScrollPastEnd();
```
veloman-yunkan: Please restore the braces - those may be omitted only if the body is a single liner but comments count as part of a body!

src/searchbar.cpp
```diff
@@ -114,6 +115,22 @@ void SearchBarLineEdit::hideSuggestions()
     m_completer.popup()->hide();
 }
 
+bool SearchBarLineEdit::eventFilter(QObject *, QEvent *event)
+{
+    if (!m_aboutToScrollPastEnd)
+        return false;
+
+    if (auto e = dynamic_cast<QKeyEvent *>(event))
```
veloman-yunkan, line 130: This time I checked [the documentation of `QObject::eventFilter()`](https://doc.qt.io/qt-5/qobject.html#eventFilter) as I was curious about the need for its first parameter). I saw in the example that a `dynamic_cast` is not used because there is a cheaper way - `QEvent::type()`.

src/searchbar.cpp
```diff
@@ -114,6 +115,22 @@ void SearchBarLineEdit::hideSuggestions()
...
+    {
+        if (e->key() == Qt::Key_PageDown && e->modifiers().testFlag(Qt::NoModifier))
```
veloman-yunkan, line 132: Now I wonder if it's better to handle key down here too (instead of your previous fix for it). Won't it eliminate the need to undo the handling of wrapping from bottom to top? If yes then the final code will be simpler, and so let's do it.

src/searchbar.cpp
```diff
@@ -208,16 +208,16 @@ void SearchBarLineEdit::onScroll(int value)
...
     }
 
-    /* m_completer.popup()->currentIndex() being invalid means the user used the
+    /* m_suggestionView->currentIndex() being invalid means the user used the
        keyboard up/down to scroll past the end.
     */
-    if (!m_completer.popup()->currentIndex().isValid())
-        return onScrollPastEnd();
+    if (!m_suggestionView->currentIndex().isValid())
+        onScrollPastEnd();
```
veloman-yunkan: The `return` statement was lost (though if key down is handled in `SearchBarLineEdit::eventFilter()` this piece of code will simply have to be deleted).

src/searchbar.cpp
```diff
@@ -208,16 +208,16 @@ void SearchBarLineEdit::onScroll(int value)
...
 
-    const auto suggestionScroller = m_completer.popup()->verticalScrollBar();
-    const auto scrollMin = suggestionScroller->minimum();
-    const auto scrollMax = suggestionScroller->maximum();
-    const bool scrolledToEnd = value == suggestionScroller->maximum();
+    auto suggestionScroller = m_suggestionView->verticalScrollBar();
+    auto scrollMin = suggestionScroller->minimum();
+    auto scrollMax = suggestionScroller->maximum();
+    bool scrolledToEnd = value == suggestionScroller->maximum();
```
veloman-yunkan: The value brought by `const`s should be long lasting rather than ephemeral (limited to a few commits).

src/searchbar.cpp
```diff
@@ -165,6 +170,44 @@ void SearchBarLineEdit::updateCompletion()
     fetchSuggestions(&SearchBarLineEdit::onInitialSuggestions);
 }
 
+void SearchBarLineEdit::fetchMoreSuggestions()
+{
+    fetchSuggestions(&SearchBarLineEdit::onAdditionalSuggestions);
+}
```
veloman-yunkan, line 212: This encapsulates a performance bug. `fetchSuggestions()` creates a temporary `SuggestionListWorker` which creates  `zim::SuggestionSearcher` that is discarded after a single use. In a more optimal implementation the `zim::SuggestionSearcher` should be reused for loading additional suggestions (since those incremental calls to `zim::SuggestionSearcher::getResults()` should be much cheaper compared to the first call). I am not implying that that optimization be implemented in this PR. I propose to do it myself after this PR is merged, but if you feel like becoming married to this suggestions feature and would like to take care of it to the end I will humbly stay away :)

veloman-yunkan (changes requested):
An issue was still left because of hasty conflict resolution. Since you have to fix it, please address a few other comments as well.

src/searchbar.cpp
ShaopengLin: So it seems the scrolling event from Key_down either reaches the popup first or isn't blocked from reaching the popup. I simply removed onScrollPastEnd and put the scrolling only inside the onScroll `!m_completer.popup()->currentIndex().isValid()` case.

src/searchbar.cpp
ShaopengLin: Let's do this separately.

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -177,6 +199,17 @@ void SearchBarLineEdit::fetchMoreSuggestions()
...
+    /* Scrolling using key_down past end will teleport scroller to the top.
+       We undo this here. Block signal to avoid recursion.
+    */
+    if (!m_completer.popup()->currentIndex().isValid())
+    {
+        auto old = m_completer.popup()->verticalScrollBar()->blockSignals(true);
+        m_completer.popup()->scrollToBottom();
+        m_completer.popup()->verticalScrollBar()->blockSignals(old);
+        return;
+    }
```
veloman-yunkan: The need for this is not clear. It almost defies the uniform handling of scrolling through suggestions with the keyboard.

Maybe you should install the event filter on `m_completer.popup()` instead of on (or in addition to) `SearchBarLineEdit`?

If you cannot find a way to eliminate this piece of code (but not at the expense of similar hack!) then please add a comment describing that there is something wrong (or something that we don't understand) happening here.

src/searchbar.cpp
veloman-yunkan: Definitely. But please add a TODO comment in the body of `SearchBarLineEdit::fetchMoreSuggestions()` in this PR.

veloman-yunkan (commented):
This is the last but one review.

So far I haven't run/tested the code of this PR, but I will do it before I approve during the next iteration. Please test the PR yourself too before requesting final review, and update the description of the PR accordingly (if there are known issues that are related to this PR but weren't supposed to be fixed here please list them in the description).

src/searchbar.cpp
```diff
@@ -114,6 +115,27 @@ void SearchBarLineEdit::hideSuggestions()
...
+bool SearchBarLineEdit::eventFilter(QObject *, QEvent *event)
+{
+    if (!m_aboutToScrollPastEnd)
+        return false;
+
+    if (auto e = dynamic_cast<QKeyEvent *>(event))
+    {
+        auto key = e->key();
+        bool isScrollDownKey = key == Qt::Key_Down || key == Qt::Key_PageDown;
+        bool noModifiers = e->modifiers().testFlag(Qt::NoModifier);
```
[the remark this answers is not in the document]
ShaopengLin: It will take some time to be a natural. It took me a while to get muscle memory for the member function const :)

src/searchbar.cpp
ShaopengLin: I cannot pinpoint the source of what triggers the scrolling (been this way when I first started tackling the Issue), so unfortunately this might be here to stay. I honestly suspect it might be an internal private class at this point, and it is rather confusing...

src/searchbar.cpp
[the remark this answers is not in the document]
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

veloman-yunkan: I deleted this piece of code and nothing seems to be broken. Maybe the bug is fixed by a later change in the PR (e.g. measures taken against flickering)?

src/searchbar.cpp
```diff
@@ -164,40 +208,132 @@ void SearchBarLineEdit::updateCompletion()
...
-    connect(suggestionWorker, &SuggestionListWorker::searchFinished, this,
-    [=] (const QStringList& suggestions, const QVector<QUrl>& urlList, int token) {
-        if (token != m_token) {
-            return;
+    fetchSuggestions(&SearchBarLineEdit::onInitialSuggestions);
+}
+
+void SearchBarLineEdit::fetchMoreSuggestions()
+{
+    /* TODO: Refactor suggestion worker to use single zim::SuggestionSearcher */
```
veloman-yunkan, line 216: This comment doesn't make much sense. A superficial review of the `SuggestionListWorker` will find a single instance of `zim::SuggestionSearcher`. Please explain the problem with this implementation so that an outsider (who didn't follow this PR) understands what exactly has to be addressed.

src/searchbar.cpp
```diff
@@ -297,7 +297,9 @@ QModelIndex SearchBarLineEdit::getDefaulSuggestionIndex() const
     const auto firstSuggestionText = firstSuggestionIndex.data().toString();
```
veloman-yunkan, line 331: I appreciate the added `const` however while resolving merge conflicts during rebase you lost the the update to the above comment.

veloman-yunkan: See https://github.com/kiwix/kiwix-desktop/compare/07aa791ed7844595451c25b5bc313b9fd80b1923..80acbcaeb601eeb4813dbe6bd4e1fe66352b39b9

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
ShaopengLin: Done

veloman-yunkan:
What about https://github.com/kiwix/kiwix-desktop/pull/1224#discussion_r1807765018?

src/searchbar.cpp
ShaopengLin: The code is more for the 'smoothness' of the fetching when using Key_Down. You might notice when the Key_Down triggers the fetching, the scroller jumps up then down, which creates a flicker. If you add this piece of code, it smoothly extends the suggestions. It is should be the same smoothness as the Key_PageDown.

You are correct that functionally it still achieves the end result.

veloman-yunkan (approved):
It took longer than expected but we finally did it. Let's move on to the follow up PRs and try to reduce the latency of each review iteration to have them merged faster.

src/searchbar.cpp
veloman-yunkan: Please don't start working on this until the rest of your `kiwix-desktop` PRs are merged.

veloman-yunkan:
This PR broke the handling of pressing ENTER in the search bar - it always opens the first suggestion (even if it doesn't fully match the text in the searchbar).

veloman-yunkan:
> @veloman-yunkan I believe its mainly from this [#1224 (comment)](https://github.com/kiwix/kiwix-desktop/pull/1224#discussion_r1796894831) where we discarded the checks for text comparison inside openCompletion.

I cannot agree. I don't see any flaw with the design of that approach. `getDefaultSuggestionIndex()` picks up the suggestion that needs to be opened and `openCompletion()` just opens it. It looks like there is a bug in `getDefaultSuggestionIndex()` but it's not immediately clear what's wrong with it.

ShaopengLin:
@veloman-yunkan getDefaultSuggestionIndex is only called during typing. openCompletion is also signaled normalled when user activate(click/enter) on a cell when not typing. openCompletion itself does not use getDefaultSuggestionIndex.

veloman-yunkan:
> @veloman-yunkan getDefaultSuggestionIndex is only called during typing. openCompletion is also signaled normalled when user activate(click/enter) on a cell when not typing. openCompletion itself does not use getDefaultSuggestionIndex.

@ShaopengLin Yes. But when an item from the list is activated `openCompletion()` is called with its exact index. `getDefaultSuggestionIndex()` is (or, rather, should be) invoked only when the ENTER key is pressed with no item selected.

veloman-yunkan:
The bug is that `openCompletion(getDefaulSuggestionIndex());` is called only from `onInitialSuggestions()`. It works as intended only if ENTER is pressed *before* suggestions are generated.

kelson42:
@veloman-yunkan Please open a dedicated issue, not very much cinfirtable with discussion on closed issues/PRa.

veloman-yunkan:
> @veloman-yunkan Please open a dedicated issue, not very much cinfirtable with discussion on closed issues/PRa.

Done (#1230)

PR #1232, opened 2024-10-30
BugFix: Should Open Default Suggestion on Search Bar when Nothing is Selected

### LINKED ISSUE #1230: Pressing ENTER in the search bar opens the first suggestion
#1224 broke the handling of pressing ENTER in the search bar - it always opens the first suggestion (even if it doesn't fully match the text in the searchbar).

_Originally posted by @veloman-yunkan in https://github.com/kiwix/kiwix-desktop/issues/1224#issuecomment-2435785981_

ShaopengLin:
Fix #1230

Two potential implementations listed here:
  
  1. Open default suggestions whenever enter is pressed on an invalid index. Added 'relevant' search bar input checker to avoid race conditions. 2929d91469bb3a0c493ff2651410cc5e62a9af7a
  
  2. Similar approach as before #1224, where we check text in search bar matches the suggestion text before opening. ad2e2b323c4e49be95f07413e6f44f2f56ee6b95

I am leaning toward the first one since it is a much cleaner change.

### DIFF
--- a/src/searchbar.cpp
+++ b/src/searchbar.cpp
@@ -132,6 +132,15 @@ SearchBarLineEdit::SearchBarLineEdit(QWidget *parent) :
     });
     connect(this, &QLineEdit::returnPressed, this, [=]() {
         m_returnPressed = true;
+
+        /* Open default index when nothing is selected.
+        
+           Key_Return can be pressed during typing, where suggestions no longer
+           match the text typed. Hence the suggestionsValid check.
+         */
+        const bool suggestionsValid = m_suggestionsAreValidFor == m_searchbarInput;
+        if (!m_suggestionView->currentIndex().isValid() && suggestionsValid)
+            openCompletion(getDefaulSuggestionIndex());
     });
     
     auto app = KiwixApp::instance();
@@ -312,10 +321,9 @@ void SearchBarLineEdit::onInitialSuggestions(int)
     } else {
         m_completer.complete(getCompleterRect());
 
-        /* Make row 0 appear but do not highlight it */
-        const auto completerFirstIdx = m_suggestionView->model()->index(0, 0);
+        /* Select nothing by default */
         const auto completerSelModel = m_suggestionView->selectionModel();
-        completerSelModel->setCurrentIndex(completerFirstIdx, QItemSelectionModel::Current);
+        completerSelModel->setCurrentIndex(QModelIndex(), QItemSelectionModel::Current);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/searchbar.cpp
```diff
@@ -233,6 +242,7 @@ void SearchBarLineEdit::updateCompletion()
         return;
     }
     m_token++;
+    m_suggestionsAreValidFor = m_searchbarInput;
```
veloman-yunkan: Is this the correct place to set `m_suggestionsAreValidFor`? Shouldn't it rather be set in `onInitialSuggestions()`? Of course it will slightly complicate the solution since that information will have to be somehow passed to `onInitialSuggestions()`, however we cannot compromise correctness. A workaround may be to set it in the handler of the `SuggestionListWorker::searchFinished()` signal in `fetchSuggestions()`.

veloman-yunkan (changes requested):

src/searchbar.cpp
ShaopengLin: Done

src/searchbar.cpp
```diff
@@ -332,12 +340,13 @@ void SearchBarLineEdit::fetchSuggestions(NewSuggestionHandlerFuncPtr callback)
     const int start = m_suggestionModel.countOfRegularSuggestions();
     const auto suggestionWorker = new SuggestionListWorker(m_searchbarInput, m_token, start, this);
     connect(suggestionWorker, &SuggestionListWorker::searchFinished, this,
-            [=] (const QList<SuggestionData>& suggestionList, int token) {
+            [=] (const QList<SuggestionData>& suggestionList, const QString& text, int token) {
```
veloman-yunkan: You could have passed the search text as a captured value of the lambda. But your implementation drew my attention to the not so self-explaining `token`/`m_token` thing. Let's improve the design as follows:

1. Replace  `int SearchBarLineEdit::m_token` with `QString SearchBarLineEdit::m_suggestionWereRequestedFor` (or maybe we can instead reuse the existing `m_searchbarInput` data member?)
2. Get rid of the `token` parameter in `searchFinished()` signal of `SuggestionListWorker` - use the new `text` parameter instead (BTW, the latter better be renamed to `searchText` and moved to the beginning of the parameter list).

veloman-yunkan (changes requested):

src/searchbar.cpp
ShaopengLin: @veloman-yunkan Removing `m_token` might introduce a race condition bug:
1. user type "i", and suggestions are retrived.
2. user trigger fetch more.
3. user further type to "i am", and new suggestions are triggered.
4. user immediately types back to "i", and the fetch more from 2. returns from asynchronous call back.
5. the insertion order can be raced, and m_completer.complete() is not called before fetch more.

veloman-yunkan: Good analysis! You are a shrewd antiracist! :)

src/searchbar.cpp
```diff
@@ -313,9 +322,8 @@ void SearchBarLineEdit::onInitialSuggestions(int)
         m_completer.complete(getCompleterRect());
 
         /* Make row 0 appear but do not highlight it */
```
veloman-yunkan: Is this comment still valid?

veloman-yunkan (commented):
If your preference is still in favour of the first approach then please remove the last two commits and we will be done with this PR.

src/searchbar.cpp
ShaopengLin: Done

veloman-yunkan (approved):

PR #1201, opened 2024-09-09
Introduce Table of Content

ShaopengLin:
Fix #42 

Changes:
- Inject to pages a Table of Contents comprised of a nested hierarchy list of hyperlinks to <h> tags in the page.
- Add a button in the corner to toggle the table.
- No button and table are displayed if the table is empty.

PR #1237, opened 2024-11-07
Introduce Table of Content Without Intense JS Invasion

### LINKED ISSUE #42: Implement table of content

ShaopengLin:
Alternative to #1201. FIx #42 

Changes:
- Uses QTreeWidget to handle TOC display
- The TOC now resides in the sidebar along with the Content Manager Side and ReadingList Bar.
- Styling is slightly different since QTreeWidget doesn't perform the same way as HTML and JS.

Benefits:
- More extensible when future features come
- Easier handling inside Qt for styling and parsing.

Loses:
- The TOC requesting is asynchronous and requires some fine-tuned race checking.
- Slow due to extra parsing and passing around of the TOC.

### DIFF
--- a/src/kiwixwebchannelobject.h
+++ b/src/kiwixwebchannelobject.h
@@ -9,6 +9,11 @@ class KiwixWebChannelObject : public QObject
 
 public:
     explicit KiwixWebChannelObject(QObject *parent = nullptr) : QObject(parent) {};
+
+    Q_INVOKABLE void sendHeaders(const QJsonObject& headers) { emit headersChanged(headers); };
+
+signals:
+    void headersChanged(const QJsonObject& headers);
 };
 
 #endif // KIWIXWEBCHANNELOBJECT_H
--- a/src/kiwixwebchannelobject.h
+++ b/src/kiwixwebchannelobject.h
@@ -1,0 +1,14 @@
+#ifndef KIWIXWEBCHANNELOBJECT_H
+#define KIWIXWEBCHANNELOBJECT_H
+
+#include <QObject>
+
+class KiwixWebChannelObject : public QObject
+{
+    Q_OBJECT
+
+public:
+    explicit KiwixWebChannelObject(QObject *parent = nullptr) : QObject(parent) {};
+};
+
+#endif // KIWIXWEBCHANNELOBJECT_H
--- a/src/kiwixwebchannelobject.h
+++ b/src/kiwixwebchannelobject.h
@@ -14,6 +14,7 @@ public:
 
 signals:
     void headersChanged(const QJsonObject& headers);
+    void navigationRequested(const QString& url, const QString& anchor);
 };
 
 #endif // KIWIXWEBCHANNELOBJECT_H
--- a/src/kprofile.cpp
+++ b/src/kprofile.cpp
@@ -58,6 +58,8 @@ KProfile::KProfile(QObject *parent) :
 #endif
 
     scripts()->insert(getScript(":/js/headerAnchor.js"));
+    scripts()->insert(getScript(":/qtwebchannel/qwebchannel.js", 
+                      QWebEngineScript::DocumentCreation));
 }
 
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
--- a/src/kprofile.cpp
+++ b/src/kprofile.cpp
@@ -4,6 +4,26 @@
 #include <QFileDialog>
 #include <QMessageBox>
 #include <QWebEngineSettings>
+#include <QWebEngineScript>
+#include <QWebEngineScriptCollection>
+
+namespace
+{
+
+QWebEngineScript getScript(QString filename,
+    QWebEngineScript::InjectionPoint point = QWebEngineScript::DocumentReady)
+{
+    QWebEngineScript script;
+    script.setInjectionPoint(point);
+    script.setWorldId(QWebEngineScript::UserWorld);
+
+    QFile scriptFile(filename);
+    scriptFile.open(QIODevice::ReadOnly);
+    script.setSourceCode(scriptFile.readAll());
+    return script;
+}
+
+}
 
 QString askForSaveFilePath(const QString& suggestedName)
 {
@@ -36,6 +56,8 @@ KProfile::KProfile(QObject *parent) :
 #else // Qt 5.13 and later
     setUrlRequestInterceptor(new ExternalReqInterceptor(this));
 #endif
+
+    scripts()->insert(getScript(":/js/headerAnchor.js"));
 }
 
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
--- a/src/tableofcontentbar.cpp
+++ b/src/tableofcontentbar.cpp
@@ -9,7 +9,9 @@ TableOfContentBar::TableOfContentBar(QWidget *parent) :
     ui(new Ui::tableofcontentbar)
 {
     ui->setupUi(this);
+    ui->titleLabel->setFont(QFont("Selawik", 18, QFont::Weight::Medium));
     ui->titleLabel->setText(gt("table-of-content"));
+    ui->hideLabel->setFont(QFont("Selawik", 12));
     ui->hideLabel->setTextFormat(Qt::RichText);
 
     /* href is needed to make hide clickable, but not used. So Kiwix it is :) */
@@ -44,6 +46,7 @@ QTreeWidgetItem* createChildItem(QTreeWidgetItem* parent, const QString& childNo
 
     const auto display = childNo + "  " + headerObj["text"].toString();
     item->setData(0, Qt::DisplayRole, display);
+    item->setData(0, Qt::FontRole, QFont("Selawik", 12));
     item->setData(0, Qt::UserRole, headerObj["anchor"].toString());
 
     return item;
--- a/src/tableofcontentbar.cpp
+++ b/src/tableofcontentbar.cpp
@@ -12,11 +12,18 @@ TableOfContentBar::TableOfContentBar(QWidget *parent) :
     ui->titleLabel->setText(gt("table-of-content"));
     ui->tree->setRootIsDecorated(false);
     ui->tree->setItemsExpandable(false);
+    connect(ui->tree, &QTreeWidget::itemClicked, this, &TableOfContentBar::onTreeItemActivated);
+    connect(ui->tree, &QTreeWidget::itemActivated, this, &TableOfContentBar::onTreeItemActivated);
 }
 
 TableOfContentBar::~TableOfContentBar()
 {
     delete ui;
+}
+
+void TableOfContentBar::onTreeItemActivated(QTreeWidgetItem *item)
+{
+    emit navigationRequested(m_url, item->data(0, Qt::UserRole).toString());
 }
 
 namespace
@@ -27,6 +34,7 @@ QTreeWidgetItem* createChildItem(QTreeWidgetItem* parent, const QJsonObject& hea
     const auto item = new QTreeWidgetItem(parent);
     item->setExpanded(true);
     item->setData(0, Qt::DisplayRole, headerObj["text"].toString());
+    item->setData(0, Qt::UserRole, headerObj["anchor"].toString());
 
     return item;
 }
@@ -68,6 +76,7 @@ void TableOfContentBar::setupTree(const QJsonObject& headers)
     if (headerUrl != currentUrl)
         return;
     
+    m_url = headerUrl;
     ui->tree->clear();
     QJsonArray headerArr = headers["headers"].toArray();
     createSubTree(ui->tree->invisibleRootItem(), headerArr);
--- a/src/tableofcontentbar.cpp
+++ b/src/tableofcontentbar.cpp
@@ -10,6 +10,14 @@ TableOfContentBar::TableOfContentBar(QWidget *parent) :
 {
     ui->setupUi(this);
     ui->titleLabel->setText(gt("table-of-content"));
+    ui->hideLabel->setTextFormat(Qt::RichText);
+
+    /* href is needed to make hide clickable, but not used. So Kiwix it is :) */
+    ui->hideLabel->setText("<a href=\"https://kiwix.org/en/\">" + gt("hide") + "</a>");
+    connect(ui->hideLabel, &QLabel::linkActivated, this, [=](){
+        KiwixApp::instance()->getAction(KiwixApp::ToggleTOCAction)->setChecked(false);
+    });
+
     ui->tree->setRootIsDecorated(false);
     ui->tree->setItemsExpandable(false);
     connect(ui->tree, &QTreeWidget::itemClicked, this, &TableOfContentBar::onTreeItemActivated);
--- a/src/tableofcontentbar.cpp
+++ b/src/tableofcontentbar.cpp
@@ -29,11 +29,13 @@ void TableOfContentBar::onTreeItemActivated(QTreeWidgetItem *item)
 namespace
 {
 
-QTreeWidgetItem* createChildItem(QTreeWidgetItem* parent, const QJsonObject& headerObj)
+QTreeWidgetItem* createChildItem(QTreeWidgetItem* parent, const QString& childNo, const QJsonObject& headerObj)
 {
     const auto item = new QTreeWidgetItem(parent);
     item->setExpanded(true);
-    item->setData(0, Qt::DisplayRole, headerObj["text"].toString());
+
+    const auto display = childNo + "  " + headerObj["text"].toString();
+    item->setData(0, Qt::DisplayRole, display);
     item->setData(0, Qt::UserRole, headerObj["anchor"].toString());
 
     return item;
@@ -54,15 +56,16 @@ QJsonArray takeDeeperEntries(QJsonArray& headerArr, int level)
     return result;
 }
 
-void createSubTree(QTreeWidgetItem* parent, QJsonArray& headerArr)
+void createSubTree(QTreeWidgetItem* parent, QString parentNo, QJsonArray& headerArr)
 {
     while (!headerArr.isEmpty())
     {
         const auto childHeader = headerArr.takeAt(0).toObject();
         const int childLevel = childHeader["level"].toInt();
-        QTreeWidgetItem* childItem = createChildItem(parent, childHeader);
+        const QString childNo = parentNo + QString::number(parent->childCount() + 1);
+        QTreeWidgetItem* childItem = createChildItem(parent, childNo, childHeader);
         QJsonArray deeperEntries = takeDeeperEntries(headerArr, childLevel);
-        createSubTree(childItem, deeperEntries);
+        createSubTree(childItem, childNo + ".", deeperEntries);
     }
 }
 
@@ -79,5 +82,5 @@ void TableOfContentBar::setupTree(const QJsonObject& headers)
     m_url = headerUrl;
     ui->tree->clear();
     QJsonArray headerArr = headers["headers"].toArray();
-    createSubTree(ui->tree->invisibleRootItem(), headerArr);
+    createSubTree(ui->tree->invisibleRootItem(), "", headerArr);
 }
--- a/src/tableofcontentbar.cpp
+++ b/src/tableofcontentbar.cpp
@@ -1,0 +1,16 @@
+#include "tableofcontentbar.h"
+#include "ui_tableofcontentbar.h"
+#include "kiwixapp.h"
+
+TableOfContentBar::TableOfContentBar(QWidget *parent) :
+    QFrame(parent),
+    ui(new Ui::tableofcontentbar)
+{
+    ui->setupUi(this);
+    ui->titleLabel->setText(gt("table-of-content"));
+}
+
+TableOfContentBar::~TableOfContentBar()
+{
+    delete ui;
+}
--- a/src/tableofcontentbar.cpp
+++ b/src/tableofcontentbar.cpp
@@ -45,6 +45,7 @@ QTreeWidgetItem* createChildItem(QTreeWidgetItem* parent, const QString& childNo
     item->setExpanded(true);
 
     const auto display = childNo + "  " + headerObj["text"].toString();
+    item->setToolTip(0, display);
     item->setData(0, Qt::DisplayRole, display);
     item->setData(0, Qt::FontRole, QFont("Selawik", 12));
     item->setData(0, Qt::UserRole, headerObj["anchor"].toString());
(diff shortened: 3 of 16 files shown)


veloman-yunkan:
Bug report - when the TOC is not empty the attempt to open the TOC sidebar results in the latter growing wider and wider indefinitely.

ShaopengLin:
@veloman-yunkan I can't seem to replicate this. Is this happening on any zim?

resources/js/tableofcontent.js
```diff
@@ -1,3 +1,75 @@
...
+    var levelArray = Array.from(recurseData.levelSet).sort();
+    levelArray.sort(function(a, b){return a - b});
+    var level = levelArray.indexOf(recurseData.level) + 1; 
+
+    /* End un-closed lists */
+    if (level)
+        recurseData.toc += (new Array(level + 1)).join(']}');
+    recurseData.toc += ']}';
+    return recurseData.toc;
+}
```
veloman-yunkan, line 70: I think that instead of constructing the final JSON string it is better to construct a recursive JS object and then call `JSON.stringify()` on it.

resources/js/tableofcontent.js
```diff
@@ -1,3 +1,75 @@
...
+{
+    if (elem !== "undefined") 
+    {
+        if(elem.nodeName.match(/^H\d+$/) && elem.textContent)
+        {
+            var headerText = elem.textContent.trim();
+            var prevLevel = recurseData.level;
+            var level = elem.nodeName.substr(1);
+            var anchor = "kiwix-toc-" + recurseData.count;
+            var anchorLink = window.location.href.replace(location.hash,"") + '#' + anchor;
```
veloman-yunkan, line 20: I think that it doesn't make sense to include the full URL with every link. The URL of the page can be provided only once at the root level of the returned JSON object while each TOC entry will include only the anchor.

resources/js/tableofcontent.js
```diff
@@ -72,4 +72,10 @@ function tocJSON()
 new QWebChannel(qt.webChannelTransport, function(channel) {
     var kiwixObj = channel.objects.kiwixChannelObj;
     kiwixObj.sendTableOfContent(tocJSON());
+    kiwixObj.navigationRequested.connect(function(anchor) {
+        var anchorIndex = anchor.lastIndexOf('#');
+        var anchorPageUrl = anchor.substr(0, anchorIndex);
+        if (window.location.href.replace(location.hash,"") === anchorPageUrl)
+            window.location.href = anchor;
+    })
```
veloman-yunkan, line 80: This comment is about the full commit - can you elaborate on the scenario that is problematic if the navigation is implemented in C++? Is your solution fully devoid of similar problems or they are just made less embarrassing for the UX?

veloman-yunkan:
> @veloman-yunkan I can't seem to replicate this. Is this happening on any zim?

It happens on ray_charles.zim for sure

veloman-yunkan:
[Screencast from 11-07-2024 08:22:37 PM.webm](https://github.com/user-attachments/assets/2c85f6df-da75-4961-bcc3-aedc59c7f02c)

ShaopengLin:
@veloman-yunkan Would like to confirm if you are on Qt6? I have tested Qt5 and Qt6, where Qt6 seems to be the version that this bug happens

veloman-yunkan (commented):
Here is the initial feedback from a quick and very superficial review

veloman-yunkan:
> @veloman-yunkan Would like to confirm if you are on Qt6? I have tested Qt5 and Qt6, where Qt6 seems to be the version that this bug happens

No, I am a qtsaur. I don't use the most recent version.

veloman-yunkan:
> * Slow due to extra parsing and passing around of the TOC.

Is this assessment based on your understanding of the implementation or a result of some measurement?

ShaopengLin:
> Bug report - when the TOC is not empty the attempt to open the TOC sidebar results in the latter growing wider and wider indefinitely.

@veloman-yunkan Should be fixed now. Somehow under the same settings, QListWidget in ReadingListBar doesn't have this problem. This might be due to either styling or underlying widget differences.

resources/js/tableofcontent.js
ShaopengLin: Ok, I thought I replaced this commit already... I previously forgot the setUrl method and was using this method. Another victim from rebasing.

src/kprofile.cpp
```diff
@@ -4,6 +4,21 @@
...
+{
+    QWebEngineScript script;
+    script.setInjectionPoint(point);
+    script.setWorldId(QWebEngineScript::UserWorld);
+
+    QFile scriptFile(filename);
+    scriptFile.open(QIODevice::ReadOnly);
+    script.setSourceCode(scriptFile.readAll());
+    return script;
+}
```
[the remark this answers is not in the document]
ShaopengLin: Done

src/webview.cpp
```diff
@@ -105,6 +105,26 @@ WebView::WebView(QWidget *parent)
...
+                */
+                if (tabbar->currentWebView() == this && tableValid)
+                    emit tableOfContentChanged(m_tableOfContent);
+            });
+    connect(kiwixChannelObj, &KiwixWebChannelObject::tableOfContentChanged, this, 
+            [=](const QString& tableJson){
+                m_tableOfContent = QJsonDocument::fromJson(tableJson.toUtf8());
+                if (tabbar->currentWebView() == this)
+                    emit tableOfContentChanged(m_tableOfContent);
+            });
```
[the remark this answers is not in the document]
ShaopengLin: Done

resources/js/tableofcontent.js
ShaopengLin: Done

resources/js/tableofcontent.js
ShaopengLin: Done

resources/js/tableofcontent.js
ShaopengLin: Done

veloman-yunkan:
> > Is this assessment based on your understanding of the implementation or a result of some measurement?
> 
> @veloman-yunkan It is more like my own understanding and observation from other Stackoverflowers:
> 
>     1. QTreeWidget have seen performance issues regarding larger data amounts: [1](https://stackoverflow.com/questions/72377375/qtreewidget-fast-updating), [2](https://stackoverflow.com/questions/62516121/using-qteeview-instead-of-qtreewidget)

Do we ever expect "large data amounts" for TOC?

>     2. In Javascript, instead of creating JSON and doing another round of parsing,  we can just create a `<ul>` `<li>` list on the first round and immediately append this to the sidebar.

Back to the previous comment, if the TOC is not going to be overwhelmingly large this shouldn't matter a lot.

>     3. In Qt, everytime we go to a new page, the TOC needs to reload. While If we display using Javascript, the TOC is already in the DOM and no need to reload.

How come the TOC is already in the DOM when we go to a new page? Did you mean switching tabs?

veloman-yunkan:
The entries in the TOC need to be double clicked for the navigation to happen. Was it intended to be like that?

ShaopengLin:
@veloman-yunkan
1. I don't expect the average TOC to be large.
2. Yes indeed.
3. Yes I meant switching tabs.

I do believe this PR is better than #1201. The loses I mentioned are more for acknowledging some minor inconveniences we face. In the grand scheme of things, it is not closr enough to offset the benefits by TOC to Qt side.

ShaopengLin:
> The entries in the TOC need to be double clicked for the navigation to happen. Was it intended to be like that?

@veloman-yunkan Its supposed to be single click. I will change this.

ShaopengLin:
> Do I get it right that in this implementation TOC subsections are collapsible while in the JS-based implementation they aren't?

@veloman-yunkan In JS they weren't collapsible.

kelson42:
@ShaopengLin @veloman-yunkan Don't try to handle collabsable sections in the content. Focus only on the behaviour of the TOC sidebar.

resources/js/tableofcontent.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * Construct recurseData.toc as a JSON-like that contains each header's text
+ * anchor, and their list of child headers.
+ * 
+ * References:
+ * https://stackoverflow.com/questions/187619/is-there-a-javascript-solution-to-generating-a-table-of-contents-for-a-page
+ * @param elem DOM element
+ * @param recurseData Object with fields: { toc : Object, count : int, parent : Object }
+ */
+function recurseChild(elem, recurseData)
```
veloman-yunkan, line 10: This (and any other global name introduced by this script) may conflict with/override the global variables/functions created in the webpage's JS world by its own scripts. There are ways to avoid that. At least you can limit the potential damage by packaging all your utilities into a single class/object. I believe there is a zero-pollution solution (nice rhyme!), too.

resources/js/tableofcontent.js
```diff
@@ -0,0 +1,75 @@
...
+{
+    if (elem !== "undefined")
```
veloman-yunkan, line 12: 1. IMHO, `undefined` shouldn't be quoted
2. Is this check needed here at all? I believe it has to be performed only before the top-level call of `recureChild()` (to guard against the missing `<body>`).

resources/js/tableofcontent.js
```diff
@@ -0,0 +1,75 @@
...
+                /* Same level as parent meaning we should be sibling with parent. */
+                parent.parent.table.push(headerEntry);
+                parent = {parent: parent.parent, table: headerEntry.table, level: level};
+            }
+            else
+            {
+                /* We are the children of parent. */
+                parent.table.push(headerEntry);
+                parent = {parent: parent, table: headerEntry.table, level: level};
+            }
```
veloman-yunkan, line 44: This seems more convoluted than it could be. But let's address the more important issues first.

veloman-yunkan (changes requested):
In this iteration my focus was on the injected JS script.

veloman-yunkan:
> I do believe this PR is better than #1201. The loses I mentioned are more for acknowledging some minor inconveniences we face. In the grand scheme of things, it is not closr enough to offset the benefits by TOC to Qt side.

I am now also inclined to consider this solution as a superior one. Hoping that we won't discover a major deficiency in this approach that will make us shift back to the JS-based version.

resources/js/tableofcontent.js
```diff
@@ -0,0 +1,75 @@
...
+            var headerText = elem.textContent.trim();
+            var level = elem.nodeName.substr(1);
+            var anchor = "kiwix-toc-" + recurseData.count;
+            recurseData.count += 1;
+
+            var anchorElem = document.createElement("a");
+            anchorElem.id = anchor;
+
+            /* Wrap header content with something we can reference. */
+            elem.insertAdjacentElement("afterbegin", anchorElem);
```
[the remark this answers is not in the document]
ShaopengLin: @veloman-yunkan Tab switching doesn't trigger insertion. Instead, whenever we use setUrl, tocJSON is triggered due to the need to reload the page. I will do a check in onTableOfContentRecieved to make sure we don't send a new TOC when we just did a navigation.

resources/js/tableofcontent.js
ShaopengLin: @veloman-yunkan So the scripts are added to the UserWorld (see `getScript()` in kprofile.cpp ) which doesn't conflict with any of the webpage JS scripts.

resources/js/tableofcontent.js
ShaopengLin: I made an attempt to refactor the code.

resources/js/tableofcontent.js
ShaopengLin: Done

resources/js/tableofcontent.js
```diff
@@ -0,0 +1,82 @@
...
+    /* parent field is the object's parent used for nest checking.
+       table field is the reference to the parent field's table array.
+       level field is the level of the parent field
+    */ 
+    recurseData.parent =  { parent: null, table: recurseData.toc.table, level: 0 };
+
+    if (document.body !== undefined) 
+        recurseChild(document.body, recurseData); 
+    return JSON.stringify(recurseData.toc);
+}
```
veloman-yunkan, line 73: I think the presence of two hierarchies (DOM and TOC) is a source of some confusion here. To the best of my understanding the DOM hierarchy is completely ignored while building the TOC hierarchy. Then the code could better be structured as follows:
1. Run depth first search (DFS) traversal of the DOM sending `<h*>` elements to a callback (or just build a list of those elements in DFS order).
2. Create the TOC hierarchy from the sequence of `<h*>` DOM elements (and at the same time create and insert the anchor elements).

Does it make (more) sense?

veloman-yunkan: Note that there are different variants of depth first search (pre-order, in-order, post-order) but the difference between them would only matter if we expect that a  `<h*>` element may contain other nested `<h*>` elements in its subtree. I think that pre-order DFS should be used.

ShaopengLin: Done

resources/js/tableofcontent.js
```diff
@@ -1,3 +1,63 @@
...
+
+    return { headerEntry: { text: headerText, anchor, table: [] }, level };
+}
+
+function getDOMHeadersPreorder(elem, array)
+{
+    if (isHeaderElement(elem))
+        array.push(elem);
+
+    Array.from(elem.children).forEach(child => getDOMHeadersPreorder(child, array));
```
veloman-yunkan, line 24: Why is this better than simple iteration (`for ( const child of elem.children ) { ... }`)? It might matter while you were modifying the DOM while iterating over it but now you can enjoy the separation of DOM traversal and DOM modification.

resources/js/tableofcontent.js
```diff
@@ -1,3 +1,63 @@
...
+}
```
veloman-yunkan, line 25: For such name of the function (though I would make it slightly more generic `getDOMElementsPreorderDFS(elem, pred)`) I would prefer a clearer functional style implementation:

```js
function getDOMElementsPreorderDFS(elem, pred)
{
    var result = [];
    if (pred(elem))
        result.push(elem);

    for ( const child of elem.children) {
        result = result.concat(getDOMHeadersPreorder(child));
    }
    return result;
}
```

It may be a little slower because of the extra concatenations, however that should be typically negligible compared to the full iteration of the DOM.

If you want to keep the optimized version you better rename it to `collectHeaderElements(elem, outputArray)`.

resources/js/tableofcontent.js
```diff
@@ -1,3 +1,63 @@
...
+
+function tocJSON()
+{
+    const toc = { url: window.location.href.replace(location.hash,""), table: [] };
+
+    if (document.body !== undefined)
+    {
+        const headers = [];
+        getDOMHeadersPreorder(document.body, headers);
+        createTOCFromHeaders(headers, toc);
```
veloman-yunkan, line 55: Why not `toc.table = createTOC(headers);`?

resources/js/tableofcontent.js
```diff
@@ -1,3 +1,63 @@
...
+            levelStack.pop();
+
+        if (levelStack.length === 0)
+            toc.table.push(headerEntry);
+        else
+            levelStack[levelStack.length - 1].table.push(headerEntry);
+
+        levelStack.push({ table: headerEntry.table, level });
+    }); 
+}
```
veloman-yunkan, line 45: What do you think about representing the TOC as a flat array of `{level, headerText, anchor}` objects and moving the conversion to hierarchical form to C++?

src/kiwixwebchannelobject.h
```diff
@@ -14,6 +14,7 @@ class KiwixWebChannelObject : public QObject
 
 signals:
     void tableOfContentChanged(const QString& tableJson);
+    void onNavigationRequested(const QString& tableUrl, const QString& anchor);
```
veloman-yunkan, line 17: `tableUrl` is not a good name

veloman-yunkan:
I think you can remove the Draft status from this PR. Although it may still take a few review iterations to have it merged we are now operating in the regular PR review mode.

resources/js/tableofcontent.js
ShaopengLin: It would likely be faster and JS array references are probably very confusing for C++ programmers...

I will do this in webview and cache it. I will also try to reduce the parsing cost by trying to see if webchannel can pass a JS object directly as JSON.

veloman-yunkan: Don't fall into the premature optimization trap. This is not the place where we should mind about performance too much. Simplicity of design and clarity of the code are more important. It is always beneficial having the simplest implementation in the history and then optimizing upon it if needed.

resources/js/tableofcontent.js
ShaopengLin: Done

resources/js/tableofcontent.js
ShaopengLin: Done

resources/js/tableofcontent.js
ShaopengLin: Done

src/kiwixwebchannelobject.h
ShaopengLin: Done

src/webview.cpp
```diff
@@ -198,7 +202,76 @@ void WebView::saveViewContent()
...
+}
+
+void WebView::onAnchoredHeadersRecieved(const QJsonObject& headers)
+{
+    const auto tabbar = KiwixApp::instance()->getTabWidget();
+
+    m_tableOfContent = QJsonObject(headers);
+
+    QJsonArray headerArr = m_tableOfContent["headers"].toArray();
+    m_tableOfContent["table"] = getHierarchicalTableOfContent(headerArr);
```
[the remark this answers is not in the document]
ShaopengLin: Yes indeed simpler, almost wondering why I haven't thought of this approach in the first place. Done and renamed commits and the js file to better fit its purpose.

src/tableofcontentbar.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+        const auto itemNum = numberList.join(".");
+        const auto display = itemNum + "  " + header["text"].toString();
+        item->setData(0, Qt::DisplayRole, display);
+        item->setData(0, Qt::FontRole, QFont("Selawik", 12));
+        item->setData(0, Qt::UserRole, header["anchor"].toString());
+        item->setToolTip(0, display);
+        populateItem(headerArr, item, headerLevel);
+    }
+}
```
veloman-yunkan, line 68: My intuition tells me that this combination of iteration and recursion is redundant (and to my taste a little confusing). I feel that this algorithm can be rewritten in a clearer way with iteration alone (you will need an additional nested loop for ascending up the hierarchy to the `QTreeWidgetItem` corresponding to the new level). To add more clarity please extract the code responsible for the creation and initialization of a new `QTreeWidgetItem` into a helper function of its own.

ShaopengLin: Done

src/tableofcontentbar.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        /* find correct parent level to insert */
+        while (headerLevel <= curParentItem->data(0, LevelRole).toInt())
+        {
+            curParentItem = curParentItem->parent();
+
+            /* Unfortunate that top-level items not parented by invisibleRoot */
+            if (!curParentItem)
+                curParentItem = ui->tree->invisibleRootItem();
+        }
```
veloman-yunkan, line 96: What are the assumptions about level values in the header data? Can there be gaps like in the following sequence: `{h1, h3, h4, h4, h2, h4}` ? Does your algorithm handle such cases? Otherwise, assuming absence of gaps i.e. all levels on a branch from the root to the leaf are filled, why do you need `LevelRole`?

ShaopengLin: Yes there can be gaps. During parsing, we use their level to determine the correct parent, but how the tree is constructed  as if there are no gaps ( i.e. `{ h1, h3 }` will be repesented in the list as if it is `{ h1, h2 }`). Having perfect headers is an unlikely assumption...

veloman-yunkan: Well, I see now that probably under such conditions combining recursion with iteration makes more sense. Without recursion you have to store in the tree items intermediate data like level and the number-list (this one was present in your previous implementation too). Here is a cleaner and more comprehensible scheme:

```c++
QJsonArray takeDeeperEntries(QJsonArray& headerArr, int level)
{
    QJsonArray result;
    while (!headerArr.isEmpty() ) {
        const auto& nextHeader = headerArr.first().toObject();
        if ( nextHeader["level"].toInt() <= level )
           break;
        result.push_back(nextHeader);
        headerArr.pop_front();
    }
    return result;
}

void createSubTree(QTreeWidgetItem* parent, QString parentNo, QJsonArray headerArr)
{
    while (!headerArr.isEmpty()) {
        const auto& childHeader = headerArr.takeAt(0).toObject();
        const int childLevel = childHeader["level"].toInt();
        const QString childNo = parentNo + "." + QString::number(parent->childCount() + 1);
        QTreeWidgetItem* childItem = createChildItem(parent, childNo, childHeader);
        const QJsonArray deeperStuff = takeDeeperEntries(headerArr, childLevel);
        setupSubTree(childItem, childNo, deeperStuff);
    }
}
```

ShaopengLin: Done

src/webview.h
```diff
@@ -67,12 +69,15 @@ public slots:
 
 private slots:
     void gotoTriggeredHistoryItemAction();
+    void onCurrentTitleChanged();
+    void onHeadersRecieved(const QJsonObject& headers);
```
veloman-yunkan, line 74: recieve -> receive

veloman-yunkan (commented):
LGTM.

Since C++ is not a typo-safe language the compiler didn't detect the contamination of this PR by a subtle spelling error, however I believe that no user will be affected if we merge the PR as is.

A couple of comments regarding functionality:
1. navigation via TOC isn't recorded in the history.
2. when a section in the content is collapsed clicking a TOC entry belonging to that section has no effect

veloman-yunkan:
> A couple of comments regarding functionality:
> 
> 1. navigation via TOC isn't recorded in the history.
> 
> 2. when a section in the content is collapsed clicking a TOC entry belonging to that section has no effect

I believe those can be addressed in a follow-up PR (if they need to be fixed at all).

src/webview.h
ShaopengLin: Done

kelson42:
> > A couple of comments regarding functionality:
> > 
> > 1. navigation via TOC isn't recorded in the history.
> > 2. when a section in the content is collapsed clicking a TOC entry belonging to that section has no effect
> 
> I believe those can be addressed in a follow-up PR (if they need to be fixed at all).

Don't forget to open tickets!

veloman-yunkan (approved):

PR #1246, opened 2024-11-13
BugFix: Prevent crashes from tab switches due to Table Of Content

ShaopengLin:
Bug: Crashes can happen to the TOC management when the user switches tabs. Qt's QJsonObject copies are unreliable (though they shouldn't). I am uncertain as to why, but it appears references to the copied header array elements trigger segmentation faults. It also seems to be a race condition so in some cases it doesn't happen (when I tested it before merging it never occurred).

Fixes:
- Pass the header array as a string and create our own copy of QJsonObject instead.
- A nullptr guard was added to `tableofcontentbar.cpp` to prevent potential nullptr access. (This is irrelevant to the bug but something to carry over.)

### DIFF
--- a/resources/js/headerAnchor.js
+++ b/resources/js/headerAnchor.js
@@ -31,7 +31,7 @@ function anchorHeaderElements(headers)
     });
 }
 
-function getHeaders()
+function getHeadersJSONStr()
 {
     const headerInfo = { url: window.location.href.replace(location.hash,""), headers: [] };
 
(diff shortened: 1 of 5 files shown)


ShaopengLin:
@kelson42 @veloman-yunkan It would be great if we get this in ASAP. Though at least for me it doesn't always occur, it can be very frustrating for other devs.

kelson42:
@ShaopengLin Inwill merge and a new full testing round this WE.

ShaopengLin:
Just another piece of info, this issue doesn't occur in Qt6.

veloman-yunkan (approved):

PR #1256, opened 2024-11-23
Improved UI for Reading List Bar and Table of Contents

### LINKED ISSUE #1247: Make ReadingList UI Consistent
Now that #42 has merged, the reading list bar should match the aesthetics of the others. This would make the app feel segregated and incomplete.

We would at least need to have:
  1. Same background color
  2. Same item height/size
  3. Same hover and scroller visuals

| Table of Content Sidebar | Reading List Sidebar
| -------- | ------- |
|   Image |  Image |

heropj:
fix #1247 

Changes:

- Standardized UI for Reading List Bar and Table of Contents.

Screenshot 2024-11-23 195403
Screenshot 2024-11-23 195416


kelson42:
@heropj Thank you for your PR. I have rebased it on latest `main` branch and we will review it.

ShaopengLin (changes requested):
@heropj Please first remove the superfluous changes before we move to a formal review. The changes to some files like the icons may be due to your formatter? It is important to keep commit footprint at a minimum.

heropj:
@ShaopengLin Sorry for the inconvenience caused by the superfluous changes. In my latest commit I have ensured no such changes are present. I’ve addressed other changes also as per your suggestion.

ShaopengLin:
@ShaopengLin There are still 46 files changed. Please only leave those which are relevant.

heropj:
@ShaopengLin Got it..I have removed the commits which caused those unnecessary changes and have made a fresh commit with relevant changes only.

ShaopengLin (commented):
@heropj The issue wants you to match ReadingList to the behaviour and look of the TOC, and we should not change the TOC. There is a reason for the design of TOC to be like this so please do not change it and focus on the 3 points I mentioned in the Issue.

SInce there are only a few UI changes, I will first focus on the UI. Once that's up-to-par, I can review the coding part.

heropj:
@ShaopengLin Thank you for the feedback, I will undo changes in TOC and make RL ui similar to it.

heropj:
@ShaopengLin 
Screenshot 2024-12-01 123045
Screenshot 2024-12-01 123058

src/mainwindow.cpp
```diff
@@ -197,6 +197,7 @@ void MainWindow::tableOfContentToggled(bool state)
     if (state) {
         checkActionNoSignal(KiwixApp::ToggleReadingListAction, false);
         mp_ui->sideBar->setCurrentWidget(mp_ui->tableofcontentbar);
+        mp_ui->sideBar->setStyleSheet("background-color: white;");
```
ShaopengLin, line 200: This is a bug. If the state has not changed, the stylesheet won't change. Like image below.

You should be able to change the background color like TOC, try to investigate the CSS changes before doing ad-hoc changes like this.

Screenshot from 2024-12-02 14-44-11

src/readinglistbar.cpp
```diff
@@ -15,6 +15,16 @@ ReadingListBar::ReadingListBar(QWidget *parent) :
     ui(new Ui::readinglistbar)
 {
     ui->setupUi(this);
+
+    //to remove gap between RL components
+    if (layout()) {
```
ShaopengLin, line 20: Why is this check necessary? I don't see setLayout() calls later so this state won't change.

ShaopengLin (changes requested):

ShaopengLin:
@heropj The change LGTM, but do change the commit message to be more generalized and meaningful.

@veloman-yunkan @kelson42 The UI and code look ok to me. Please have a round of review if you have some further suggestions.

heropj:
@ShaopengLin hello, sorry for those silly ad-hoc changes...I revisited the whole code and found that readinglistbar was of type QWidget whereas TOC was inheriting from class QFrame....so i changed RLB to a QFrame and then applied same styling as of toc.
pleass review the changes now...i hope everything is right this time :)

ShaopengLin (approved):

src/mainwindow.cpp
```diff
@@ -197,7 +197,6 @@ void MainWindow::tableOfContentToggled(bool state)
     if (state) {
         checkActionNoSignal(KiwixApp::ToggleReadingListAction, false);
         mp_ui->sideBar->setCurrentWidget(mp_ui->tableofcontentbar);
-        mp_ui->sideBar->setStyleSheet("background-color: white;");
```
ShaopengLin: This change shouldn't be there in the first place. Do not let things like this appear in the commit history.

src/readinglistbar.cpp
```diff
@@ -11,17 +11,10 @@
...
 
 ReadingListBar::ReadingListBar(QWidget *parent) :
-    QWidget(parent),
+    QFrame(parent),
     ui(new Ui::readinglistbar)
 {
     ui->setupUi(this);
-
-    //to remove gap between RL components
-    if (layout()) {
```
ShaopengLin: Same here.

ShaopengLin (changes requested):
Please fix the following

ShaopengLin:
@heropj We want the commit message to be both concise and informative.
> final fix of RL

is not informative on the changes you did.

The second commit is way too long and thus not concise.

heropj:
@ShaopengLin done😢😅

ShaopengLin (approved):

kelson42:
@heropj @ShaopengLin Thank you very much. I will review this PR from a user perspective this week-end and I hope @veloman-yunkan can do the same from the code perspective.

src/readinglistbar.cpp
```diff
@@ -11,10 +11,13 @@
 const QString documentsDir = QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation);
 
 ReadingListBar::ReadingListBar(QWidget *parent) :
-    QWidget(parent),
+    QFrame(parent),
     ui(new Ui::readinglistbar)
 {
     ui->setupUi(this);
+    ui->titleLabel->setFont(QFont("Selawik", 18, QFont::Weight::Medium));
```
veloman-yunkan, line 18: Why is the font set in C++ code? Can't it be controlled via CSS?

src/readinglistbar.cpp
```diff
@@ -37,7 +40,6 @@ ReadingListBar::ReadingListBar(QWidget *parent) :
     auto importAction = app->getAction(KiwixApp::ImportReadingListAction);
     connect(exportAction, &QAction::triggered, this, &ReadingListBar::onExport);
     connect(importAction, &QAction::triggered, this, &ReadingListBar::onImport);
-    ui->label->setText(gt("reading-list-title"));
```
veloman-yunkan, line 40: So this was replaced with `ui->titleLabel->setText(gt("reading-list"));` (and moved to the beginning of the constructor body). The new name of the label data member makes more sense, however it's not clear why the i18n text string id was changed from `"reading-list-title"` to `"reading-list"`.

But checking `resources/i18n/qqq.json` and `resources/i18n/en.json` raises another question - do we really need both those strings? One of them is used as the tooltip for the reading list button and the other as the heading of the reading list bar. In English translation file the difference between them is only in capitalization, while their descriptions in `qqq.json` are identical. We should either drop one of them, or describe them differently in `qqq.json`.

@ShaopengLin @kelson42

heropj: Hello! thank you for reviewing.. the reason i changed it from `"reading-list-title"` to `"reading-list"` is to maintain consistency with code present in tableofcontentbar.cpp😅 and since it worked i didn't got into checking the difference it might cause...but now that you have investigated about these strings, let me know how we should proceed further...drop one of them or change their description.

src/readinglistbar.cpp
heropj: Yes it can be set via CSS but TOC also had font set in cpp code in tableofcontentbar.cpp... so to maintain code consistency i added this line here in readinglistbar.cpp.

ShaopengLin: If possible I am ok with doing this in CSS. Its just my personal preference to control font in Qt, as I remember some widgets doesn't support CSS fonts.

heropj: Hello.. I just checked it's working via CSS, so should i do it in CSS??

veloman-yunkan: Yes, please

src/readinglistbar.cpp
veloman-yunkan: @kelson42 Will you please provide your feedback to my comment above?

kelson42: @veloman-yunkan Nomstrong opinion, but if I has to choose I would say that having two strings seems wrong.

veloman-yunkan: @heropj Then let's get rid of the `"reading-list-title"` and use `"reading-list"` instead. But I think that its capitalization must be changed to title case.

heropj: @veloman-yunkan Do i need to remove `"reading-list-title"` from all translation files?? or just from `qqq.json` and `en.json`??

veloman-yunkan: Only `qqq.json` and `en.json` must be updated.

kelson42 (approved):
LGTM from a user perspective

veloman-yunkan (commented):
The final result looks good to me, but the commit history falls a little short of what I expect from a good code evolution history. The main problem is that the second commit combines two unrelated changes and rather reflects the review process than a logical and clean series of steps of arriving at the agreed state. Please rewrite the commit history of this PR as follows:

1. Dropping of the "reading-list-title" entry from translated strings (C++ code using the removed entry must be updated accordingly, each commit must be self-consistent).
2. Restyling of the Reading List sidebar (rest of the changes).

veloman-yunkan (approved):

PR #1258, opened 2024-11-25
Switch Linux CI to jammy

veloman-yunkan:
Fixed the Windows CI (broken because of kiwix/kiwix-build#779) and swtiched the Linux CI to jammy.

PR #1259, opened 2024-11-26
Under Windows settings are persisted in files too

### LINKED ISSUE #1254: Application settings are not persistent under Windows
Under Windows, for some reason `kiwix-desktop` is unable to persist its settings.

veloman-yunkan:
Fixes #1254 

Before this fix the settings and session information should have been written to the Windows registry however for some reason it didn't work.

In any case we want the settings to be saved in files so that it plays better with portable mode (when we start supporting it for Windows builds too).

Explicit settings (those configurable through the settings dialog, and library filters) are saved in
**C:\\Users\\**_USER_**\\AppData\\Local\\kiwix-desktop\\Kiwix-desktop.conf**

Session info (window size, list of open tabs, last directory used for opening a ZIM file) are saved in
**C:\\Users\\**_USER_**\\AppData\\Roaming\\kiwix-desktop\\kiwix-desktop.session**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
       run: |
         set PKG_CONFIG_PATH=%cd%\BUILD_win-amd64\INSTALL\lib\pkgconfig
         qmake PREFIX=%cd%\BUILD_win-amd64\INSTALL
-        nmake debug-all
+        nmake release-all
 
   Linux:
     strategy:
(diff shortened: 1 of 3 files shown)


veloman-yunkan:
> Failure of the CI is due to the issue fixed in #1258 (BTW, should we split that PR into two?)

Well I moved the commit fixing the Windows CI from #1258 here because here it makes more sense.

PR #1288, opened 2024-12-17
Workaround for strange libkiwix-dev dependency issue

### LINKED ISSUE #1285: Upload of 2.4.0 PPA fails
... although it works fine with dev!

https://github.com/kiwix/kiwix-desktop/actions/runs/12331074751

veloman-yunkan:
Should fix #1285

For some mysterious reason the dependency requirement `libkiwix-dev (>= 14.0.0)` in the Debian control file results in PPA build failure. After unpacking `kiwix-build-deps_2.4.0~focal_all.deb` the `apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes install -f` commands reports:

    Reading package lists...
    Building dependency tree...
    Reading state information...
    Starting pkgProblemResolver with broken count: 1
    Starting 2 pkgProblemResolver with broken count: 1
    Investigating (0) kiwix-build-deps:amd64 < 2.4.0~focal @iU mK Nb Ib >
    Broken kiwix-build-deps:amd64 Depends on libkiwix-dev:amd64 < none -> 14.0.0~focal @un uN > (>= 14.0.0)
      Removing kiwix-build-deps:amd64 because I can't find libkiwix-dev:amd64
    Done

Shortening the libkiwix-dev minimal version string to "14.0" (`libkiwix-dev (>= 14.0)`) works around that issue.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,8 +7,8 @@ Build-Depends: debhelper-compat (= 13),
  qtbase5-dev,
  qtwebengine5-dev,
  libqt5texttospeech5-dev,
- libkiwix-dev (>= 14.0.0), libkiwix-dev (<< 15.0.0),
- libzim-dev (>= 9.0.0), libzim-dev (<< 10.0.0),
+ libkiwix-dev (>= 14.0), libkiwix-dev (<< 15.0),
+ libzim-dev (>= 9.0), libzim-dev (<< 10.0),
 Standards-Version: 4.5.0
 Homepage: https://www.kiwix.org/
 Rules-Requires-Root: no


veloman-yunkan:
After this PR is merged the 2.4.0 release tag has to be moved.

kelson42 (changes requested):
@veloman-yunkan Thx for this strange finding and fix. To me -whatever we do, we should be coherent how we represent the upper version bound and the lower one. So I would prefer to have "15.0".

veloman-yunkan:
The lower and upper bound versions of the libzim dependency have been shortened too for consistency.

kelson42:
@veloman-yunkan Thx. On my side I would be ready to merge, but lets try to understand first why it was working on dev.

kelson42:
OK, then I merge because we need to have `2.4` in the PPA to complete the release process.

PR #1295, opened 2024-12-20
Fixes for Qt 6.4.3 on Windows

adamlamar:
kiwix-build was unable to build on Windows with Qt 6.4.3. The fix makes use of the `gui-private` module (aka platform-specific Qt code) on windows instead of the direct win32 calls.

Tested out using artifacts generated from kiwix-build and the application appears to work normally.

### DIFF
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -9,6 +9,10 @@ QT       += webenginewidgets webchannel
 QT       += printsupport
 qtHaveModule(texttospeech): QT += texttospeech
 
+win32 {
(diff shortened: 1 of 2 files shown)


kelson42:
@adamlamar I'm not familiar with this module.... but why not considering using it all the time?

adamlamar:
Mostly because it introduces platform-specific code. Some problems for example might be that you could use classes in Ubuntu that don't exist on Windows and vice-versa.

veloman-yunkan (approved):

Issue #1243, opened 2024-11-13
Rename to Kiwix


rgaudin:
At this stage, that could be classified as a bug I think 😀

According to the [branding document](https://github.com/kiwix/overview/wiki/Branding), `kiwix-desktop` should be exposed as `Kiwix`.
End user should never see `kiwix-desktop` except in a link to the repository.

That means renaming the App so the Title bar says `Kiwix`, renaming the binaries (`Kiwix[.exe]`) and I believe the Archive names but I'm not clear on what it should actually be: `Kiwix Windows_x64_2.4.0.zip` ?? (see https://github.com/kiwix/overview/issues/106 as well)

Side question for @kelson42 and @Popolechien. Are you OK with having both `Kiwix Windows` (ie `Kiwix` + Platform) and `Kiwix Hotspot` (ie. a Kiwix product) ?

Popolechien:
Yes to the first question (Kiwix-desktop -> Kiwix), but I am not sure I understand the other tbh. Shouldn't it be `Kiwix for Windows` ?

rgaudin:
> Yes to the first question (Kiwix-desktop -> Kiwix)

That was not a question 😉

> but I am not sure I understand the other tbh. Shouldn't it be `Kiwix for Windows` ?

That's the question. The branding doc says that all readers are `Kiwix` from their own perspective, and there is no debate AFAIK.
To the outside world though, we need to distinguish the different products and the doc says we should have

- Kiwix Windows
- Kiwix Linux
- Kiwix Android
- Kiwix Hotspot

This is only marginally relevant to this ticket as it would only apply to the _packaging_ (the archive) to download kiwix-desktop and that happens on kiwix-build but as I wrote, we might want to consider updating the branding doc as I find it odd to have both `Kiwix {Platform}` and `Kiwix Hotspot` using the same pattern. Your suggestion makes sense (`Kiwix for {Platform}`).

kelson42:
@rgaudin yes, yes yes ;)

Popolechien:
Looks like we're all in agreement then:

- `Kiwix-desktop` displayed as `Kiwix`
- Tweak product list with `for` (e.g. `Kiwix for Windows`, except the `Kiwix Hotspot`, which is a different product approach)
- Repositories remain as-is.

rgaudin:
OK then.

PR #1302, opened 2024-12-24
Rename binary from `kiwix-desktop` to `kiwix`

kelson42:
Part of #1243

### DIFF
--- a/kiwix-desktop.pro
+++ b/kiwix-desktop.pro
@@ -20,7 +20,7 @@ CONFIG += link_pkgconfig
 
 greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
 
-TARGET = kiwix-desktop
+TARGET = kiwix
 TEMPLATE = app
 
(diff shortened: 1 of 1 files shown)


resources/i18n/en.json
```diff
@@ -6,7 +6,7 @@
     "name":"English",
     "error-title":"Error",
     "error-downloader-window-title":"Cannot create downloader",
-    "error-downloader-launch-message":"Impossible to launch downloader, Kiwix-desktop will start but all download functions will not working!",
+    "error-downloader-launch-message":"Impossible to launch downloader, Kiwix will start but all download functions will not working!",
```
veloman-yunkan, line 9: Please also fix the grammar in this message. It should rather read

"Impossible to launch the downloader, Kiwix will start but download functionality will not work!"

veloman-yunkan (commented):
This will also require respective changes in kiwix-build. I think that a draft version of that PR should be prepared before we merge this one.

PR #1307, opened 2024-12-26
Default qt6

### LINKED ISSUE #1300: Requires Qt6
Secure that Qt6 is required to compile. If yomeone still want to compile with Qt5 it will have tomset proper env variable. README should be also updated.

kelson42:
Fixes #1300


kiwix-desktop.pro
```diff
@@ -20,12 +20,22 @@ CONFIG += link_pkgconfig
...
 TEMPLATE = app
 
 QMAKE_CXXFLAGS += -std=c++17
 QMAKE_LFLAGS +=  -std=c++17
 
+# Kiwix requires per default on Qt6 (but can still compile with Qt5 if explicitly
+# requested, see README.md for more information).
+QT_SELECT += $$(QT_SELECT)
+lessThan(QT_MAJOR_VERSION, 6) {
+    !equals(QT_SELECT, "qt5") {
```
adamlamar, line 33: Do you want a kiwix-specific env var to control the build behavior? The `QT_SELECT` env var is not the only way to select which Qt installation to use, so someone might be intentionally using a qt5 toolchain yet the build would fail because the `QT_SELECT` env var was not set. Similarly, if someone already set `QT_SELECT=qt5` as part of their default shell environment, the build _wouldn't_ fail even though maybe it should have.

A kiwix-specific var like `KIWIX_QT=qt5` would prevent this.

kelson42: I have no strong opinion on this point... I mostly focus on simplicity. Just to be sure I understand your scenarios:

> someone might be intentionally using a qt5 toolchain yet the build would fail because the QT_SELECT env var was not set.

OK, but then he clearly ignores the `README` right? If he ignores the `README` he won't find out how to set `KIWIX_QT=qt5` either.

> Similarly, if someone already set QT_SELECT=qt5 as part of their default shell environment, the build wouldn't fail even though maybe it should have.

You mean, he would have set `QT_SELECT=qt5` but actually compiling with Qt6 for example? OK... but that sounds pretty weird!

kelson42:
@adamlamar Somehow I keep failing to get our PPA packages using Qt6... maybe you have an idea?!

kiwix-desktop.pro
adamlamar: > OK, but then he clearly ignores the README

I typically assume folks ignore the README, at least to some degree. They will usually follow any initial instructions but I would not expect beginners to dig into the details of the README. Even if they read it, they don't often know the significance of what they read, or what context it applies. Just saying there are limits to what beginners will pick up on.

With `KIWIX_QT=qt5`, my understanding was that you wanted a default Qt6 experience, where everyone compiles with qt6 by default, and only drops back to qt5 for certain cases like compiling on older distros. If that's what you want, using a kiwix-specific env var would accomplish that.

Regarding my original comment - `someone might be intentionally using a qt5 toolchain` - what I mean is that they could use a different avenue of configuring `qtchooser`, like via a config file. They could have intentionally setup their environment to build with Qt5, but the build would fail because they configured `qtchooser` via config instead of `QT_SELECT`. This is not a huge deal, but could trip up someone who doesn't understand the intent of the environment variables. This is why I think `KIWIX_QT` is more clear all around. If we use `QT_SELECT` to determine whether a Qt5 build is allowed, we're conflating the configuration of `qtchooser` ("which version of Qt does the system provide") with the requirements of kiwix ("which version of Qt does the project require").

> he would have set QT_SELECT=qt5 but actually compiling with Qt6

No, I mean that a Qt developer with a preexisting default of `QT_SELECT=qt5` in their environment could walk into the project and accidentally compile with Qt5.  If I understand what you want - compile with Qt6 by default except some limited circumstances - this scenario seems to break your requirement. Also not a huge deal, but I bring it up because it seems contrary to the desired experience.

adamlamar:
@kelson42 I see the qt6 packages being installed but there is an error in the CI output:

```
qmake: could not find a Qt installation of 'qt6'
```

IIRC this means you'll need to run:

```
qtchooser -install qt6 $(which qmake6)
```

prior to attempting to invoke qmake.

kelson42:
Post `2.5.0`, we will remove Qt5 support and this will simplify significantly the job and the compilation infrastructure. Closing for the moment, but still available as documentation for further work.

Issue #1267, opened 2024-12-05
Not able to read ZIM when launch from Parrallels Desktop Profile Shared Folder


rgaudin:
When launched from a ~~network drive~~ Parrallels Desktop Profile Shared Folder (aka a shared folder which to Apps appear to be a regular `C:\Mac` prefixed folder), clicking Open on a ZIM opens a new tab with the correct Icon but that tab is blank.
Very first time I clicked `+` for a new tab I got an error message which makes me doubt #1236 is completely fixed.

 ![Image](https://github.com/user-attachments/assets/40925475-bf7b-44f8-9e52-66ece5bbbd25)

I dont want to open a general ticket but I think the person fixing this should test the overall behavior of the app after fixing it…

kelson42:
Indeed, exactly the same as #885

veloman-yunkan:
Could not reproduce with the [2024-12-06 nightly build](https://download.kiwix.org/nightly/2024-12-06/kiwix-desktop_windows_x64_2024-12-06.zip) extracted into a samba shared folder in an Ubuntu 22.04 virtual machine.

rgaudin:
It's the same with this build on my Windows 11 VM

veloman-yunkan:
Is it because of the Windows version? Mine is Windows 10 (Pro).

veloman-yunkan:
BTW, is the confirmation box about running `kiwix-desktop` from a remote disk/network folder displayed in your case?

rgaudin:
Never seen that dialog box.

veloman-yunkan:
Then the problem is that the path of your `kiwix-desktop` is not recognized as a remote location/network path. What does it look like?

rgaudin:
Indeed the VM engine I'm using implemented a trick to prevent those apparently common shared-folder issue… creating this one.

See https://kb.parallels.com/130138

I'm not sure there's much we can do.

veloman-yunkan:
I think that the title and description of this ticket must be updated so that the problem reported doesn't appear much more general than it actually is. Also I believe that at this stage we shouldn't spend any effort fixing issues caused by interoperability between different components of heterogeneous systems (at least when, like int this case, there is a workaround - don't run `kiwix-desktop` from a shared folder, copy it to your local machine instead).

kelson42:
I don't fully get the problem, but postponing.

rgaudin:
Edited. I'll report it to Parallels as well.

[...]

PR #1313, opened 2024-12-31
Parrallels shared folder support

kelson42:
Fixes https://github.com/kiwix/kiwix-desktop/issues/1267

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -17,9 +17,17 @@
 bool wasAppStartedFromARemoteDrive()
 {
     const std::string exePath = kiwix::getExecutablePath();
+
+    // Usual mounting point for Windows Network Drives
     if ( exePath.substr(0, 2) == "\\\\" )
         return true;
 
+    // Support macOS Parallels shared folders.
+    // See https://kb.parallels.com/130138
+    if ( exePath.substr(0, 7) == "C:\\Mac\\" )
+        return true;
+
+    // Last chance to identify as remote drive
     return GetDriveTypeA(exePath.substr(0, 3).c_str()) == DRIVE_REMOTE;
 }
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -19,16 +19,16 @@ bool wasAppStartedFromARemoteDrive()
...
     // Support macOS Parallels shared folders.
     // See https://kb.parallels.com/130138
-    if ( exePath.substr(0, 7) == "C:\\Mac\\" )
+    if ( exePath.length() >= 7 && exePath.substr(0, 7) == "C:\\Mac\\" )
         return true;
 
     // Last chance to identify as remote drive
-    return GetDriveTypeA(exePath.substr(0, 3).c_str()) == DRIVE_REMOTE;
+    return exePath.length() >= 3 && GetDriveTypeA(exePath.substr(0, 3).c_str()) == DRIVE_REMOTE;
 }
```
veloman-yunkan, line 32: This entire commit is not needed. [`std::string::substr()`](https://en.cppreference.com/w/cpp/string/basic_string/substr) is safe on the overflow caused by its second parameter - it then simply returns the substring until the end of the string.

> Returns a substring `[pos, pos + count)`. If the requested substring extends past the end of the string, i.e. the count is greater than `size() - pos` (e.g. if `count == npos`), the returned substring is `[pos, size())`.

veloman-yunkan (changes requested):

src/main.cpp
kelson42: @veloman-yunkan I will remove it and merge if OK for you?

veloman-yunkan: @kelson42 That's OK

PR #1315, opened 2024-12-31
fix: Update filters after sideloading a ZIM file

### LINKED ISSUE #1255: Side loading a ZIM file manually does not update the local category filter
I have sideloaded a Wikipedia, but the "Category" filter of the local library still does not show "Wikipedia". I have to go to the online library and back to the local library to see this appearing the filter.

heropj:
fix #1255 

Changes:

- Added calls to `setCategories()` and `setLanguages()` to be triggered whenever the `booksChanged` signal is emitted.

Now, whenever `booksChanged` event is called, eg. when a file is deleted or sideloaded.., the filter list will get updated.


kelson42:
@heropj Thank you for your PR, I will check if it works later today.

kelson42 (approved):
LGTM!

veloman-yunkan (approved):

PR #1317, opened 2025-01-01
Feat: Added speed control option for Text-to-Speech module

### LINKED ISSUE #1299: Add option to speed-up Text To Speech
because its painfully slow. wheres the bar to set speed like 1.5x 2x 3x

heropj:
fix #1299 

- Made a function to change rate of speech based on user's choice.
- User can select speed of their choice from the ui.
- User can also use keyboard shortcut `>`/`<` to increase/decrease speed.
- When a user changes speed, the speech is stopped and 'restarted' with new speed.

image

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -185,5 +185,8 @@
     "select-read-voice": "Select reading voice",
     "select-read-language": "Select reading language",
     "save-or-open": "Save or Open file",
-    "save-or-open-text": "What should Kiwix do with this file?"
+    "save-or-open-text": "What should Kiwix do with this file?",
+    "speed": "Speed",
+    "increase-tts-speed": "Increase TTS speed.",
+    "decrease-tts-speed": "Decrease TTS speed."
 }
--- a/src/texttospeechbar.cpp
+++ b/src/texttospeechbar.cpp
@@ -21,6 +21,7 @@ TextToSpeechBar::TextToSpeechBar(QWidget *parent)
     connect(ui->closeButton, &QPushButton::pressed,
                 this, &TextToSpeechBar::speechClose);
 
+    setupSpeedOptionsComboBox();
     setupVoiceComboBox();
     setupLanguageComboBox();
     languageSelected(ui->langComboBox->currentIndex());
@@ -28,6 +29,28 @@ TextToSpeechBar::TextToSpeechBar(QWidget *parent)
             this, &TextToSpeechBar::toggleLanguage);
     connect(app->getAction(KiwixApp::ToggleTTSVoiceAction), &QAction::triggered,
             this, &TextToSpeechBar::toggleVoice);
+    connect(app->getAction(KiwixApp::IncreaseTTSSpeedAction), &QAction::triggered,
+            this, &TextToSpeechBar::increaseSpeed);
+    connect(app->getAction(KiwixApp::DecreaseTTSSpeedAction), &QAction::triggered,
+            this, &TextToSpeechBar::decreaseSpeed);
+    connect(ui->speedComboBox, QOverload<int>::of(&QComboBox::currentIndexChanged),
+            this, &TextToSpeechBar::onSpeedChanged);
+}
+
+void TextToSpeechBar::increaseSpeed()
+{
+    int currentIndex = ui->speedComboBox->currentIndex();
+    if (currentIndex < ui->speedComboBox->count() - 1) {
+        ui->speedComboBox->setCurrentIndex(currentIndex + 1);
+    }
+}
+
+void TextToSpeechBar::decreaseSpeed()
+{
+    int currentIndex = ui->speedComboBox->currentIndex();
+    if (currentIndex > 0) {
+        ui->speedComboBox->setCurrentIndex(currentIndex - 1);
+    }
 }
 
 void TextToSpeechBar::speak(const QString &text)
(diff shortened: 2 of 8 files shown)


kelson42 (changes requested):
@heropj Thank you again very much for this PR. To me it works fine, but could you please save in the preferences (beside the chosen voice) the speed, so the user does not have always to change back the TTS speed to his prefered speed?

src/texttospeechbar.cpp
```diff
@@ -54,6 +58,16 @@ void TextToSpeechBar::setLocale(const QLocale& locale)
     }
 }
 
+void TextToSpeechBar::setupSpeedOptionsComboBox()
+{
+    ui->speedLabel->setText(gt("speed"));
```
veloman-yunkan, line 63: I think you forgot to include the change to `resources/i18n/en.json`. Please also don't forget to update `resources/i18n/qqq.json` too.

src/texttospeechbar.cpp
```diff
@@ -54,6 +58,16 @@ void TextToSpeechBar::setLocale(const QLocale& locale)
...
+    ui->speedComboBox->setMaxVisibleItems(10);
+    ui->speedComboBox->setLineEdit(new ComboBoxLineEdit(ui->speedComboBox));
+
+    QStringList speedOptions = {"0.25x","0.50x","0.75x","1.00x","1.25x","1.50x","1.75x","2.00x"};
```
veloman-yunkan, line 67: Should we include x in the speed rate value? Won't internationalization be slightly affected? In youtube there is no x.

heropj: 👍👍i'll remove it.

heropj:
hello @kelson42 , @veloman-yunkan 
There are shortcuts to select language and voice...
eg `CREATE_ACTION_SHORTCUT(ToggleTTSLanguageAction, gt("select-read-language"), QKeySequence(Qt::ALT | Qt::SHIFT | Qt::Key_L));` to select language

do we need a shortcut for selecting voice also?? if yes then what should be the key-combination..
`ALT+SHIFT+S` is already selected as shortcut for "reading selected"

kelson42:
@heropj Ready for a new review?

heropj:
@kelson42 yes👍👍

veloman-yunkan (changes requested):

src/texttospeechbar.cpp
```diff
@@ -21,13 +21,44 @@ TextToSpeechBar::TextToSpeechBar(QWidget *parent)
...
     connect(app->getAction(KiwixApp::ToggleTTSLanguageAction), &QAction::triggered,
             this, &TextToSpeechBar::toggleLanguage);
     connect(app->getAction(KiwixApp::ToggleTTSVoiceAction), &QAction::triggered,
             this, &TextToSpeechBar::toggleVoice);
+    connect(ui->speedComboBox, QOverload<int>::of(&QComboBox::currentIndexChanged),
+            this, &TextToSpeechBar::onSpeedChanged);
+
+//keyboard shortcuts to control speed of tts
+QAction* increaseSpeedAction = new QAction(this);
+increaseSpeedAction->setShortcut(QKeySequence("Shift+."));
```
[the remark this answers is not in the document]
heropj: shift+. is clearer for users?? it's clear that they have to press shift and .
while for '>' they might get confused..??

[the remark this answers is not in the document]
veloman-yunkan: The question is whether you want the shortcut to be bound to the `"Shift + ."` key sequence (which is equivalent to `">"` on my keyboard but may resolve to something else on different keyboard layouts)¹?  Or the underlying reason for selecting this key combination is mnemonics - increase with ">" and decrease with "<"?

---
¹ BTW, what if there is a keyboard layout where `"."` itself has to be entered via a `Shift` key?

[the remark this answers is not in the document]
heropj: okay..i got it.., we should go with mnemonics... i'll change it to '>' for speed up and '<' for speed down.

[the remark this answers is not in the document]
heropj: done..
image

resources/i18n/en.json
```diff
@@ -184,5 +184,8 @@
     "select-read-voice": "Select reading voice",
     "select-read-language": "Select reading language",
     "save-or-open": "Save or Open file",
-    "save-or-open-text": "What should Kiwix do with this file?"
+    "save-or-open-text": "What should Kiwix do with this file?",
+    "speed": "Speed",
+	"increase-tts-speed": "Represents the action of increasing the speed of the text-to-speech.",
+	"decrease-tts-speed": "Represents the action of decreasing the speed of the text-to-speech."
```
veloman-yunkan, line 190: 1. Wrong text (copied from `qqq.json`). Replace with text that should appear in the UI.
2. Fix indentation (eliminate the tab-vs-spaces discrepancy)

veloman-yunkan (changes requested):

resources/i18n/en.json
veloman-yunkan: Indentation was not fixed (see the diff in github)

veloman-yunkan: This file uses tabs instead of spaces whereas your editor likely inserts spaces.

heropj: fixed

veloman-yunkan (approved):

resources/i18n/en.json
veloman-yunkan: > This file uses tabs instead of spaces whereas your editor likely inserts spaces.

Actually, it was the other way around (I absent-mindedly checked in `libkiwix` instead of `kiwix-desktop`), but you fixed it correctly. Thanks!

PR #1348, opened 2025-02-27
Fix: Record TOC navigations in history (#1248)

### LINKED ISSUE #1248: TOC navigations should be recorded in history
Currently, the TOC navigations do not go through the Qt system and use Javascript to scroll directly. It would be great to have an **in-page** history, where the history records those actions but sends them to Javascript for scrolling instead of refreshing the current page to the URL.

Note that if a page is refreshed, all DOM states are also reset, which is not what we want. This is the reason why we cannot use setUr to navigate to TOC entries.

Jinash-Rouniyar:
Closes #1248

Currently, TOC navigations do not go through the Qt system and use JavaScript to scroll directly. 
This PR ensures that in-page history records those actions while still allowing JavaScript to handle scrolling. 
This prevents unnecessary page refreshes that reset DOM states.

@kiwix/lieutenants Please review when available.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -185,5 +185,8 @@
     "select-read-voice": "Select reading voice",
     "select-read-language": "Select reading language",
     "save-or-open": "Save or Open file",
-    "save-or-open-text": "What should Kiwix do with this file?"
+    "save-or-open-text": "What should Kiwix do with this file?",
+    "speed": "Speed",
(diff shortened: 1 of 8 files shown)


kelson42:
@Jinash-Rouniyar Thank you for your PR. Please make commits only related and necessary to your change. I won't review 30.000 lines of code changes.

PR #1373, opened 2025-05-02
Replace deprecated /catalog/search catalog end-point call

### LINKED ISSUE #1372: Stop using `/catalog/search` deprecated API
This API seems to be used according to:
https://github.com/kiwix/operations/issues/409#issuecomment-2846712635

But according to the doc this is deprecated and should not be used anymore, see:
https://kiwix-tools.readthedocs.io/en/latest/kiwix-serve.html#catalog-search

kelson42:
Stop using `/catalog/search` deprecated end-point and use instead `/catalog/v2/entries` instead.

Fixes #1372

### DIFF
--- a/src/opdsrequestmanager.cpp
+++ b/src/opdsrequestmanager.cpp
@@ -41,7 +41,7 @@ void OpdsRequestManager::doUpdate(const QString& currentLanguage, const QString&
         query.addQueryItem("category", categoryFilter);
     }
 
-    auto mp_reply = opdsResponseFromPath("/catalog/search", query);
+    auto mp_reply = opdsResponseFromPath("/catalog/v2/entries", query);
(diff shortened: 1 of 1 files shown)


veloman-yunkan (commented):
There are some slight differences between the output of `/catalog/search` and `/catalog/v2/entries` endpoints. Did you test that `kiwix-desktop`'s functionality isn't affected by this change?

kelson42:
@veloman-yunkan I made a quick overview ; but not knowing the details of the changes, I'm not in oosition to do more. Can you please check yourself?

veloman-yunkan:
The differences seem insignificant. If it works for some set of filters it should work in general too.

veloman-yunkan (approved):

PR #1387, opened 2025-07-07
Localisation updates from https://translatewiki.net.

translatewiki:
Translation updates


resources/i18n/es-formal.json
```diff
@@ -0,0 +1,133 @@
+{
+	"@metadata": {
+		"authors": [
+			"Camachevaz"
+		]
+	},
+	"name": "Inglés",
```
veloman-yunkan, line 7: This should be the name of the language in itself. I guess this one should be "Español (formal)" or something like that.

kelson42: Strange... this is not one of our steings right?

kelson42: We can not fix this like this. A translation is a translation. The language name should come from the code and shoukd not be trnslated. We always display a language in its own language.

Where is this string used? We need to fox this in a dedicated issue... and it sounds like a regression.

veloman-yunkan: In the previous comment, my links point to `libkiwix` instead of `kiwix-desktop` but the same argumentation applies to `kiwix-desktop`.

veloman-yunkan: > > Where is this string used?
> 
> The string is used in the UI language selector.

It looks like in `kiwix-desktop` there is no UI to select the UI language (at least I couldn't find it). Then this string isn't used anywhere.

Issue #1406, opened 2025-09-22
qfutureinterface.h:455:37: error: template-id not allowed for constructor in C++20


kelson42:
```
/usr/lib/qt6/libexec/uic src/choiceitem.ui -o ui_choiceitem.h
/usr/lib/qt6/libexec/uic src/contentmanagerview.ui -o ui_contentmanagerview.h
/usr/lib/qt6/libexec/uic src/findinpagebar.ui -o ui_findinpagebar.h
/usr/lib/qt6/libexec/uic ui/kiwixchoicebox.ui -o ui_kiwixchoicebox.h
/usr/lib/qt6/libexec/uic ui/kiwixmessagebox.ui -o ui_kiwixmessagebox.h
/usr/lib/qt6/libexec/uic ui/mainwindow.ui -o ui_mainwindow.h
/usr/lib/qt6/libexec/uic ui/about.ui -o ui_about.h
/usr/lib/qt6/libexec/uic src/contentmanagerside.ui -o ui_contentmanagerside.h
/usr/lib/qt6/libexec/uic src/readinglistbar.ui -o ui_readinglistbar.h
/usr/lib/qt6/libexec/uic ui/localkiwixserver.ui -o ui_localkiwixserver.h
/usr/lib/qt6/libexec/uic ui/settings.ui -o ui_settings.h
ui/settings.ui: Warning: The name 'verticalLayout' (QVBoxLayout) is already in use, defaulting to 'verticalLayout1'.
/usr/lib/qt6/libexec/uic src/tableofcontentbar.ui -o ui_tableofcontentbar.h
/usr/lib/qt6/libexec/uic src/texttospeechbar.ui -o ui_texttospeechbar.h
[...]

PR #1407, opened 2025-09-22
Fix compilation error with buggy Qt versions

kelson42:
README update to fix https://github.com/kiwix/kiwix-desktop/issues/1406

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,25 @@ qmake .
 make
 ```
 
+Handling of 'template-id' related compilation errors
+----------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


kelson42:
This is the Qt of Ubuntu 24.04, so it has to work out of the box. We don't face here an "exotic" Linux  Distribution.

kelson42:
The compilation fix has proven to be more complex than I though initialy. The support of the `-Wtemplate-id-cdtor` is not implemented in all versions of the compiler. The combination of this with different versions of the Qt in the CI creates a matrix of cases complex to handle to me. Therefore I come back to your early suggestion and have upgraded the `README` with a new section.

Issue #1253, opened 2024-11-22
In a Windows debug build search suggestion popup triggers a Qt assertion


veloman-yunkan:
![Image](https://github.com/user-attachments/assets/784c07e9-ce0e-40cf-a993-7807e2c69d27)

The assertion is caused by `QStyledItemDelegate::paint()` being called with an invalid `QModelIndex`:

https://github.com/kiwix/kiwix-desktop/blob/63b421e6d338ee76ceb385ad16326e998174709d/src/suggestionlistdelegate.cpp#L9-L19

Under Linux this doesn't happen because in a normal development flow the debug build of `kiwix-desktop` uses the same shared libraries of Qt as the release build, whereas under Windows the version of the dlls used depends on the type of the build. Probably the issue can be reproduced under Linux by linking to the debug versions of QT .so files.

veloman-yunkan:
Note that this bug hides behind two other bugs fixed by #1252 which is not yet merged as of opening this ticket.

PR #1414, opened 2025-10-13
Fix Qt assertion error in SuggestionListDelegate::paint()

VikramAditya33:
Fixes issue #1253: Windows debug build assertion 'index.isValid()' when search suggestion popup is triggered.

The issue was caused by calling QStyledItemDelegate::paint() with an invalid QModelIndex(), which triggers Qt assertions in debug builds.

Solution:
- Clear text and icon from QStyleOptionViewItem to prevent default painting
- Call parent paint method with valid index instead of invalid QModelIndex()
- Maintain all existing visual behavior and functionality

This follows the same pattern used by ContentManagerDelegate in the codebase.

Files changed:
- src/suggestionlistdelegate.cpp (4 lines modified)

### DIFF
--- a/src/suggestionlistdelegate.cpp
+++ b/src/suggestionlistdelegate.cpp
@@ -10,10 +10,18 @@ void SuggestionListDelegate::paint(QPainter *painter,
                                    const QStyleOptionViewItem &option,
                                    const QModelIndex &index) const
 {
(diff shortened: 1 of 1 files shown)


kelson42:
@VikramAditya33 Thank you for your PR. Give us a few days for the review.

PR #1418, opened 2025-10-31
Remove deprecated language name

kelson42:
Should allow us to finally merge https://github.com/kiwix/kiwix-desktop/pull/1387


kelson42:
The reason why IMO it shoudl be dropped:
* It is not used
* It is not the right place to do that (here we talk about translations, not creating a list of language)
* I believe we should not have such a list of language here, as it should be in a central place (so libkiwix or maybe even in libicu... but this is not the discussion here).

To conclude, we could need language names for the UI (not used here as we don't have such setting in Kiwix Desktop) or for the library (it comes from libkiwix).

So I really not see why we shoukd keep that.

veloman-yunkan:
> * It is not the right place to do that (here we talk about translations, not creating a list of language)

We are not creating a list of languages. We just need to know the name of the language of the translation, and the translation file is the best place for it. Translators are the best source of that information. kiwix-desktop and libkiwix are unrelated projects in that regards. If some user of kiwix-desktop decides to add a new language, it is easier if they do it entirely in one place instead of providing translations via translatewiki and filing a request to register the language name in libkiwix.

PR #1422, opened 2025-11-16
Introduce "copy link" in context menu

Bad-ptr:
For internet URLs


kelson42:
@Bad-ptr Thank you very much for you PR, we will review it within the next days.

PR #1423, opened 2025-11-16
QRadioButton -> QCheckBox in MultiZimButton

Bad-ptr:
Blocked by https://github.com/openzim/libzim/issues/932

Allow to search in multiple books
Search in all books if none selected

Fixes the https://github.com/kiwix/kiwix-desktop/blob/5c3392b0aee9c6983800c26221caf4face29e933/src/suggestionlistworker.cpp#L20 ?

### DIFF
--- a/src/multizimbutton.cpp
+++ b/src/multizimbutton.cpp
@@ -1,8 +1,7 @@
 #include <QListWidget>
 #include <QMenu>
 #include <QWidgetAction>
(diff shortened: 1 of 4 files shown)


kelson42:
Multizim search doesnot work fine, see issues in `openzim/libzim`. This is the reason why this is done like this. Putting the PR in draft.

PR #1388, opened 2025-07-10
- Handle 301/302 redirects automatically in OpdsRequestManager

depuc:
- Follow redirects up to 5 times to prevent infinite loops
- Add debug logging for redirect chains
- Fixes #1374

PR #1429, opened 2025-12-06
Support HTTP redirections for OPDS feed

### LINKED ISSUE #1374: Online library requests should be able to deal with HTTP (301 and 302) redirects
Reasons is that:
* We move official API end-point hostname from `library.kiwix.org` to `opds.library.kiwix.org`, see https://github.com/kiwix/operations/issues/414#issuecomment-2847747481
* We want for the future to have a bit of flexibility and allow to redirect from one domain to another one.

depuc8:
Fixes #1374

- Switch OPDS catalog to opds.library.kiwix.org
- Enable HTTP 301/302 redirects for OPDS requests

Superseeds #1388

### DIFF
--- a/src/opdsrequestmanager.cpp
+++ b/src/opdsrequestmanager.cpp
@@ -10,7 +10,7 @@ QString OpdsRequestManager::getCatalogHost()
     const char* const envVarVal = getenv("KIWIX_CATALOG_HOST");
     return envVarVal
          ? envVarVal
-         : "library.kiwix.org";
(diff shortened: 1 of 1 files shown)


depuc8:
@veloman-yunkan I’ve addressed the requested changes and split the work into two commits as suggested.  
Review it when you have time, thanks!

depuc8:
thank you @kelson42  and @veloman-yunkan

PR #1430, opened 2025-12-08
Suggestion rendering fix

### LINKED ISSUE #1428: Bad rendering of suggestions by a local build of kiwix-desktop on Ubuntu 22.04
In my local build of kiwix-desktop on Ubuntu 22.04 suggestions are rendered twice with some offset:

Image

\
\
\
No such problem is present when using official release appimage kiwix-desktop 2.4.1:

Image

veloman-yunkan:
Fixes #1428

PR #1414 had resulted in suggestions being rendered twice

1. Default painting that was intended to be overriden by custom rendering introduced by #1189 (this is needed to automatically highlight the selected item)
    
2. followed by custom painting with some offset
    
PR #1414 had created a false impression that the default rendering of the icon and text was suppressed by setting those fields of `QStyleOptionViewItem` to empty values, whereas the text and icon in fact were obtained from the `QModelIndex`. The PR should have been tested as part of the review!
    
Now the bug is worked around by moving all the displayable data of `SuggestionListModel` entries to `UserRole`, while returning empty values for `DisplayRole` and `DecorationRole` data.

I haven't verified that the new solution preserves the intended effect of #1414 (elimination of the failure of an assertion when using a debug build of Qt) since I don't have easy access to a debug variant of Qt libs, however I don't see why that problem might return with my fix.

### DIFF
--- a/src/suggestionlistdelegate.cpp
+++ b/src/suggestionlistdelegate.cpp
@@ -13,21 +13,22 @@ void SuggestionListDelegate::paint(QPainter *painter,
     // Call parent paint with valid index to get proper background/selection
     QStyledItemDelegate::paint(painter, opt, index);
 
+    const auto userData = SuggestionListModel::getUserData(index);
     // Now paint our custom icon and text
-    paintIcon(painter, opt, index);
-    paintText(painter, opt, index);
+    paintIcon(painter, opt, userData.icon);
+    paintText(painter, opt, userData.text);
 }
 
 void SuggestionListDelegate::paintIcon(QPainter *p,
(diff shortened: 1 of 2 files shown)


kelson42:
> I haven't verified that the new solution preserves the intended effect of #1414 (elimination of the failure of an assertion when using a debug build of Qt) since I don't have easy access to a debug variant of Qt libs, however I don't see why that problem should return with my fix.

This part will be difficult to check before merging I guess. Therefore merging, and testing the nightly would the proposed approach from my side.

kelson42 (approved):
From a user perspective, LGTM

PR #1434, opened 2025-12-18
Reorganise Save as... to Save article and allow saving as both pdf and mhtml

### LINKED ISSUE #1433: Reorganise "Save as..." to save article
Today the "Save as..." is available topbar menu:

Image

and in right-click context menu

Image

This has many problems:
* We can not save as `.mhtml`
* The contextual menu part comes in conflict with saving a resources (images, videos, ...)

We should:
* Keep only the topbar menu
* In the dialog we should be able to choose the `.mhtml` format (at the top bottom selector of the file picker) amd save in .mhtml
* 

Image

vighnesh-sawant:
Fixes #1433 

Most of the code change is adding the option to save as mhtml, should i split this off into another commit?
Also I have moved the downloadFinished() code out of the KProfile namespace to allow webview to also use it.


vighnesh-sawant:
@kelson42 Please do have a look at this

kelson42:
@vighnesh-sawant I just have tested and here my feedbacks:

* Thank you for splitting this, in general it works, but we have to fix a few things

* We should rephrase the dialog. I would propose title to be 'Article saved' with content  'The article "foobar" has been saved at: <carrirage return>/my/path.pdf'. But maybe you have a better idea (would be great to have a toast message like on Android). 'ok' button should always be written uppercase (so 'OK').
image

* If I save three time in a row the same article at the same path I get a warning tell me the file will be reaplaced (this part is OK)... but then the dialog box (above) is displayed 3 times in a row!

* Don't call the mhtml format "Web archive" (which is pretty unclear for the end user), just says "Web Page, complete" (like on Firefox)

* The file extension should not be visible in the filepicker dialog (so not `foobar.pdf`, but just `foobar`). This is important, otherwise you will have to handle the extension change, depending what people choose (which might not even be possible).

* You have to add the file extension at save time. Use `.pdf` or `.html` (so no `.mhtml`). If you see that the user has already added an extension (so `.pdf` or `.html`) please don't add one on the top.

vighnesh-sawant:
@kelson42 
1.Will work on this. 
Maybe we can make adding toasts a new issue? Qt does not have toasts by default, we would have to implement them/ import a library in our project for it.
2.Fixed this!
3. Done
4. Qt already handles the extension change, would you prefer no extension or qt handling it?
Also I wanted to confirm does it do that on your machine also? (I suspect it may not) (It does on mine)
5. Changed to html

vighnesh-sawant:
1.Changed the dialog
image

vighnesh-sawant:
@kelson42 Bump!
I dont really know if the last comment pinged you or not.

PR #1436, opened 2025-12-25
docs: clarify Windows usage and add quick start section

likhith1253:
This PR clarifies usage for Windows users.

It adds a small quick start section for end users and explains that the
existing Windows compilation instructions are intended for advanced or
WSL-based setups.

Docs-only change, no impact on builds or CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,229 +1,200 @@
-     <a href="https://flathub.org/apps/details/org.kiwix.desktop">
-     <img width="100" alt="Download on Flathub" src="https://flathub.org/assets/badges/flathub-badge-en.png" />
-     </a>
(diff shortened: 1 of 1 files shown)


kelson42:
@likhith1253 This PR breaks everything, see:
image

Please change the bare minimum to achieve your goal. Improving only one or two sections should be enough. To me is is not clear at this stage what is exactly the problem you try to fix (user? developers?) Why WSL which is basically a Linux?

likhith1253:
Thanks for the clarification.

I’ve reset the changes and am preparing a new PR with a minimal, focused
update that only adds a short Windows installation note, without modifying
any existing build instructions.Also, sorry for the confusion created with
linux and WSL.

PR #1440, opened 2025-12-27
docs: clarify Windows installation using prebuilt binaries

likhith1253:
This PR adds a short note to the Installation section clarifying that
Windows users can use the official prebuilt binaries.

This avoids confusion for Windows users while keeping the existing
build instructions unchanged.


likhith1253:
can i know if there is any issue with this pr... like do i need to modify anything else

PR #1467, opened 2026-02-15
Remove Qt5 support

### LINKED ISSUE #1300: Requires Qt6

kelson42:
Fixes #1300

* Update `README.md`
* Remove `ubuntu-22.04/qt5` (standart version) compilation from CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,8 +30,8 @@ jobs:
     - name: Install QT
       uses: jurplel/install-qt-action@v4
       with:
-        version: 5.15.2
-        modules: "qtwebengine"
+        version: 6.8.3
+        modules: "qtwebengine qtwebchannel qtpositioning"
         setup-python: false
 
     - name: Setup MSVC compiler
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,12 +16,12 @@ jobs:
     runs-on: windows-2025
     steps:
     - name: Checkout code
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
 
-    - name: Setup python 3.13
-      uses: actions/setup-python@v5
+    - name: Setup python 3.14
+      uses: actions/setup-python@v6
       with:
-        python-version: '3.13'
+        python-version: '3.14'
 
     - name: Install packages
       run: |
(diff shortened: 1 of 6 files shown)


veloman-yunkan (commented):
I think that the PR title is a little misleading (at least I hope so, since I want to be able to build `kiwix-desktop` on Ubuntu 22.04 jammy using Qt5).

kelson42:
> I think that the PR title is a little misleading (at least I hope so, since I want to be able to build `kiwix-desktop` on Ubuntu 22.04 jammy using Qt5).

I don't understand this comment. 2.6.0 will build only on Qt6 because there is no reason to keep supporting buggy old version of Qt.

veloman-yunkan:
I meant that this PR doesn't enforce Qt6 and therefore building `kiwix-desktop` on Ubuntu 22.04 is still possible, and I like that because it enables me to keep developing also on my computer running jammy.

kelson42:
@veloman-yunkan That is indeed correct, although bit surprising to me. Should we make a formal check in the .pro file?

.github/workflows/ci.yml
```diff
@@ -55,23 +55,7 @@ jobs:
...
-            qt_version: 5
-          - target: noble-qt6.4
-            image_variant: noble
-            lib_postfix: '/x86_64-linux-gnu'
-            qt_version: 6
-          - target: oracular-qt6.6
-            image_variant: resolute
-            lib_postfix: '/x86_64-linux-gnu'
-            qt_version: 6
+        image_variant: [noble, resolute]
```
adamlamar, line 59: nit: this works fine, but removing the `-qt6.x` suffix drops some helpful information IMO. There always seem to be minor issues in qt and quickly knowing which version a particular CI job is running might be helpful in tracking down the problem.

adamlamar:
LGTM. However note that the Ubuntu 22.04 OS is still supported for another year or so.

adamlamar (approved):

.github/workflows/ci.yml
kelson42: @adamlamar Each time we will rebuild the container image, this might change, I can not put it here, hardcoded. But I have put a `qmake --version` in the workflow.

kelson42:
> I think that the PR title is a little misleading (at least I hope so, since I want to be able to build `kiwix-desktop` on Ubuntu 22.04 jammy using Qt5).

`qmake` nows exits with error if it does not have Qt5

kelson42:
@legoktm Unfortunately, I'm in the impossibility tomget the package compiling properly with Qt6. Do you have maybe an idea why?

PR #1475, opened 2026-03-27
Modernize UI with Chromium-inspired Frontend for enhanced UX

Prathvi321:
This PR introduces a comprehensive overhaul of the Kiwix Desktop frontend. The primary goal of this redesign is to align the user interface with modern web browser standards (specifically the Chromium/Chrome layout). By leveraging a familiar design language, we reduce the learning curve for new users and provide a more intuitive navigation experience for offline content.

**Key Improvements**

- **Familiar Tab Management**: Reimplemented the tab bar to mimic modern browser behavior, including "New Tab" functionality and improved drag-and-drop support.

- **Unified Address & Search Bar:** Combined the navigation and search functions into a central "Omnibox" style header for quicker access to ZIM content.

- **Streamlined Sidebar:** Relocated library management and bookmarks into a collapsible side panel, maximizing screen real estate for reading.

- **Modernized Iconography:** Updated the UI assets to a cleaner, flat-design aesthetic consistent with current desktop environments.

**Motivation**
While the current Kiwix interface is functional, many users coming from standard web browsers find the specialized layout a point of friction. By adopting a "Chrome-like" interface:

1. **Onboarding is faster**: Users already know where the back button, tabs, and settings are located.
2. **Productivity increases:** The layout prioritizes the reading area while keeping navigation tools within a "muscle memory" distance.
3. **Visual Appeal:** The application feels more like a modern 2026 desktop app rather than a legacy utility.

### DIFF
--- a/resources/css/style.css
+++ b/resources/css/style.css
@@ -1,6 +1,6 @@
 QWidget {
-    background-color: #EAECF0;
-    color: #000000;
+    background-color: #DFE1E5;
+    color: #202124;
 }
 
 
@@ -14,7 +14,8 @@ TopWidget {
     margin: 0;
     padding: 5px;
     border: none;
-    border-bottom: 1px solid #ccc;
+    border-bottom: 1px solid #DADCE0;
+    background-color: #FFFFFF;
 }
 
 TopWidget::separator {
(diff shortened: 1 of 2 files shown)


kelson42:
@Prathvi321 Thank you very much for your PR. At this stage I have no judgement about your PR, but this will be hard to merge because:
* For each of the point you bring, I have no issue and I'm therefore not even sure about a problem or the real nature of the problem
* No screenshots before, after or detailed description about what has changed
* One single commit with all the changes, so no way to isolate/understand clearly what has been done.

Would you be able to isolate the improvements you have done to the tabbar and bring it to a dedicated PR and clearly explains both problems and corresponding solution?

Issue #810, opened 2022-03-02
kiwix-desktop appimage webengine is broken on Fedora (and ubuntu 22.04)


mgautierfr:
kiwix-desktop appimage (both released (https://download.kiwix.org/release/kiwix-desktop/kiwix-desktop_x86_64_2.1.0-1.appimage) and nigthly (https://download.kiwix.org/nightly/2022-03-02/kiwix-desktop_x86_64_2022-03-02.appimage)) do not display anything in the web views (library and zim) on Fedora 35

Appimage is run with the workaround explained here https://github.com/kiwix/kiwix-desktop/issues/745#issuecomment-1017530749
`MESA_LOADER_DRIVER_OVERRIDE=i965 ./kiwix-desktop_x86_64_2022-03-02.appimage`

![Capture d’écran du 2022-03-02 15-00-18](https://user-images.githubusercontent.com/86161/156375978-72cfd7da-a32f-41bf-80dd-ba49a18b9239.png)

Else kiwix-desktop seems to work correctly. It is possible to open a zim file (but we cannot see its content. Traces are ok....
kiwix-desktop compiled localy works fine.

mgautierfr:
AppImage build on local system works correctly.

mgautierfr:
Few more information :

- Last nightly version (https://download.kiwix.org/nightly/2022-03-16/kiwix-desktop_x86_64_2022-03-16.appimage) doesn't need the workaround as it as been fixed by https://github.com/kiwix/kiwix-build/pull/520
- Running kiwix-destkop appimage with env var `QTWEBENGINE_REMOTE_DEBUGGING=8080` start a server to debug QtWebengine but connecting to localhost:8080 show no page to inspect.
- However, subprocesses `QtWebEngineProcess` have been correctly (?) started.

I've tried to trace dynamic libraries loaded if somethings seems wrong but I've found nothing relevant.
Pinging @probonopd maybe he has a idea of what going on or a lead to follow.

probonopd:
@mgautierfr 

> AppImage build on local system works correctly.

Are you saying you have a working and a non-working AppImage? Then it's probably a matter of tracing/comparing the two...

mgautierfr:
Yes, I've tried to compare strace/`LD_DEBUG=1`/`QT_DEBUG_PLUGINS=1 QML_IMPORT_TRACE=1` but I've found nothing convincing.
[...]

PR #1478, opened 2026-04-01
Undo the hack breaking WebGL in appimage

### LINKED ISSUE #1477: appimage build (at least) does not support WebGL
Test ZIM: https://tmp.kiwix.org/zims/webgl-test_en_all_2026-03-31.zim

Expected result:
Image

ZIM source code: https://gist.github.com/benoit74/abb74a205ac80abb48a52931e07b1b35

Works OK on locally built kiwix-desktop with Qt6.4.2 on Ubuntu 24.04 LTS

Doesn't work with appimage nightly (from `2026-03-27`). Errors (last error probably comes from the HTML itself):

```bash
> ./kiwix-desktop_x86_64_2026-03-27.appimage 
Warning: Ignoring WAYLAND_DISPLAY on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
qt.webenginecontext: 

GLImplementation: desktop
Surface Type: OpenGL
Surface Profile: CompatibilityProfile
Surface Version: 4.6
Using Default SG Backend: yes
Using Software Dynamic GL: no
Using Angle: no

Init Parameters:
  *  application-name kiwix-desktop 
  *  browser-subprocess-path /tmp/.mount_kiwix-tT20CY/usr/libexec/QtWebEngineProcess 
  *  create-default-gl-context  
  *  disable-features DnsOverHttpsUpgrade,ConsolidatedMovementXY,InstalledApp,BackgroundFetch,WebOTP,WebPayments,WebUSB,PictureInPicture 
  *  disable-gpu  
  *  disable-setuid-sandbox  
  *  disable-speech-api  
  *  enable-features NetworkServiceInProcess,TracingServiceInProcess 
  *  enable-threaded-compositing  
  *  in-process-gpu  
  *  no-sandbox  
  *  use-gl desktop 

add widget
js: Uncaught Error: WebGL not supported
```

This is a blocker for new `maps` scraper release.

veloman-yunkan:
Fixes #1477

The hack was introduced as a workaround for a bug in `linuxdeployqt` (see #810). Since we no longer use `linuxdeployqt` in our AppImage creation script (see kiwix/kiwix-build#771) the hack is rolled back in the hope that it doesn't revive the issue reported in #810 (I didn't test that; I only checked that the AppImage created after applying this change doesn't have the WebGL problem unlike the AppImage created before applying this change).

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -34,14 +34,6 @@ bool wasAppStartedFromARemoteDrive()
 
 int main(int argc, char *argv[])
 {
-// Small hack to make QtWebEngine works with AppImage.
-// See https://github.com/probonopd/linuxdeployqt/issues/554
-    if (qEnvironmentVariableIsSet("APPIMAGE")) {
-        qputenv("QTWEBENGINE_CHROMIUM_FLAGS", "--disable-gpu --no-sandbox");
-        QApplication::setAttribute(Qt::AA_UseOpenGLES);
-    }
-// End of hack ^^^
-
 #ifdef Q_OS_WIN
     const bool appWasStartedFromARemoteDrive = wasAppStartedFromARemoteDrive();
     if ( appWasStartedFromARemoteDrive ) {


benoit74 (approved):
LGTM, I would have done the same move but didn't dare to propose it 😉

Issue #1465, opened 2026-02-14
Persistence of libtorrent session state


veloman-yunkan:
`libtorrent` session must be persisted so that incomplete downloads can be resumed after restarting `kiwix-desktop`.

ayushman1210:
Hi @veloman-yunkan,
Are you currently working on this issue? If not, I would be happy to take it up and submit a PR.
Please let me know if I can proceed.
Thank you!

veloman-yunkan:
@ayushman1210 Thank you for your interest in contributing to Kiwix projects but please look for another ticket. This one is on my plate :)

veloman-yunkan:
@kelson42 #1496 achieves the desired effect with minimal code (though can be slow for large downloads). The question is whether we should pursue real persistence of libtorrent resume data now, or we can defer that enhancement until later times.

kelson42:
> [@kelson42](https://github.com/kelson42) [#1496](https://github.com/kiwix/kiwix-desktop/pull/1496) achieves the desired effect with minimal code (though can be slow for large downloads). The question is whether we should pursue real persistence of libtorrent resume data now, or we can defer that enhancement until later times.

@veloman-yunkan I'm not sure to fully understand what "real" means in this context. Kind of guess that we would save the chunk list rebuilding time... but not sure. To me sounds like an optim, and glad to have a dedicated issue.... if needed at all.

veloman-yunkan:
> I'm not sure to fully understand what "real" means in this context. Kind of guess that we would save the chunk list rebuilding time...

Exactly

> To me sounds like an optim, and glad to have a dedicated issue.... if needed at all.

I concur on this. Let's get back to it when it shows up as a problem worth to be addressed.

PR #1469, opened 2026-02-20
Support for libtorrent-based downloads

veloman-yunkan:
Fixes #1464

This PR introduces minimal support for `libtorrent`-based downloads on Linux. The `libtorrent-rasterbar-dev` package must be installed on the build system. Thus, this feature is disabled on Windows.

`libtorrent` kicks in (instead of `aria2c`) if the URL of the OPDS catalog entry is a magnet link (rather than
  an HTTP link).

PR #1496, opened 2026-06-23
Libtorrent downloads persist across kiwix-desktop sessions

veloman-yunkan:
Related to #1465

As a result of this enhancement incomplete libtorrent-backed downloads from the previous app session now reappear when kiwix-desktop is launched again (this should apply to a crash of kiwix-desktop too).

However no special download resume data is persisted. The download state is restored from the partially downloaded file (which has to be fully scanned in order to detect the locally available pieces).

Note that resuming a libtorrent download remaining from a previous session doesn't immediately show the progress of the download. The download progress data appears after the torrent data is re-fetched from the magnet link and the local file is compared against it. Both of these steps can take a while.

### DIFF
--- a/src/downloadmanagement.cpp
+++ b/src/downloadmanagement.cpp
@@ -18,12 +18,26 @@ namespace
 {
 
 #if defined(ENABLE_LIBTORRENT)
+const std::string LTPREFIX = "libtorrent:";
+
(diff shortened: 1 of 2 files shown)


veloman-yunkan:
Can be tested using the package provided in https://github.com/kiwix/kiwix-desktop/pull/1469#issuecomment-3935517055

PR #1507, opened 2026-09-05
fix resolve surfing latency, missing stemming log, and window switch …

### LINKED ISSUE #1506: Surfing has become really slow with the dev version
Each type I click on a link it take 2 second to load the article...

Two strange things:
* I always see `No stemming for language 'ms'` on the log (whatever the language of the ZIM I read)
* After clicking on a link, the title of the window switch shortly to `Library - Kiwix`

Probably the bug is very stupid and very bad impact. Loading a page should be "immediate"

ayush213:
Problem:
In the development version, surfing has experienced a noticeable performance regression where clicking any link takes approximately 2 seconds to load. Additionally, two distinct bugs occur concurrently: a persistent log warning stating No stemming for language 'ns' for ZIM files, and an unexpected window focus switch that briefly brings the Library view to the foreground instead of keeping focus on the active article reader.
Solution:
Investigated and resolved the latency bottlenecks causing a delay during article loading and link traversal.
Fixed the language code checking and fallback logic to eliminate the No stemming for language 'ns' log spam.
Corrected the UI state and event handling to prevent the window focus from unintentionally switching to the Library view when links are clicked.
Closes #1506

### DIFF
--- a/src/tabbar.cpp
+++ b/src/tabbar.cpp
@@ -352,7 +352,10 @@ void TabBar::onCurrentChanged(int index)
         emit webActionEnabledChanged(QWebEnginePage::Back, view->isWebActionEnabled(QWebEnginePage::Back));
         emit webActionEnabledChanged(QWebEnginePage::Forward, view->isWebActionEnabled(QWebEnginePage::Forward));
         emit tabDisplayed(TabType::ZimViewTab);
-        QTimer::singleShot(0, [=](){emit currentTitleChanged(view->title());});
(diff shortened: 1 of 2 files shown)


kelson42:
@ayush213 Thank younfor the fix. Can you split this PR in two PRs considering they do very different things? Inwill split the issue on two issues, on my side.

ayush213:
@kelson42 Thanks for the feedback! I have split this into two separate pull requests as requested to address each change independently. I am closing this PR in favor of the new ones.

PR #1508, opened 2026-09-06
fix resolve surfing latency, missing stemming log, and window switch …

ayush213:
Surfing Performance Fix (PR #1506 - Related to the earlier discussion on fixing surfing latency)

Problem Statement: In the development version, surfing has experienced a noticeable performance regression where clicking any link takes approximately 2 seconds to respond.

Solution: Optimized event handling and timer triggers in the tab bar component (src/tabbar.cpp) to eliminate the delay and restore snappy navigation performance.


ayush213:
@kelson42 check

kelson42:
@ayush213 the two different fixes are still in only one PR?!?!

kelson42:
@ayush213 Any feedback?

